PR #48, opened 2021-07-28
TOC in a foldable sidebar

TheLortex:
Some progress on https://github.com/tmattio/ocaml.org/issues/19, https://github.com/tmattio/ocaml.org/issues/34 and https://github.com/tmattio/ocaml.org/issues/12

Folded:

Screenshot 2021-07-28 at 15-32-37 Packages

Unfolded:

Screenshot 2021-07-28 at 15-32-52 Packages

A global navbar like https://hexdocs.pm/phoenix/overview.html  is hard to achieve because the website isn't a full-page one. I therefore opted for a local one.
On mobile the bar will go over the content instead of moving it around

### DIFF
--- a/lib/ocamlorg_web/templates/pages/package_doc_template.eml
+++ b/lib/ocamlorg_web/templates/pages/package_doc_template.eml
@@ -19,11 +19,13 @@ let toc = toc_of_toc doc.toc in
         </div>
       </div>
       <div class="relative flex-1 bg-white lg:shadow-md">
+        <% if toc <> [] then begin %>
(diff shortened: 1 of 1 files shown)


TheLortex:
Voodoo has been updated to generate a `package.json` file (example: https://docs-data.ocaml.org/current/p/base/v0.14.1/package.json). 

I'm now working on generating the global navbar from this file's data

TheLortex:
New PR as this is definitely not a foldable sidebar and that code is too old in any case

PR #52, opened 2021-07-29
Improve mobile view of the package site

### LINKED ISSUE #13: Mobile view of the documentation looks off

tmattio:
Fixes #13


TheLortex:
Hello, this PR broke a little the rendering of the navigation breadcrumbs on Firefox:

Screenshot 2021-09-01 at 12-08-10 Base Applicative · base v0 14 1 · OCaml Packages

The culprit is the addition of `overflow-auto`.  If I understand correctly, the only kind of overflow needed on mobile is `overflow-x`, so the correct classes should be `overflow-x-auto` and `overflow-y-hidden`. What do you think ?

Screenshot 2021-09-01 at 12-08-26 Base Applicative · base v0 14 1 · OCaml Packages

tmattio:
That makes sense! I noticed it as well but was waiting for more feedback on the design to address them in bulk. Feel free to send a PR or open an issue and I'll fix that when I get back to working on the design 🙂

PR #60, opened 2021-07-30
Add first graphQL endpoint

dinakajoy:
This creates the first graphQL endpoint to return packages and package

### DIFF
--- a/lib/ocamlorg_web/graphql.ml
+++ b/lib/ocamlorg_web/graphql.ml
@@ -29,8 +29,8 @@ let is_package s1 s2 =
     let s2 = String.lowercase_ascii s2 in
     String.(equal s2 s1)
 
-let get_packages_result total_packages offset limit startsWith packages =
-  match startsWith with
+let get_packages_result total_packages offset limit startswith packages =
+  match startswith with
   | None ->
     let packages =
       List.filteri (fun i _ -> offset <= i && i < offset + limit) packages
@@ -40,11 +40,11 @@ let get_packages_result total_packages offset limit startsWith packages =
   | Some filter ->
     let packages =
       List.filteri
-        (fun i _ -> offset <= i && i < offset + limit)
-        (List.filter
-           (fun package ->
-             starts_with filter (Package.Name.to_string (Package.name package)))
-           packages)
+        (fun i package ->
+          offset <= i
+          && i < offset + limit
+          && starts_with filter (Package.Name.to_string (Package.name package)))
+        packages
     in
     let packages_result = { total_packages; packages } in
     packages_result
@@ -249,7 +249,7 @@ let schema : Dream.request Graphql_lwt.Schema.schema =
                   ~doc:
                     "Filter packages by passing a search query which lists out \
                      all packages that starts with the search query if any"
-                  "startsWith"
+                  "startswith"
                   ~typ:string
               ; arg'
                   ~doc:
@@ -267,13 +267,13 @@ let schema : Dream.request Graphql_lwt.Schema.schema =
                   "limit"
                   ~typ:int
               ]
-          ~resolve:(fun _ () startsWith offset limit ->
+          ~resolve:(fun _ () startswith offset limit ->
             let packages = Package.all_packages_latest () in
             let total_packages = List.length packages in
             let limit =
               match limit with None -> total_packages | Some limit -> limit
             in
-            get_packages_result total_packages offset limit startsWith packages)
+            get_packages_result total_packages offset limit startswith packages)
       ; field
           "package"
           ~typ:package


lib/ocamlorg_web/graphql.ml
```diff
@@ -0,0 +1,330 @@
...
+            ~args:Arg.[]
+            ~typ:url
+            ~resolve:(fun _ p ->
+              let info = Package.info p in
+              info.Package.Info.url)
+        ]))
+
+let response =
+  Graphql_lwt.Schema.(
+    obj "response" ~fields:(fun _ ->
```
tmattio, line 244: I would specialize the name of this field to the package response. E.g. `packages_result`

lib/ocamlorg_web/graphql.ml
```diff
@@ -0,0 +1,330 @@
...
+  if len1 > String.length s2 then
+    false
+  else
+    let s1 = String.lowercase_ascii s1 in
+    let s2 = String.lowercase_ascii s2 in
+    String.(equal (sub s2 0 (length s1)) s1)
+
+let is_package s1 s2 =
+  let len1 = String.length s1 in
+  if len1 > String.length s2 then
```
tmattio, line 35:
```suggestion
  if len1 <> String.length s2 then
```

lib/ocamlorg_web/graphql.ml
```diff
@@ -0,0 +1,330 @@
...
+  Graphql_lwt.Schema.(
+    obj "response" ~fields:(fun _ ->
+        [ field
+            "totalPackages"
+            ~doc:"total number of packages"
+            ~args:Arg.[]
+            ~typ:(non_null int)
+            ~resolve:(fun _ p -> p.total_packages)
+        ; field
+            "nodes"
```
tmattio, line 252: I wouldn't use `nodes` here, this conflicts with Relay's conventions and might confuse users. How about `packages`?

tmattio (commented):
Fantastic work @dinakajoy, thanks for the contribution 🎉 🎉 

I left a few comments, but LGTM overall!

lib/ocamlorg_web/graphql.ml
```diff
@@ -0,0 +1,330 @@
+open Ocamlorg
+
+type package_dep =
+  { name : string
+  ; constraints : string option
+  }
+
+type package_owner =
```
patricoferris, line 8: I think we can do a similar thing here that we did offline with `Package.Info.url` and reuse the `Opam_user` type. If you look at your `get_owners` function below it is simply rewrapping each field into this new type so since there's no converting of values, I think it makes sense to reuse the `Opam_user` type. 

So the schema for the user would replace the resolvers to directly use `Opam_user` for example the name field would become.

```ocaml
field
    "name"
    ~typ:(non_null string)
    ~args:Arg.[]
    ~resolve:(fun _ user -> user.Opam_user.name)
```

lib/ocamlorg_web/graphql.ml
```diff
@@ -0,0 +1,330 @@
...
+  { full_name : string
+  ; handle : string option
+  ; image : string option
+  }
+
+type package_url =
```
patricoferris, line 14: I believe this type can be removed now, I think it was left over from when you were converting the `Package.Info.url` type but now we are using it directly.

patricoferris (changes requested):
This is excellent @dinakajoy and thanks @tmattio for the helpful review (I concur with everything you said), just some little bits and pieces and then I think we can merge this. Let me know if anything is unclear or if you need help with any of our suggestions :))

lib/ocamlorg_web/graphql.ml
tmattio: We should add an `mli` file so that unused variables and types are caught at compile time

tmattio:
Also, I know we haven't been very rigorous on the tests (that's an understatement..), but it would be nice to take good habits now that the repository is open to external contributions. I'm fixing the CI in #68, so we'll be able to run the tests in the CI.

patricoferris:
> Also, I know we haven't been very rigorous on the tests (that's an understatement..), but it would be nice to take good habits now that the repository is open to external contributions. I'm fixing the CI in #68, so we'll be able to run the tests in the CI.

Sounds like a great idea, and this something @dinakajoy had expressed a desire to learn, perhaps we could do that in a follow-up PR, I can create the issue for it or would you prefer it in this one ?

tmattio:
> Sounds like a great idea, and this something @dinakajoy had expressed a desire to learn, perhaps we could do that in a follow-up PR, I can create the issue for it or would you prefer it in this one ?

Ideally, we would implement the tests at the same time as the PR, but given that we don't have any at the moment, I don't think it's reasonable to block the PR, especially as the PR is not impacting the rest of the code. I'll defer to you and @dinakajoy to choose how we want to proceed.

tmattio:
(although we can at least rebase to make sure the build and lint pass)

patricoferris:
Okay, we'll probably revisit on Monday. We could also implement a PR template to mention all of these things (tests, `make fmt` etc.) ?

dinakajoy:
Regarding the test, I am still new to that but will definitely write a test for this once I can.

lib/ocamlorg_web/graphql.ml
```diff
@@ -0,0 +1,307 @@
...
+    false
+  else
+    let s1 = String.lowercase_ascii s1 in
+    let s2 = String.lowercase_ascii s2 in
+    String.(equal s2 s1)
+
+let get_packages_result total_packages offset limit filter packages =
+  if filter = "sortByName" then
+    let packages =
+      List.filteri (fun i _ -> offset <= i && i <= offset + limit) packages
```
patricoferris, line 34: I think we want `offset <= i && i < offset + limit` here as we are off by one (asking for `offset: 10, limit: 10` returns `11` packages, I would suspect most would think it should return 10) ? This would be a good thing to test in the follow-up PR.

lib/ocamlorg_web/graphql.ml
```diff
@@ -0,0 +1,307 @@
...
+                  ~default:0
+              ; arg'
+                  ~doc:
+                    "Specifies the limit which means the number of packages \
+                     you want to return if you do not want all packages \
+                     returned at once. By default, all the packages are \
+                     returned"
+                  "limit"
+                  ~typ:int
+                  ~default:0
```
patricoferris, line 270: Can we make this an optional type i.e. 

```ocaml
arg
  ~doc:
    "Specifies the limit which means the number of packages \
     you want to return if you do not want all packages \
     returned at once. By default, all the packages are \
     returned"
  "limit"
  ~typ:int
```

And then below pattern-match on `limit` with:

```ocaml
match limit with 
  | None -> ... (* return all *)
  | Some limit -> ... (* limit packages *)
```

Using `0` here feels a bit wrong to me, if someone (for whatever reason...) made a query like `allPackages(offset: 20, limit: 0, filter: "sortByName")` I don't think they would expect all of the packages to be returned.

patricoferris (commented):
Once these changes are made I think this is ready to go, if you could also rebase on the `main` branch, that should pull in the latest Github Actions file and we should get all the tests to pass that way :))

patricoferris:
Thanks @dinakajoy all tests are passing! 

It doesn't look like you made the changes to the `limit` parameter that I suggested? Instead you've changed the filter default to be `""` instead of `sortByName`. Which now seems correct, there is no sorting by name right?

I think both `filter` and `limit` could do with being changed into optional types as I suggested. Instead of then using `0` and `""` to distinguish the "not used case" you would use `None`. And then otherwise you can get the parameter from `Some param`, does that make sense?

dinakajoy:
> It doesn't look like you made the changes to the `limit` parameter that I suggested? Instead you've changed the filter default to be `""` instead of `sortByName`. Which now seems correct, there is no sorting by name right it's supposed to filter using `starts_with`?

I made the changes.
It was => List.filteri (fun i _ -> offset <= i && i <= offset + limit) packages
But now changed to => List.filteri (fun i _ -> offset <= i && i < offset + limit) packages
I also tested it and you were right, thank you for pointing that out.

> I think both `filter` and `limit` could do with being changed into optional types as I suggested. Instead of then using `0` and `""` to distinguish the "not used case" you would use `None`. And then otherwise you can get the parameter from `Some param`, does that make sense?
Sure, I will do that. It sounds great.

dinakajoy:
> > I think both `filter` and `limit` could do with being changed into optional types as I suggested. Instead of then using `0` and `""` to distinguish the "not used case" you would use `None`. And then otherwise you can get the parameter from `Some param`, does that make sense?

Done, please confirm

lib/ocamlorg_web/graphql.ml
```diff
@@ -0,0 +1,307 @@
...
+              let packages_result =
+                get_packages_result totalPackages offset limit filter packages
+              in
+              packages_result
+            | Some limit ->
+              let limit = limit in
+              let packages_result =
+                get_packages_result totalPackages offset limit filter packages
+              in
+              packages_result)
```
pitag-ha, line 286:
```suggestion
            let limit = 
            match limit with
              | None -> totalPackages
              | Some limit -> limit
            in
            get_packages_result totalPackages offset limit filter packages
            )
```

Probably, I haven't formatted this correctly.

lib/ocamlorg_web/graphql.ml
```diff
@@ -0,0 +1,307 @@
...
+      ; field
+          "package"
+          ~typ:package
+          ~doc:
+            "Returns details of a specified package. It returns the latest \
+             version if no version is specifed or returns a particular version \
+             of the package if a specified"
```
pitag-ha, line 293: Is it true what you say about the version? I don't find where in the code you allow to specify a version and where you handle it.

lib/ocamlorg_web/graphql.mli
```diff
@@ -0,0 +1 @@
+val schema : Dream.request Graphql_lwt.Schema.schema
```
pitag-ha, line 1: It might be nice to add a general comment to point out that this schema reflects the data on the opam-repository and allows to query for opam packages. If I'm not missing anything, without a comment it's not directly clear that the graphql endpoint contains information about opam packages.

pitag-ha (commented):
Nice work, @dinakajoy! I just have a few very minor comments.

lib/ocamlorg_web/graphql.ml
```diff
@@ -0,0 +1,307 @@
...
+  | Some filter ->
+    let packages =
+      List.filter
+        (fun package ->
+          starts_with filter (Package.Name.to_string (Package.name package)))
+        packages
+    in
+    let packages =
+      List.filteri (fun i _ -> offset <= i && i < offset + limit) packages
+    in
```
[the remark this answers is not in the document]
dinakajoy: > You could merge the two filterings into one by using `List.filteri` just once and making use of the second parameter of its filter function for the `starts_with` filtering. What do you think?

You mean like 
` let packages =
      List.filteri (fun i _ -> offset <= i && i < offset + limit) (List.filter (fun package ->
          starts_with filter (Package.Name.to_string (Package.name package)))
        packages)
    in
    let packages_result = { total_packages; packages } in
    packages_result`

lib/ocamlorg_web/graphql.ml
```diff
@@ -0,0 +1,307 @@
...
+      [ field
+          "allPackages"
+          ~typ:(non_null packages_result)
+          ~doc:
+            "Filter packages by name passing search query. Packages can also \
+             be paginated by setting the offset and limit as desired"
+          ~args:
+            Arg.
+              [ arg
+                  ~doc:"Filter packages by passing a search query"
```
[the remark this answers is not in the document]
dinakajoy: > What kind of search query?

Like `allPackages(filter: "oc"){}`
will return all packages that starts with "oc"

lib/ocamlorg_web/graphql.ml
dinakajoy: > Is it true what you say about the version? I don't find where in the code you allow to specify a version and where you handle it.

Yea, it was there before but you guys mentioned to leave out the versions for now. So I forgot to update this line

lib/ocamlorg_web/graphql.ml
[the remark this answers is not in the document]
pitag-ha: I mean something along the lines
```
let packages =
   List.filteri (fun i package ->
                        offset <= i && i < offset + limit &&
                        starts_with filter (Package.Name.to_string (Package.name package))
  ) packages
in

```

lib/ocamlorg_web/graphql.ml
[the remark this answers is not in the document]
pitag-ha: Ok, thanks! Do you think it might make sense to make that clearer in the `doc` or do you think it's already quite clear?

lib/ocamlorg_web/graphql.ml
pitag-ha: I think that to specify a version you only need a check if two versions are equal and not a version inequality check, so it should be straight-forward and uncontroversial to add it. So if you've already written it, I'd say it's ok to add it (@tmattio and @patricoferris , do you agree?). But I wouldn't invest time into it now.

patricoferris:
Thanks for the review @pitag-ha, all sounds very reasonable to me. Once these changes land, I think we're ready to merge. Do ping me @dinakajoy when they're in :))

lib/ocamlorg_web/graphql.ml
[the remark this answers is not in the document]
dinakajoy: This is cool but I think it will make that line to ambiguous to understand

lib/ocamlorg_web/graphql.ml
dinakajoy: > Thanks for the review @pitag-ha, all sounds very reasonable to me. Once these changes land, I think we're ready to merge. Do ping me @dinakajoy when they're in :))

Hi, @patricoferris the changes has been effected.

lib/ocamlorg_web/graphql.ml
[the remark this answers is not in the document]
pitag-ha: With "ambiguous", do you mean that someone reading the code could understand it in two different ways? What are those two ways?

An advantage of only filtering once would be a small improvement in performance: instead of going through the list twice, you only go through once. But to be fair, it probably wouldn't make a huge difference.

lib/ocamlorg_web/graphql.ml
pitag-ha: In the end, do you prefer to correct the `doc`? Or to add the code to handle the version?

lib/ocamlorg_web/graphql.mli
```diff
@@ -0,0 +1,9 @@
+(* This function compares two string and returns true if they are equal *)
```
pitag-ha, line 1:
```suggestion
(* This function compares two strings and returns true if they are equal up to capitalization *)
```

lib/ocamlorg_web/graphql.mli
```diff
@@ -0,0 +1,9 @@
...
+val starts_with : string -> string -> bool
+
+(* This function compares two string and returns true if they are equal *)
```
pitag-ha, line 4:
```suggestion
(** This function compares two strings and returns true if they are equal up to capitalization *)
```

lib/ocamlorg_web/graphql.mli
```diff
@@ -0,0 +1,9 @@
...
+val is_package : string -> string -> bool
+
+(* This schema allows to query for opam packages from the opam-repository and
+   returns graphql data as graphiql playground and grphql json data*)
```
pitag-ha, line 8: Is "and returns graphql data as graphiql playground and grphql json data" part of the schema definition? I was under the impression that that was done in `router.ml`.

patricoferris:
Why has `get_deps` been renamed `get_info` ? Was that in a code review that I missed perhaps? I think `deps` is maybe clearer ? (it is also why the CI is failing, the `mli` file declares a function `get_deps` and the `ml` file changed it to `get_info`, similarly the name of the type was also changed).

dinakajoy:
> Why has `get_deps` been renamed `get_info` ? Was that in a code review that I missed perhaps? I think `deps` is maybe clearer ? (it is also why the CI is failing, the `mli` file declares a function `get_deps` and the `ml` file changed it to `get_info`, similarly the name of the type was also changed).

Yea, I just noticed that it was also for  depopts and  conflicts not only dependencies, hence I renamed it to get_info but forgot to change it in the test. I am pushing the updated one now.

lib/ocamlorg_web/graphql.mli
```diff
@@ -0,0 +1,35 @@
...
+  ; packages : Package.t list
+  }
+
+val starts_with : string -> string -> bool
+(** This function compares two strings and returns true if they are equal up to
+    capitalization *)
+
+val is_package : string -> string -> bool
+(** This function compares two strings and returns true if the first string is
+    equal to the beginning of the second string up to capitalization *)
```
pitag-ha, line 19: Isn't this the other way around?

lib/ocamlorg_web/graphql.ml
pitag-ha: @dinakajoy, have you seen this? :)

lib/ocamlorg_web/graphql.ml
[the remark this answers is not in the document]
pitag-ha: @dinakajoy, have you seen this? :)

lib/ocamlorg_web/graphql.ml
dinakajoy: Sorry, I forgot to implement it but it will be part of my next push

patricoferris:
@dinakajoy ping me when the two unresolved issues brought up by @pitag-ha are resolved and then I think we can merge

lib/ocamlorg_web/graphql.ml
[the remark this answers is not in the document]
pitag-ha: Have you pushed that change? :) In the PR, there still seem to be the [two filters](https://github.com/ocaml/v3.ocaml.org-server/pull/60/files#diff-7b9c15d336b381008174d3a481ce2b8f7812523a2ee852caf532f4fa332857feR42-R48): first the [starts_with filter](https://github.com/ocaml/v3.ocaml.org-server/pull/60/files#diff-7b9c15d336b381008174d3a481ce2b8f7812523a2ee852caf532f4fa332857feR44) and then the [offset/limit filter](https://github.com/ocaml/v3.ocaml.org-server/pull/60/files#diff-7b9c15d336b381008174d3a481ce2b8f7812523a2ee852caf532f4fa332857feR42).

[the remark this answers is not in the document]
dinakajoy: Hi @pitag-ha 
It was like this initially 
```List.filter
        (fun package ->
          starts_with filter (Package.Name.to_string (Package.name package)))
        packages
    in
    let packages =
      List.filteri (fun i _ -> offset <= i && i < offset + limit) packages
    in```
   
which was merged as this 
```List.filteri
        (fun i _ -> offset <= i && i < offset + limit)
        (List.filter
           (fun package ->
             starts_with filter (Package.Name.to_string (Package.name package)))
           packages)
    in```
    
    Maybe I misunderstood the merge you specified. Can you please clarify it a little bit more for me. Thank you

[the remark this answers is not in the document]
pitag-ha: Ok, that was a misunderstanding then. What I meant was only to filter only once as described in my comment https://github.com/ocaml/v3.ocaml.org-server/pull/60#discussion_r681833259. You're still filtering twice: once using filter and once using filteri.

Maybe have a quick look if you are able to implement this suggestion and if not I'd say to merge this PR as is and maybe to implement this suggestion in a separate (very short) PR.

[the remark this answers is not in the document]
pitag-ha: Perfect, thanks for implementing the change!

patricoferris (approved):

PR #69, opened 2021-07-30
Update setup-ocaml to use opam.2.1.0

tmattio:
Follow up on #68 to update opam 2.1.0


tmattio:
Merging now, the action repository should be replaced once https://github.com/ocaml/setup-ocaml/pull/184 is merged.

PR #70, opened 2021-07-30
Serve site from fs

tmattio:
Updates the v3.ocaml.org static file serving to read the files from the filesystem instead of crunching them at compile time.

The path of the site dir can be set at runtime time with the `OCAMLORG_SITE_DIR` env variable.

cc @avsm

### DIFF
--- a/lib/ocamlorg_web/router.ml
+++ b/lib/ocamlorg_web/router.ml
@@ -1,21 +1,12 @@
-let v3_loader root path request =
-  let headers, path =
+let v3_loader root path =
+  let path =
     let fpath = Fpath.v path in
     if Fpath.is_dir_path fpath || not (Fpath.exists_ext fpath) then
-      (* charset is used internally in dream when setting this *)
-      ( [ "Content-Type", "text/html; charset=utf-8" ]
-      , Filename.concat path "index.html" )
+      Filename.concat path "index.html"
     else
-      Dream.mime_lookup path, path
+      path
   in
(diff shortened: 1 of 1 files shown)


lib/ocamlorg_web/router.ml
```diff
@@ -8,11 +8,14 @@ let v3_loader root path request =
...
-    Page_handler.not_found request
-  | Some asset ->
-    Dream.respond ~headers asset
+  let file = Filename.concat root path in
+  Lwt.catch
+    (fun () ->
+      Lwt_io.(with_file ~mode:Input file) (fun channel ->
+          let open Lwt.Syntax in
+          let* content = Lwt_io.read channel in
+          Dream.respond ~headers content))
```
avsm, line 17: Isn't `Dream.from_filesystem` better than constructing a manual response here?

tmattio: We need a custom handler because the requested path is not the path of the file we want to serve (i.e. we need to add an `index.html` if it's not there)

tmattio: (The IO code is taken from `Dream.from_filesystem`)

patricoferris:
This LGTM if it is ready to go (I'm guessing this is the last thing blocking having continuous deployment via ocurrent deployer?)

tmattio:
Thanks for the review @patricoferris!

I think it is, @avsm needed this when working on the deployment. I'll merge now.

PR #73, opened 2021-08-06
Add test for  first graphQL endpoint

### LINKED ISSUE #71: Add tests for graphql endpoint
Once #60 is merged, we will need to right some tests for it, we've decided to split the two given how young this repository is (and lack of testing there already is) so as to not block that PR.

dinakajoy:
Fixes #71

### DIFF
--- a/lib/ocamlorg_web/graphql.mli
+++ b/lib/ocamlorg_web/graphql.mli
@@ -11,12 +11,12 @@ type packages_result =
   }
 
 val starts_with : string -> string -> bool
-(** This function compares two strings and returns true if the first string is
-    equal to the beginning of the second string up to capitalization *)
+(** [starts_with] function compares two strings and returns true if the first
+    string is equal to the beginning of the second string up to capitalization *)
 
 val is_package : string -> string -> bool
-(** This function compares two strings and returns true if they are equal up to
-    capitalization *)
+(** [is_package] function compares two strings and returns true if they are
+    equal up to capitalization *)
 
 val get_packages_result
   :  int
@@ -25,14 +25,21 @@ val get_packages_result
   -> string option
   -> Package.t list
   -> packages_result
-(** This function returns a list of the latest version of all packages *)
+(** [get_packages_result total_packages offset limit startswith packages]
+    returns a list of the latest version of all packages with [limit] and
+    [offset] options which is used to cut out some parts of the
+    [packages_result] list. It also takes a [startswith] option which will
+    search the list of packages looking for the package with a name that begins
+    with [startswith] *)
 
 val get_single_package : Package.t list -> string -> Package.t option
-(** This function returns the latest version of the pakage name specified *)
+(** [get_single_package packages name] will search the list of packages looking
+    for the package with a name that matches [name], [None] is returned if no
+    package can be found. *)
 
 val get_info : (Package.Name.t * string option) list -> package_info list
-(** This function returns the list part of Package.Info such as
-    Package.Info.dependencies *)
+(** [get_info] function loops through the list part of Package.Info such as
+    Package.Info.dependencies and returns a list of name and constraints record *)
 
 val schema : Package.state -> Dream.request Graphql_lwt.Schema.schema
 (** This schema allows to query for opam packages from the opam-repository as
--- a/test/ocamlorg_web/graphql_test.ml
+++ b/test/ocamlorg_web/graphql_test.ml
@@ -121,10 +121,16 @@ let get_info_test () =
   let num_of_deps = List.length deps in
   Alcotest.(check int) "returns 6 dependencies" 6 num_of_deps
 
-let get_single_package_test name cond () =
+let get_single_package_test name () =
   let pkg = Ocamlorg_web.Graphql.get_single_package packages name in
-  let res = match pkg with None -> false | Some _ -> true in
-  Alcotest.(check bool) "ocaml exist in package list" cond res
+  match pkg with
+  | None ->
+    Alcotest.fail "Expected a package"
+  | Some package ->
+    Alcotest.(check string)
+      "Same name"
+      name
+      (Package.Name.to_string (Package.name package))
 
 let () =
   Alcotest.run
@@ -169,12 +175,6 @@ let () =
       , [ Alcotest.test_case
             "returns true"
             `Quick
-            (get_single_package_test "ocaml" true)
-        ] )
-    ; ( "test to get a single package by name that does not exist"
-      , [ Alcotest.test_case
-            "returns false"
-            `Quick
-            (get_single_package_test "ban" false)
+            (get_single_package_test "ocaml")
         ] )
     ]


dinakajoy:
Thanks @patricoferris, please review

test/ocamlorg_web/graphql_test.ml
```diff
@@ -0,0 +1,201 @@
...
+  let pkg =
+    List.find_opt
+      (fun package ->
+        Ocamlorg_web.Graphql.is_package
+          name
+          (Package.Name.to_string (Package.name package)))
+      packages
+  in
+  let res = match pkg with None -> false | Some _ -> true in
+  Alcotest.(check bool) "ocaml exist in package list" cond res
```
pitag-ha, line 148: You've copied [the resolve function](https://github.com/ocaml/v3.ocaml.org-server/pull/60/files#diff-7b9c15d336b381008174d3a481ce2b8f7812523a2ee852caf532f4fa332857feR290-R293) for the query by package name here. The problem about copying code from your implementation in your test is that if there's a problem in the logic of your implementation, you won't capture that in the test.

Since you're already testing the function `Ocamlorg_web.Graphql.is_package` above, I assume that here you'd like to test the whole [resolve function](https://github.com/ocaml/v3.ocaml.org-server/pull/60/files#diff-7b9c15d336b381008174d3a481ce2b8f7812523a2ee852caf532f4fa332857feR288-R293) for the query by package name. Is that correct? If so, I'd suggest that you extract the [logic of the resolve function](https://github.com/ocaml/v3.ocaml.org-server/pull/60/files#diff-7b9c15d336b381008174d3a481ce2b8f7812523a2ee852caf532f4fa332857feR290-R293) in `graphql.ml` into a separate function and expose that one in `graphql.mli`. Then you can call that function directly here on your mock data `packages` and on your test parameter `name`. What do you think?

test/ocamlorg_web/graphql_test.ml
```diff
@@ -0,0 +1,201 @@
...
+
+type info_result =
+  { name : string
+  ; constraints : string option
+  }
+
+type packages_result =
+  { total_packages : int
+  ; packages : Package.t list
+  }
```
pitag-ha, line 15: What do you need these types for?

pitag-ha (commented):
Thanks,@dinakajoy , it's great to have some tests!

test/ocamlorg_web/graphql_test.ml
```diff
@@ -0,0 +1,201 @@
...
+  let all_packages =
+    Ocamlorg_web.Graphql.get_packages_result
+      total_packages
+      offset
+      limit
+      filter
+      packages
+  in
+  let num_of_pakgs = List.length all_packages.packages in
+  Alcotest.(check int) "returns 3 packages" 3 num_of_pakgs
```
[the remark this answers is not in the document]
dinakajoy: It returns returns graphql error
```{
  "errors": [
    {
      "message": "Line 12 col 1: Syntax error"
    }
  ]
}

File: test/ocamlorg_web/graphql_test.ml [hunk 4]
```diff
@@ -0,0 +1,201 @@
...
+
+let get_deps_test () =
+  let deps = Ocamlorg_web.Graphql.get_info dependencies in
+  let num_of_deps = List.length deps in
+  Alcotest.(check int) "returns 6 dependencies" 6 num_of_deps
```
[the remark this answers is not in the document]
dinakajoy: It is testing the the get_info function in graphql.ml file.
The goal was to test all small functions independently, hence the reason we have get_deps_test renamed to get_info_test, start_with_test, is_package_test and  get_single_package_test

test/ocamlorg_web/graphql_test.ml
dinakajoy: Done, thank you for pointing that out

test/ocamlorg_web/graphql_test.ml
dinakajoy: Wow, turns out they were not necessary. I added them to match the types with graphql.ml. It's been removed, thank you

patricoferris:
Both your initial GraphQL API and the "initialise opam state with a function" PRs are now merged so your tests should fit on top of `main` now :))

test/ocamlorg_web/graphql_test.ml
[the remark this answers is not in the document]
pitag-ha: So do I understand correctly that if you enter an `offset` or a `limit` that can't be met, all packages are returned? If so, might it make more sense to return an error in that case?

Also, what do you think about the first paragraph I wrote?

test/ocamlorg_web/graphql_test.ml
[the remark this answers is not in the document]
pitag-ha: You seem to only test that `get_info` preserves the length of the list. Is that what you want to test?

test/ocamlorg_web/graphql_test.ml
pitag-ha: Thanks!

lib/ocamlorg_web/graphql.mli
```diff
@@ -11,23 +11,27 @@ type packages_result =
...
-    equal to the beginning of the second string up to capitalization *)
-
-val is_package : string -> string -> bool
 (** This function compares two strings and returns true if they are equal up to
     capitalization *)
 
+val is_package : string -> string -> bool
+(** This function compares two strings and returns true if the first string is
+    equal to the beginning of the second string up to capitalization *)
+
```
pitag-ha, line 20: I think the comments should stay the other way around. You fixed that in the other PR and here you seem to go back to having the comment of `start_with` for `is_package` and vice-versa again. Maybe a mix-up during the rebase?

dinakajoy: Didn't see that, sorry.
It's fixed, thank you

lib/ocamlorg_web/graphql.mli
```diff
@@ -25,9 +25,13 @@ val get_packages_result
   -> string option
   -> Package.t list
   -> packages_result
+(** This function returns a list of the latest version of all packages *)
+
+val get_single_package : Package.t list -> string -> Package.t option
+(** This function returns the latest version of the pakage name specified *)
```
patricoferris, line 31: This is just a suggestion (except the typo) but one way that I think leads to better documentation comments is to actually use the function in the comment, e.g.: 
```suggestion
(** [get_single_package packages name] will search the list of packages looking for the package with a name that matches [name], [None] is returned if no package can be found. *)
```

The `[]` in documentation comments will format the docs as code (like backticks in markdown). This is especially useful for functions with labelled arguments.

test/ocamlorg_web/graphql_test.ml
```diff
@@ -0,0 +1,180 @@
...
+  in
+  let num_of_pakgs = List.length all_packages.packages in
+  Alcotest.(check int) "returns 3 packages" limit num_of_pakgs
+
+let get_info_test () =
+  let deps = Ocamlorg_web.Graphql.get_info dependencies in
+  let num_of_deps = List.length deps in
+  Alcotest.(check int) "returns 6 dependencies" 6 num_of_deps
+
+let get_single_package_test name cond () =
```
patricoferris, line 124: This is good but I think we also want to check that we got the correct package too (not just any package). To do this thoroughly we would need a `Package.t testable` (this is what `Alcotest.check` takes as an argument). The easiest way to do this is to add a "pretty printing" function to `package.ml` (This is because alcotest has `Alcotest.of_pp : 'a Fmt.t -> 'a Alcotest.testable`). Something along the lines of:

```ocaml
let pp : t Fmt.t = fun ppf t -> 
  Fmt.pf ppf {| { 
    name = %s; 
    version = %s;
    info = %a; 
  |} (Name.to_string t.name) (Version.to_string t.version) Info.pp (Lazy.force t.info)
```

Where you would also need `Info.pp`... Alternatively we can at least check the name and version of the returned package to make sure they are correct.

lib/ocamlorg_web/graphql.mli
```diff
@@ -25,9 +25,13 @@ val get_packages_result
   -> string option
   -> Package.t list
   -> packages_result
+(** This function returns a list of the latest version of all packages *)
```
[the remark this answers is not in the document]
patricoferris: Agreed, I think my suggestion below might help with this :))

patricoferris (commented):
Looking good! Just a few nit-picks :))

test/ocamlorg_web/graphql_test.ml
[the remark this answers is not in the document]
dinakajoy: > Also, what do you think about the first paragraph I wrote?
I have implemented that, thank you

> If so, might it make more sense to return an error in that case?
I will look into that

test/ocamlorg_web/graphql_test.ml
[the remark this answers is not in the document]
dinakajoy: > You seem to only test that get_info preserves the length of the list. Is that what you want to test?
Nope, the test is to ensure that get_info works fine. I just used the length to pass it to Alcotest as length. I did not know how to compare the data so I used length here

test/ocamlorg_web/graphql_test.ml
dinakajoy: I am a little bit confused on how to go about this please

test/ocamlorg_web/graphql_test.ml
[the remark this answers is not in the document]
pitag-ha: So far you've only used the most basic Alcotest testables like `int`, `string` and `bool`. However, it's also possible to construct more complex testables from those basic ones. For example, the return value of `get_info` is a list of a record called `package_info` containing a `string` and a `string option`. So you have two options here:
1. To construct a testable for the record `package_info`, you could do what [@patricoferris has explained in the context of `get_single_package_test`](https://github.com/ocaml/v3.ocaml.org-server/pull/73#discussion_r688377673): make it a custom testable by defining a pretty print function for it. Then you can use [Alcotest.list](https://mirage.github.io/alcotest/alcotest/Alcotest/index.html#val-list) to compare list of those records.
2. You can compare lists of pairs instead of comparing lists of `package_info` records. For that, you can simply also use the `Alcotest` testable constructors [Alcotest.pair](https://mirage.github.io/alcotest/alcotest/Alcotest/index.html#val-pair) and [Alcotest.option](https://mirage.github.io/alcotest/alcotest/Alcotest/index.html#val-list) additionally to [Alcotest.list](https://mirage.github.io/alcotest/alcotest/Alcotest/index.html#val-list) and write something like
```
Alcotest.(check (list (pair string (option string))))
```

Is that more or less understandable?

test/ocamlorg_web/graphql_test.ml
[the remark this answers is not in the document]
pitag-ha: What do you think about the suggestion to make both `offset` and `limit` parameters for this test function and that way run this test several times for different test data, for example also for some `offset > 0`?

lib/ocamlorg_web/graphql.mli
```diff
@@ -25,10 +25,21 @@ val get_packages_result
   -> string option
   -> Package.t list
   -> packages_result
+(** [get_packages_result total_packages offset limit startswith packages]
```
patricoferris, line 28: One last comment and this looks good to go -- to me it seems a little odd that we're passing in `total_packages` into this function. I don't think it is used in any way right, just passed directly to the record? 

Unless I'm mistaken (please tell me if I am) everywhere that passes in `total_packages` is just computing `List.length packages` e.g. https://github.com/ocaml/v3.ocaml.org-server/blob/b23e68833617ec80df6b4fe5911f7863af1217d9/lib/ocamlorg_web/graphql.ml#L276

Could this not be done inside the function so the user of the API doesn't have to do that work ? Maybe I'm missing something?

patricoferris (commented):
Just one last comment I think :))

lib/ocamlorg_web/graphql.mli
dinakajoy: It was used to match the limit 
```
let limit =
  match limit with None -> total_packages | Some limit -> limit
in

patricoferris: Ah okay makes sense 👍  The only comment I'd have about it is that we have some untested logic then. Your test that mimics what goes on here https://github.com/ocaml/v3.ocaml.org-server/pull/73/files#diff-c55503669a514a15d5842bfbe254058410c19d81d5fca26dc13dbd0545cb1030R103-R114 only captures some of the logic that is used in the resolver for example the following aren't tested:
 - What if the `limit` is `None`?
 - What if the offset is not `0`?
 - Are the correct `3` packages returned here ?
 
 I'm happy to merge this and have this be a follow-up PR/issue, let me know your thoughts :)) ?

dinakajoy:
Hi @patricoferris please merge this. We will continue with the test update in the next push

PR #74, opened 2021-08-06
make opam-repo state initialisation explicit

patricoferris:
This PR makes the opam-repository state initialisation explicit instead of happening at the `Package` module's startup-time. It's a little more verbose but we were running into problems in #73 where we needed some things from `Package` and so we were kicking off the opam clone/update. Now you very explicitly have to call `Package.init` to get the `state` and then pass this to all the functions that need it. The reason for passing it around is that if we still allow those functions to access the global state (the `s` variable) we could quite easily forget to initialise the state.

The `?disable_polling` parameter is probably not needed because in the graphql tests we simply won't be calling `init` at any point. @dinakajoy it might be worth trying to write the tests on top of this PR to see if this helps.

cc: @pitag-ha who has been helping tackle this problem :))

### DIFF
--- a/lib/ocamlorg/package.ml
+++ b/lib/ocamlorg/package.ml
@@ -171,44 +171,43 @@ let poll_for_opam_packages ~polling v =
   in
   updater ()
 
-let t =
-  let v =
-    (* Update at startup time if the directory exists so that the first page we
-       query does not contain *)
-    if Sys.file_exists (Fpath.to_string Config.opam_repository_path) then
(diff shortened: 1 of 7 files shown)


patricoferris:
@dinakajoy just rebased this to account for your now merged Graphql API -- does this look okay to you ?

dinakajoy:
Sure, thank you

dinakajoy (approved):
Sure, it looks good. Thank you

PR #77, opened 2021-08-19
add state from package list function

patricoferris:
This PR adds a function to generate `Package.state` (i.e. the opam-repository state as a map) from a list of packages. This should be useful when testing code that requires state (like the GraphQL tests) as we can now build our own list of packages and produce "state" from it rather than relying on the opam-repository cloning and building (i.e. what is used in production).


dinakajoy (approved):
Yea, great work.

PR #84, opened 2021-09-01
Add initial i18n support

tmattio:
This PR adds initial support for internationalization using [`ocaml-gettext`](https://github.com/gildor478/ocaml-gettext/).

We provide translation files (using gettext's format, a very common tool for translating content) in `gettext/`:

```
gettext
├── default.pot
├── dune
├── en
│   └── LC_MESSAGES
│       └── default.po
└── fr
    └── LC_MESSAGES
        └── default.po

4 directories, 4 files
```

The files are compiled to `.mo` files using `ocaml-gettext` at compile time. `ocaml-gettext` is then used at runtime to read the `.mo` files and provide translations for the content.

We add a middleware that reads the `Accept-Language` from the HTTP requests and sets `ocaml-gettext`'s global state to the detected language. This ensures that calls to `s_` and other translation functions will return the content in the desired language.

`ocaml-gettext` did not allow configuring the language at runtime and lacked in flexibility, so this PR depends on a fork of `ocaml-gettext` (https://github.com/tmattio/ocaml-gettext), we will probably want to upstream this before merging.

This is a very first draft that only provides translation for static content. Further work will be needed before this is ready to merge:

- [ ] Handle user preferences with cookies, so users can override their browsers settings if needed (this will be used in the frontend to provide a language selector)
- [x] Serve some translated content to demonstrate the end-to-end workflow with ood (we can use the success stories)
- [x] Handle translation in production (the PR only works in development for now)

We will probably also want to remove the introduced dependency on the filesystem and load the `.mo` file from memory. `ocaml-gettext` does not support this at the moment, but if we are satisfied with the user setup, I will work on adding support for it.

Once we are satisfied with the workflow in this PR, we will also need to adapt the frontend to mimic the new behavior. One noticeable difference in the current state of the PR is that we do not use redirections to serve the translated content: the same URL can be served with different languages.


kanishka-work:
I like it, but it seems to go in the opposite direction of strongly typed ood, ood-gen, and defining records for subsets of messages that will be fed to a given page, rather than using maps and assuming all values will be strings. I don't have an opinion on which is better. I think a good argument could be made for either approach.

tmattio:
I've updated the PR to use the yet-to-publish `gettext` library [here](https://github.com/tmattio/gettext). This allows us to remove the dependency on the filesystem and to use the `po` files directly instead of having to compile them to `mo` files. Overall, the setup should be a bit lighter and intuitive.

The next step will be to demonstrate the integration with ood by serving some french content.

tmattio:
> I like it, but it seems to go in the opposite direction of strongly typed ood, ood-gen, and defining records for subsets of messages that will be fed to a given page, rather than using maps and assuming all values will be strings.

@kanishka-work I'm not sure I understand. The solution proposed here is complementary to ood. Ood provides only the content, and will, in terms, provide variants of this content in several languages. The proposed changes here allow us to:

- Know which language to serve from ood (by parsing the `Accept-Language` HTTP header)
- Translate the static content hardcoded in the HTML pages (e.g. the navigation menu)

So it doesn't change what we had planned for ood. Or did you mean something else?

kanishka-work:
It will take some time to explain fully. It is not worth slowing down your momentum, and having an implementation is more important than synchronizing on this note.

tmattio:
I think this is ready to review. The PR now brings the following changes:

- I18n infrastructure with po files
- French translation for all the static content (navigation bar, footer, package landing page, etc)
- Gettext library largely based on `ocaml-gettext`, but with a more user-friendly workflow and API. The library is in the repo for now, but we can extract it whenever we feel it is stable enough.
- Success stories page to demonstrate the setup with ood (the design of the page is not important, this is just a demo)

I am not 100% convinced letting the users chose their language from the UI is critical. Most of the websites I've visited only made use of the `Accept-Language` header and seemed happy with this, so I'll leave this last item out of the PR.

PR #86, opened 2021-09-01
Global navigation bar

TheLortex:
Supersedes https://github.com/ocaml/v3.ocaml.org-server/pull/48

Now that voodoo generates a `package.json` file for each project containing the module map, we can implement a global navigation bar.

The first commit parses the `package.json`, the second commit use the data structure to generate a navigation bar in the side panel along with the table of contents.

This is a first iteration as a proof of concept, work on the design is needed. Moreover I had to remove the `fixed` class because the two sidebars were overlapping and were too long in any case (`fixed` disable scrolling).

Here is how it looks for the index page

Screenshot 2021-09-01 at 18-27-28 Documentation · base v0 14 1 · OCaml Packages

Unfolded to the current module:

Screenshot 2021-09-01 at 18-27-02 Base Binary_searchable Make T · base v0 14 1 · OCaml Packages


lib/ocamlorg/module_map.ml
```diff
@@ -0,0 +1,130 @@
+module StringMap = Map.Make (String)
```
tmattio, line 1:
```suggestion
module String_map = Map.Make (String)
```

To keep the current conventions

lib/ocamlorg/module_map.ml
```diff
@@ -0,0 +1,130 @@
...
+
+type kind =
+  | Module
+  | Page
+  | LeafPage
+  | ModuleType
+  | Argument
+  | Class
+  | ClassType
+  | File
```
tmattio, line 11:
```suggestion
  | Module
  | Page
  | Leaf_page
  | Module_type
  | Argument
  | Class
  | Class_type
  | File
```

Same as above, to stay in line with the conventions.

lib/ocamlorg/module_map.mli
```diff
@@ -0,0 +1,35 @@
+module StringMap : Map.S with type key = string
```
tmattio, line 1: The interface could use some documentation 🙂

lib/ocamlorg/module_map.mli
```diff
@@ -0,0 +1,35 @@
...
+  | Module
+  | Page
+  | LeafPage
+  | ModuleType
+  | Argument
+  | Class
+  | ClassType
+  | File
+
+module Module : sig
```
tmattio, line 13: Will we use the `Module` module here? If so, it would make more sense to me to have the module `Module` contain a module `Map`. What do you think?

tmattio (commented):
Thanks a lot for your work on this!

This LGTM overall - with some documentation for the added functions/modules, this is good to go I think!

lib/ocamlorg/module_map.ml
jonludlam: FWIW, it looks like this is matching odoc's conventions - see [here](https://github.com/ocaml/odoc/blob/91f6310967e64f3fa88445f3daf2fea2acc0bb49/src/document/url.ml#L80-L88)

lib/ocamlorg/module_map.mli
TheLortex: What do you mean ? Having something like this ?

``` ocaml
module Module : sig
  type t

  val name : t -> string

  val parent : t -> t option

  val submodules : t -> t StringMap.t

  val kind : t -> kind

  val path : t -> string

  module Map : sig
    type library =
      { name : string
      ; dependencies : string list
      ; modules : t StringMap.t
      }

    type t = { libraries : library StringMap.t }

    val of_yojson : Yojson.Safe.t -> t
  end
end

```

tmattio: Yes, that's what I meant

TheLortex:
@tmattio I've added a commit with the interface change. Personally I prefer the original version because the usage is more concise: the interface change requires to transform a lot of `Module_map.xxx` into `Module_map.Module.Map.xxx`.

tmattio:
> Personally I prefer the original version because the usage is more concise: the interface change requires to transform a lot of Module_map.xxx into Module_map.Module.Map.xxx

I was expecting the toplevel module to become `Module`, so calls to `Module_map.xxx` would become `Module.Map.xxx`. But if the approach doesn't work well in practice, feel free to revert back to `Module_map`, I meant it as a suggestion, it's not blocking in any way 🙂

TheLortex:
Ah yes okay I finally get it. I think the `Module` name is too generic and might refer to other things in the future. 
Here `Module.t` refers to the information we obtain from the module map (package.json). We could imagine other kinds `Module.t` (for example data extracted from an odoc file) so it makes sense to wrap it in the context of `Module_map`.

tmattio:
LGTM, thanks @TheLortex, this is going to improve the navigation of packages documentation a lot!

PR #91, opened 2021-09-14
Merge ood

tmattio:
This is the first PR that aims at simplifying our development stack and offer a more consistent and intuitive contribution experience to new contributors.

Here we merge the ocaml/ood repository:

- The `/data/` directory containing the data remains at `/data`
- The `/src/ood` directory containing the generated modules is now found at `/src/ocamlorg_data`
- The `/src/ood-gen` directory containing the code to generate the static modules is now found at `/tool/ood-gen`
- The `/src/ood-preview` HTTP server with the demo pages is not merged in the server and the pages are exposed at the `/preview` endpoint.
- The metadata files have been manually merged and reworked to offer a consistent development setup.

@patricoferris As you can see, I renamed the `ood` public library to `ocamlorg-data`. The goal is to make it obvious what this library is without knowing the accronym. Note that the library still exposes an `Ood` toplevel-module. Let me know what you think, I can easily revert this part.


patricoferris:
Reviewing this now, just as a quick comment renaming `ood` is completely fine! It started out life as a bit of prototype and then eventually got used so if you want to take this opportunity to rename it to `ocamlorg_data` (even as the exposed API part) for consistency, feel free to do so :))

src/ocamlorg_data/media/dune
```diff
@@ -0,0 +1,22 @@
+(library
+ (name ood_media)
+ (public_name ocamlorg-data.media))
+
+(rule
```
patricoferris: Do we still need this, iirc correctly this was used in the preview but now that's integrated we will just be using the actual media directory? The only reason I ask is that soon this file will exceed the 100MB limit (when ocaml.org-media is brought in for example).

src/ocamlorg_web/lib/templates/previews/dune
```diff
@@ -0,0 +1,77 @@
+(rule
```
patricoferris, line 1: Minor (feel free to ignore): I think we can avoid all of the duplication here with a single rule

```
(rule
 (targets preview_books_template.ml preview_papers_template.ml ...)
 (deps preview_books_template.eml preview_papers_template.ml ...)
 (action
  (run %{bin:dream_eml} %{deps} --workspace %{workspace_root})))
```

Just dawned on me whilst reviewing.

data/tutorials/en/020_streams.md
```diff
@@ -0,0 +1,736 @@
...
+    val compare : t -> t -> int
+    val equal : t -> t -> bool
+    val subset : t -> t -> bool
+    val iter : (elt -> unit) -> t -> unit
+    val map : (elt -> elt) -> t -> t
+    val fold : (elt -> 'a -> 'a) -> t -> 'a -> 'a
+    val for_all : (elt -> bool) -> t -> bool
+    val exists : (elt -> bool) -> t -> bool
+    val filter : (elt -> bool) -> t -> t
+    val filter_map : (elt -> elt option) -> t -> t
```
patricoferris, line 648: Was just going to say feel free to bump us up to `4.12.0` :)) 👍

Makefile
```diff
@@ -3,13 +3,30 @@
 ARGS := $(wordlist 2,$(words $(MAKECMDGOALS)),$(MAKECMDGOALS))
 $(eval $(ARGS):;@:)
 
+# Dependencies used in development, for ood-gen in particular
```
patricoferris, line 6: Would it not be easier to track this in an opam file perhaps and `pin -yn` and `install --deps-only`?

Makefile
```diff
@@ -54,12 +71,26 @@ fmt: ## Format the codebase with ocamlformat
...
 
 .PHONY: utop
 utop: ## Run a REPL and link with the project's libraries
-	opam exec -- dune utop --root . lib -- -implicit-bindings
+	opam exec -- dune utop --root . . -- -implicit-bindings
 
-.PHONY: update-site
-update-site: ## Update the site directory
-	rm -rf asset_site; ./script/update-site.sh
+.PHONY: utop
```
patricoferris: Should it be `.PHONY: gen-data` ?

patricoferris (approved):
Thanks @tmattio I'm more convinced that this will make life easier. In the future if we have *a lot* of updates heading into what used to be called `ood` we might need to rethink things as this repo would get very busy but until then having things colocated will help speed things up :))

Just a few small comments that can mostly be ignored if you think they're unimportant.

src/ocamlorg_web/lib/templates/previews/dune
tmattio: Unfortunately, the Dream CLI does not accept multiple files as input, so we have to run one action for each of the files.

src/ocamlorg_data/media/dune
tmattio: Nope, I'll remove it

Makefile
tmattio: Done!

Makefile
tmattio: `opam install --deps-only .` still works on the repo, but installs only what you need to run the server, which made sense for me (if you don't work on the data, you don't need to install anything else)
In development, you can set up your env with `make switch` which will install the dev dependencies for you.
I've copied this workflow from Dune's repo, but I'm not sure what's best.

src/ocamlorg_web/lib/templates/previews/dune
patricoferris: I think it does https://github.com/aantron/dream/blob/master/src/eml/main.ml#L40 ? It works locally for me (`dream.1.0.0~alpha2`). I think the `--help` says `FILE` when it should probably say `FILES`

avsm:
Looks good. Could we also have a companion PR to ocaml/ood to mark it as deprecated in the README and point to this repo as the canonical one? I can then mark ocaml/ood as read only and deprecated.

patricoferris:
https://github.com/ocaml/ood/pull/89 is merged pointing to this repository

PR #101, opened 2021-09-17
Server pages from the server

tmattio:
This PR aligns the various HTML generation technologies used in the v3.ocaml.org site into a single server and monorepo.  Previously, we had four different stacks while pulling together the unified site design: 
- ReScript/NextJS for static HTML
- odoc+Tailwind output for static docs
- opam2web for static package information
- Dream+styling for the documentation browser

As we move towards a more feature-complete v3 site, all of these are now consolidated under the Dream-based web server, with a single set of OCaml source code that provides the templating logic for all portions of the site. Previously, the Tailwind logic was spread among multiple tools (odoc/opam2web/nextjs) and languages (ReScript/OCaml) and difficult to keep unified across the site.

There are several features that would benefit from ReScript (primarily those that require complex interactivity such as playgrounds) which we would like to revisit in v4 of the site. We've taken the decision to consolidate around OCaml for this iteration of the site, and come back to the more ambitious plans next year once v3 goes live.  This PR maintains all of the commit histories of the ReScript site and bindings so that it can be brought back later.


avsm (approved):
LGTM

PR #103, opened 2021-09-21
Misc fixes and improvements

tmattio:
A laundry list of fixes and improvements while I was working on the frontend:

- Add redirections from V2 URLs to fix links from the tutorials (the long term fix will be to review the links from the tutorials, but in the meantime, it's a working patch)
- Simplify URLs. The URL structure does not have to match the sitemap. Prefixing all the pages by category seems unnecessary and convoluted.
- Add rules to promote the ocamlorg_data module and change the dune setup to ignore the promoted rules in opam and CI
- Fix the overflow-y in the package navbar
- Add binaries to extract translatable strings from OCaml sources and merge them with existing PO files
- Fix not found page HTTP status (previously returned HTTP code 200)
- Improve start and watch make commands to use a dune rule so dune can manage the filesystem changes instead of a custom script
- Use 90rem for the max-width instead of 80. This gives much more room to the actual content (tutorials and documentation).


patricoferris (approved):
LGTM, most of the changes are just coming from ood-gen/formatting. Checked it out locally and fixes the package search/translation bug (if that was a thing).

tmattio:
> Checked it out locally and fixes the package search/translation bug (if that was a thing).

Yes, that's what triggered me to write some tools to extract the translatable strings, the bug was a missing string in the translation table.

PR #106, opened 2021-09-22
Homepage toplevel

tmattio:
This is reviving the toplevel experimentation to provide a toplevel in the homepage.

image

Of course, the design will have to be reworked, this is just a demo.

The biggest problem is the size of the bundle. The toplevel with just Js_of_ocaml and Stdlib is 20M, and the one with Base is 30M, I'm not seeing any way around it, since we have to build with `-linkall`. Any insights on this?


src/ocamlorg_toplevel/bin/dune
```diff
@@ -0,0 +1,73 @@
...
+(subdir
+ js/
+ (rule
+  (alias toplevel)
+  (targets stdlib-4.12.0.js base-v0.14.1.js)
+  (deps ../toplevel_stdlib.js ../toplevel_base.js)
+  (action
+   (progn
+    (copy ../toplevel_stdlib.js stdlib-4.12.0.js)
+    (copy ../toplevel_base.js base-v0.14.1.js)))))
```
patricoferris: Perhaps there's some way to make the jsoo compiler do this automatically? At any rate currently this seems to half the size of the files (uncompressed). 

```suggestion
    (run jsoo_minify ../toplevel_stdlib.js -o stdlib-4.12.0.js)
    (run jsoo_minify ../toplevel_base.js -o base-v0.14.1.js)))))
```

patricoferris (approved):
Cool, thanks! As you mentioned needs some design improvements (I think the text is a little small) but with the suggested change this brings the uncompressed size down to about 4M I think which is a little more acceptable once `Dream_encoding` gets the hold of it :))

TheLortex:
I'd suggest to lazily load the top level when the user requests it. 
4M is still a lot for mobile browsing or from places with low-bandwidth networking ! For a landing page I believe it's problematic

Issue #105, opened 2021-09-21
Support code highlighting in tutorials


patricoferris:
Now that ood is within this repository we have a some more freedom to decide how to do this. I'm personally in favour of improving/rewriting (and probably vendoring) something like https://github.com/patricoferris/hilite and doing this statically. This let's us: 

 1. High-light anything vscode can (dune, opam, ocamlformat etc.) see https://github.com/ocamllabs/vscode-ocaml-platform/tree/master/syntaxes
 2. No runtime client-side highlighting via highlight js (less data sent to clients, less JS running on clients)
 3. Better control over our dependencies from a privacy standpoint too because of (2)

The cost of course is more maintenance. @tmattio any thoughts?

tmattio:
Related to https://github.com/ocaml/v3.ocaml.org-server/issues/99 - we probably want the same solution for both

tmattio:
Hilite seems very nice! I like that it allows us to do the highlighting from the server!

If it's still unreleased, feel free to add it in `src/` directly, the same way I did with `gettext`. Ultimately, I'd like to release it independently, but I'll come back to it when it is stable in ocaml.org.

ghost:
> No runtime client-side highlighting via highlight js

From my limited understanding, I don't think we have ever implemented code highlighting in the browser. In prior and proposed implementations, it was run as a final phase of markdown processing at build time.

patricoferris:
Initial work done in #108

PR #108, opened 2021-09-22
add initial server-side code-highlighting

patricoferris:
This PR adds some static code-highlighting (i.e. not done on the client-side with highlight.js or prism.js) as part of #105. The library (`hilite`) reuses the textmate grammars from vscode with some minor modifications because the textmate library wasn't happy parsing them. 

The colours are of course chosen pretty randomly and I'm happy for anyone to change them. One small issue is most OCaml blocks are formatted like a top-level, we will need some additional logic (hopefully not in this PR) to un-highlight the output, this is especially problematic for compiler errors (see image 3).

**Image 1 - Toplevel OCaml that looks okay (does leak into function type)**

Screenshot 2021-09-22 at 16 32 13

**Image 2 - Dune file highlighted**

Screenshot 2021-09-22 at 16 32 40

**Image 3 - Toplevel looks a bit odd**

Screenshot 2021-09-22 at 16 44 58


tmattio (approved):
I'm really impressed by the footprint of the library: i.e. a single call to `Hilite.Md.transform`!

I didn't review the actual library, but everything else LGTM. I would maybe inline the CSS code for the syntax to avoid another roundtrip, but that's a detail.

Feel free to merge whenever the tests are passing 🎉

patricoferris:
Thanks for the review :)) What's the best way to pull in the one dependency of `Hilite` that doesn't belong to any of the opam files in the repo? I've added it to the `Makefile` as a dev dependency but we don't install this in CI ?

tmattio:
> Thanks for the review :)) What's the best way to pull in the one dependency of Hilite that doesn't belong to any of the opam files in the repo? I've added it to the Makefile as a dev dependency but we don't install this in CI ?

Ah, right, the tests will try to run everything under `src/`. A quick workaround would be to move `hilite` to `tool/` or `vendor/`.

If we do want to run the tests in the CI, I'd suggest installing Hilite's deps in the CI step with a comment, and we'll remove the step once it's released.

tmattio:
Or we can create a package `hilite` if you prefer.

patricoferris:
Thanks!

PR #109, opened 2021-09-22
initial rev-deps experiment

patricoferris:
This PR contains some code to calculate the reverse dependencies (mostly borrowed from opam2web and made to work with our code). It isn't necessarily going to live here but I thought to generate more ideas I would open a draft PR. Essentially there are the following problems:

 - Calculating reverse dependencies takes maybe an extra 20-30 seconds with this implementation.
 - It is a blocking operation at the time of writing (although I have ideas that we could maybe fork a process or something) so that time is very noticeable.
 - Calculating revdeps means holding all of opam-repo in memory, which is certainly doable it is more an issue 

because we lose the `lazy_t` code.

Some potential solutions are to:

 - Move this code elsewhere (an ocurrent pipeline or something similar) and access the data from some endpoint.
 - Serialise the `Info.t Version.Map.t Name.Map.t` into the filesystem as json and read (and subsequently update by writing too) those files.

Another issue this brought up (I think) is that we should calculate the "latest" package that matches the formula in the dependencies or reverse dependencies so that we link to that instead. For example https://v3.ocaml.org/p/alcotest-lwt/1.2.1 has a dependency of `alcotest =version` so clicking on that should lead to https://v3.ocaml.org/p/alcotest/1.2.1 not https://v3.ocaml.org/p/alcotest/.

*Note: once we know where this code is going to live, I'll definitely add a thanks to the original authors for the parts borrowed from opam2web*

**An image of the rev deps**

Screenshot 2021-09-22 at 17 47 21


tmattio:
> Calculating reverse dependencies takes maybe an extra 20-30 seconds with this implementation.

😱  Seems like we'll need to do this in the pipeline then, even a couple of seconds would be too expensive for every request, and I doubt we can get under that by simply optimizing the code. @jonludlam what do you think?

> Another issue this brought up (I think) is that we should calculate the "latest" package that matches the formula in the dependencies or reverse dependencies so that we link to that instead. For example https://v3.ocaml.org/p/alcotest-lwt/1.2.1 has a dependency of alcotest =version so clicking on that should lead to https://v3.ocaml.org/p/alcotest/1.2.1 not https://v3.ocaml.org/p/alcotest/.

Good point! Could we fix this in a separate PR?

patricoferris:
>😱 Seems like we'll need to do this in the pipeline then, even a couple of seconds would be too expensive for every request

Ah no, it is a one time cost (or an every time opam-repository updates) cost because these changes calculate everything at once at the beginning and then hold everything in memory. After the initial update it would probably be faster than the current way because it doesn't read the filesystem... I think...

tmattio:
> Ah no, it is a one time cost (or an every time opam-repository updates) cost because these changes calculate everything at once at the beginning and then hold everything in memory. After the initial update it would probably be faster than the current way because it doesn't read the filesystem... I think...

I see. How's the performance profile if we do this dynamically, reading only the opam files needed to calculate the rev-deps?
I remember we were loading all the files in memory for the deps, and then decided to do it dynamically, and it's been working well. The difference, of course, is that we'll read multiple files for each package (if they are not already in memory), which may still be too slow to do dynamically.

patricoferris:
Unless I'm mistaken (which really could be the case) I don't think there's really an easy way to do this dynamically (if at all). We can't know before-hand the set of packages that are a given package's reverse dependencies without having previously loaded in the data (in this case we need the `OpamPackage.Set.t OpamPackage.Map.t` map of packages to their dependencies and the only way to build this is by reading every single opam file).

To not hold everything in-memory (i.e. the `Info.t Version.Map.t Name.Map.t`) we could keep a separate data-structure (similar to `OpamPackage.Set.t OpamPackage.Map.t`) of just the reverse dependency information and when looking up a package we force the opam file read + `OpamPackage.Map.find` the rev deps. This doesn't stop us from having to read every opam file to build (and update) that set of rev deps though.

tmattio:
> We can't know before-hand the set of packages that are a given package's reverse dependencies without having previously loaded in the data

Duh... I don't know what I was thinking.

In this case, even if loading the rev-deps is a one-time cost, I'd still be inclined to offloading it to the doc pipeline. IIRC, the reason we moved away from loading all the packages at startup time to loading them dynamically is that the few seconds you had to wait when starting the server had a huge impact on the feedback loop (waiting for 10 seconds to have visual feedback on a change in the frontend is very frustrating). We could mock some data during the development, of course, but since this is a recurring problem (e.g. global navbar for package doc) and that we're solving it by doing the work in the pipeline, it seems sensible to do the same for the rev-deps.

patricoferris:
Hmm we could somewhat mitigate that by having the revdeps run in the background in a different process and writing to file that the server then reads but yeah moving things out might be easier. Although to provide better package search from the server (including things like the tags, description etc.) we're going to need to read everything into memory or at least do that once to do some clever indexing of all the packages no?

TheLortex:
The documentation pipeline already being quite a heavy machine machinery, I'd recommend against adding unrelated features into it. What I imagine is either a third service or a separate process, communicating with the server using RPCs (capnproto requires an external schema file but we can also choose to use `ocaml-rpc` where the schema is defined in OCaml). This service / process could survive to the main process being restarted and avoid the cost of loading the package list.

patricoferris:
Closing in favour of @TheLortex's great work in realising this in #134 🙌

PR #110, opened 2021-09-23
resync changelog

patricoferris:


CHANGES.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased
 
+- Add code-highlighting to tutorials (#108, by @patricoferris)
+- Add initial set of Best Known Methods (#107, by @tmattio)
+- Add initial toplevel to homepage (#106, by @tmattio)
+- Serve HTML pages from server (#101, by @tmattio)
```
tmattio, line 6: This is not a user-facing change, so I'm not sure it should be included in the changelog (same for the previous items "Merge frontend" and "Merge ood")

tmattio (approved):
Thanks @patricoferris! We could add a check in the pipeline for a changelog entry also. (e.g. https://github.com/ocamllabs/vscode-ocaml-platform/blob/master/.github/workflows/changelog.yml)

Not sure whether the non-user-facing changes should be in the changelog, but if you think they should be in there, this LGTM.

patricoferris:
@tmattio any idea why the check says https://github.com/ocaml/v3.ocaml.org-server/runs/3686040719#step:3:6 ?

tmattio:
Not sure... You might need to fetch the repo maybe?

patricoferris:
Hmm maybe, looks like vscode-platform has the same issue? https://github.com/ocamllabs/vscode-ocaml-platform/runs/3570923719?check_suite_focus=true#step:3:6

patricoferris:
Yeah you were right, the checkout only gets exactly the ref/commit it needs and nothing else by default it seems. `fetch-depth: 0` gets *everything*. I think this is fixed now (https://github.com/ocaml/v3.ocaml.org-server/runs/3698943717) you might want to upstream that to vscode @tmattio ?

PR #115, opened 2021-09-24
Add initial opportunity data and pages

tmattio:
This PR adds the infrastructure to support the opportunity (aka job) data, with the following YML format:

```yml
jobs:
  - title: <title of the job post>
    link: <link to the job post on the company website>
    location: <location>
    company: <name of the company>
    company_logo: <link to the company logo>
    fullfilled: <boolean>
    description: >
      <description in markdown>
```

The opportunities page lists the jobs posts in reverse order:

image

And the Opportunity detail page displays the description with a link to the post:

image


tool/ood-gen/lib/job.ml
```diff
@@ -0,0 +1,108 @@
+type metadata =
```
patricoferris, line 1: Do you think we need some kind of date here like `published` to make it easier to sort by most recently added ?

tool/ood-gen/lib/job.ml
```diff
@@ -0,0 +1,108 @@
...
+         (List.mapi
+            (fun i ->
+              Utils.decode_or_raise (fun x ->
+                  match metadata_of_yaml x with
+                  | Ok raw ->
+                    Ok
+                      { id = i
+                      ; title = raw.title
+                      ; link = raw.link
+                      ; description_html =
```
patricoferris, line 40: I think a description is good, but if everyone is quite as thorough as you have been, this yaml file is going to get pretty big, pretty quickly. Do you think we should slim it down a bit (or alternative go for a collections model à la Tutorials)?

patricoferris (commented):
It looks reallly good :))! Just a few questions/suggestions

tool/ood-gen/lib/job.ml
tmattio: That's a good point. I don't think it's useful for sorting the job posts (we can just take the order as-is if people add their listing at the beginning of the file), but it'd be very useful in the UI (e.g. add a publication date, add a "new" tag, etc.)!

tool/ood-gen/lib/job.ml
tmattio: I agree.. I didn't want to end up with a thousand files either. I was seeing two options:

- We don't include a description, so we can remove the opportunity detail page and simply redirect to the job post on the company page
- We can delete the job posts that are not available anymore, effectively keeping the file to a reasonable size.

patricoferris: Yeah, I think trimming down the description is the way to go then. This will also lower the barrier for people adding their job postings (maybe at some point we could scrape some from discuss.ocaml.org). I think it would be nice to keep a record of jobs (I know they'll be there in git... but you know what I mean)

tmattio: I was thinking of it a bit more and it seems better to decide which implementation to go with once we've agreed on what we want to go on the website first.
From a product-only perspective, I'd be inclined to have a job description on the website. FWIW, this is what the Python job board looks like: https://www.python.org/jobs/
Thinking of data analytics, it could also be useful to have job descriptions.
What do you think? If we agree that we should only redirect to job posts on the companies website, let's trim down the descriptions as you suggested, but if we'd like to have job descriptions, I'm sure we can come up with a satisfying solution.

tmattio:
Merging now, we can address the reviews in follow-up PRs. In particular, we can continue to discuss whether we want to include a description and I'll remove it if we decide that we don't want it.

PR #116, opened 2021-09-24
reorganise and complete the OCaml events

patricoferris:
I think I've got all the "events" from v2 now including workshops, some individual events and generic "meetups" (i.e. recurring events). There's a bit of duplication between the OUPS meetup and some events but I think that's okay. I may have _borrowed_ some styling from #115 😁  (the event page needs restyled but just getting the data and page in). 

**The meetups**

Screenshot 2021-09-24 at 17 46 46

**The events (the workshop table is unchanged)**

Screenshot 2021-09-24 at 17 46 59

### DIFF
--- a/src/ocamlorg_data/ood.mli
+++ b/src/ocamlorg_data/ood.mli
@@ -86,12 +86,12 @@ module Event : sig
   type t =
     { title : string
     ; slug : string
-    ; description : string
+    ; description : string option
+    ; organiser : string option
     ; url : string
     ; date : string
     ; tags : string list
     ; online : bool
-    ; textual_location : string option
     ; location : string option
     }
 
@@ -117,6 +117,25 @@ module Job : sig
   val all_not_fullfilled : t list
 
   val get_by_id : int -> t option
+end
+
+module Meetup : sig
+  type location =
+    { lat : float
+    ; long : float
+    }
+
+  type t =
+    { title : string
+    ; slug : string
+    ; url : string
+    ; textual_location : string
+    ; location : location
+    }
+
+  val all : t list
+
+  val get_by_slug : string -> t option
 end
 
 module Industrial_user : sig
--- a/src/ocamlorg_web/lib/templates/previews/preview_events_template.eml
+++ b/src/ocamlorg_web/lib/templates/previews/preview_events_template.eml
@@ -10,7 +10,7 @@ let render_event (event : Ood.Event.t) =
       </h1>
     </div>
     <div class="mt-6 prose prose-indigo prose-lg text-gray-500 mx-auto">
-      <%s! event.description %>
+      <%s! Option.value ~default:"" event.description %>
     </div>
   </div>
 
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Import all of the events data from v2 (#116, by @patricoferris)
+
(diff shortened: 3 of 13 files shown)


tmattio:
Nice!! Do you think we could also add the Mirage and Dune retreats (and others) to the events?

src/ocamlorg_data/ood.mli
```diff
@@ -100,6 +100,25 @@ module Event : sig
   val get_by_slug : string -> t option
 end
 
+module Meetup : sig
+  type location =
```
tmattio, line 123: We already have a location type no? We could refactor the module to use the same

src/ocamlorg_data/ood.mli
```diff
@@ -86,12 +86,12 @@ module Event : sig
...
     { title : string
     ; slug : string
-    ; description : string
+    ; description : string option
+    ; organiser : string option
     ; url : string
     ; date : string
     ; tags : string list
     ; online : bool
-    ; textual_location : string option
```
tmattio, line 94: We may not have the data for it now, but from what I can tell it still makes sense to have a textual location, no?

src/ocamlorg_web/lib/templates/previews/preview_events_template.eml
```diff
@@ -10,7 +10,7 @@ let render_event (event : Ood.Event.t) =
       </h1>
     </div>
     <div class="mt-6 prose prose-indigo prose-lg text-gray-500 mx-auto">
-      <%s! event.description %>
+      <%s! Option.value ~default:"" event.description %>
```
tmattio, line 13:
```suggestion
      <%s! Option.value ~default:"No description" event.description %>
```

patricoferris:
> Nice!! Do you think we could also add the Mirage and Dune retreats (and others) to the events?

Sounds good, these would be a "meetup" (that's not really what a meetup is currently) or do you mean the individual events ? (It might be quite nice because currently the events section shows... 2014 😁 )

tmattio:
> Sounds good, these would be a "meetup" (that's not really what a meetup is currently) or do you mean the individual events ? (It might be quite nice because currently the events section shows... 2014 😁 )

TBH, I'm not sure what's the difference between a meetup and an event. Are meetup exclusively events from [Meetup](https://www.meetup.com/)?

patricoferris:
> TBH, I'm not sure what's the difference between a meetup and an event. Are meetup exclusively events from Meetup?

I think there's a better distinction now that you have brought up the retreats, something like:
 - Recurring Events (currently meetups, but would contain retreats?)
 - Workshops
 - Events (these are single events with a specified date, maybe overlapping with recurring events maybe not. This list is useful if we want to display "Upcoming events" prominently somewhere)

What do you think?

tmattio:
> This list is useful if we want to display "Upcoming events" prominently somewhere

It'd be nice to have recurring events in this list as well maybe? If that's the case, I am not sure I see a good reason to keep them separated. Meetups are recurrent, but they have their own agendas and they often don't have a specific frequency, so in my mind, they really are a set of individual events.

tmattio (approved):
We can merge now and address the other comments in follow up PRs I think: the PR as-is is already a great improvement.

We can create issues to:

- Merge the meetups and events (if we agree that we want to do this)
- Add the retreats

patricoferris:
Thanks :))

PR #120, opened 2021-09-27
Compile with OCaml 4.13.0

tmattio:
This makes the project compatible with OCaml 4.13.0 and switches the CI to the latest compiler version.

Apart from the documentation and CI changes, this pins `js_of_ocaml` to master because the only version compatible with ocaml.4.13.0 is js_of_ocaml.3.10.0, but the toplevel generation is broken in this version. A patch has been merged in https://github.com/ocsigen/js_of_ocaml/pull/1131 and is available in `master`.

The pins can be removed once a new version of `js_of_ocaml` is released. cc @hhugo

@avsm This should fix the deployment failures you saw.


patricoferris:
I've just been working on this haha... I'm not convinced bumping to 4.13.0 is the answer here just yet. I think fixing the alpine image to `4.12.0` and bumping once the rest of the ecosystem is ready might be a better long-term strategy otherwise we'll potentially be doing this every time there's a new release. I'm not a fan of having the rug swept from under our feet whenever the latest docker image jumps to a new version and would rather make it an explicit PR to bump up? What do you think @tmattio ?

patricoferris:
A race condition! 😆

PR #121, opened 2021-09-27
Pin the Docker image version

tmattio:
Follow up of #120 that pins the docker image. Thanks for the suggestion @patricoferris


patricoferris (approved):
Thanks! I think this will keep the deployments building even with new releases (we can then be a bit more prepared when bumping OCaml version) :))

avsm:
Pinning the OCaml version is a good idea, and takes the pressure off #120 to be merged too soon. Although it is really nice that our release process is now good enough that we can upgrade to 4.13 in such short order!

Issue #119, opened 2021-09-27
Turbo drive conflicts with toplevel loading


tmattio:
When navigating to the homepage from another page, the top-level does not initialize properly. Presumably, this is caused by Tubo Drive loading of the page.

patricoferris:
Indeed. I did a bit of digging into this and the fix doesn't look simple... 

The current issue is using `window.onload` which doesn't happen once turbo gets a hold of things. Instead you can add an event listener to the `document` for [`"turbo:load"`](https://turbo.hotwired.dev/handbook/building#observing-navigation-events). So something like:

```ocaml
let set_turbo_load f = 
  let open Dom in
  addEventListener Dom_html.document (Event.make "turbo:load") (
    handler (fun _ ->
      (try 
        let _ : Dom_html.element Js.t = Dom_html.getElementById "toplevel-container" in 
        f ()
      with 
        | Not_found -> ());
      Js._false)) Js._false

let _ =
  let _id : Dom.event_listener_id = set_turbo_load run in 
  ()
```

But our `run` function is not idempotent so this isn't really a solution. Even with this solution when you navigate back some state is in a not so consistent state, you can then type `Ctrl+k` to reset the toplevel, it prints some errors but then seems to be back to normal. I'm not entirely sure what's going on to be honest. A short term solution might be just to disable turbo caching on the homepage?

tmattio:
> A short term solution might be just to disable turbo caching on the homepage?

I've been experiencing other minor issues with Turbo Drive on other pages. Maybe we can disable it altogether and revisit it when we're optimizing the performance of the website.

patricoferris:
What kind of things? JS related ones or something else? If it's causing more hassle atm then sure, or just disable it on the pages it's causing issues on?

tmattio:
I think this is fixed in #125, closing now.

PR #125, opened 2021-09-27
reload index page to keep toplevel functioning

patricoferris:
A simple workaround for #119, the reasoning behind the issues is explained more in this issue. If at all possible I would quite like to keep Turbo around because it is quite noticeable when it is not! But if there are too many little issues caused by it we can revisit it.


tmattio (approved):
Looks good, I'll investigate the other minor issues I mentioned in #119 separately.

PR #126, opened 2021-09-27
fix install link and 404 issues link

patricoferris:
Very minor fix, the `install` link was 404-ing which then made me notice the old link to the issues. All fixed hopefully :))


tmattio (approved):
Thanks!

Issue #53, opened 2021-07-29
'comment-delim' items are intended to be hidden


jonludlam:
see e.g. https://ocamlorg.herokuapp.com/p/tezos-base/8.0/doc/Tezos_base/Block_header/index.html#type-shell_header - specifically the `(*` and `*)` around the documentation of the record fields.

The idea is that you can cut-n-paste the record definitions and get valid code, but that they don't actually appear on-screen.

tmattio:
Fixed in #130

tmattio:
Reopening as the problem is fixed, but this introduces some regressions for long record fields / types. A proposed solution is to set a min-width and still wrap the record fields. cc @panglesd

tmattio:
Closing again as the regression has been fixed in #142

PR #130, opened 2021-10-04
Better formatting of odoc def tables, hiding comment-delim

panglesd:
This PR addresses issue #53. 

Example: 
image
is turned into:
image

Comment delimiters are hidden but still selectable (although the selection can miss them) so copy and paste yield valid code.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Unreleased
 
+- Better formatting of odoc def tables, hiding comment-delim (#130, by @panglesd)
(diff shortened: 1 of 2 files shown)


tmattio (commented):
Thanks @panglesd - it looks great! A small entry in the changelog and this is good to be merged 🙂

panglesd:
I updated the changelog with an entry for this PR!

tmattio:
Thanks a lot! 🎉

Issue #87, opened 2021-09-02
Package search should search package `tags`, `description`, ... by default


jmid:
First of all: thanks for your work on the new website! 😀 

I just gave the package search a spin. A bit of feedback: It seems the package search is only by name (or some substring).
I find that unfortunate from a UX point of view, as newcomers don't necessarily know the name of a relevant package beforehand (so they search).

Two examples:
- If I go to https://opam.ocaml.org/packages/ and search for `quickcheck` the result is 8 packages (all relevant).
  In comparison, https://v3.ocaml.org/packages/search?q=quickcheck only gives me 2 - which are both named something with `quickcheck`. 
- The story is the same for other search terms, e.g., `profiling`: https://opam.ocaml.org/packages/ shows me 6 packages
but https://v3.ocaml.org/packages/search?q=profiling shows 0 packages.

As such, newcomers may not find relevant packages or (even worse) think that there are few/no relevant OCaml packages available!
To improve the situation, I think searching additional opam package fields (`tags`, `description`, `synopsis`, ...)
should help. I would even argue that this should be the default.

Related search issues: #11 #38  #44

jmid:
OK, I dug bit. The https://opam.ocaml.org/packages/ search seems to end here:
https://github.com/ocaml-opam/opam2web/blob/23366c29a505b96e9456cd4a1bf5027523a6a206/src/js_search/search.ml#L60
```ocaml
    if matches str name || matches str descr
    then show tr
    else hide tr
```
which searches the package description as well (client-side).

In comparison https://v3.ocaml.org/packages/search seems to search only package names (server-side):
https://github.com/ocaml/v3.ocaml.org-server/blob/7364ee120232e3ee7db8873fae232d95ae5a536a/lib/ocamlorg/package.ml#L379-L385

Modulo some forcing of the `Lazy.t` extraction of individual package info, I think something like the following could work:
```ocaml
  all_packages_latest ()
  |> List.filter (fun { name; info; _ } ->
         if contains (String.lowercase_ascii @@ Name.to_string name) pattern
            || contains (String.lowercase_ascii info.synopsis) pattern
            || contains (String.lowercase_ascii info.description) pattern       
            || List.exists (fun tag -> contains (String.lowercase_ascii tag) pattern) info.tags       
         then
           true
         else
           false)
```
[...]

PR #131, opened 2021-10-04
smarter and sorted package search

panglesd:
This PR broaden packages search to tags, description and synopsis as proposed in #87.

Additionally, the result of a search is now sorted as follow:
1) Packages whose name is exactly the given string
2) packages whose name contain the given string
3) packages having the given string as a tag
4) packages whose synopsis contain the given string
5) packages whose description contain the given string.

However, the implementation require to call `Lazy.force` on *all* package info, which might introduce performance issues. In the few tests I made, a search request (after a server restart) never took more than 1s to be answered.

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Unreleased
 
+- Smarter and sorted package search (#131, by @panglesd)
+
+  Improve the search algorithm to sort packages by relevance.
+
+  The algorithm now uses the synopsis, descriptions and tags of the package.
+
 - Better formatting of odoc def tables, hiding comment-delim (#130, by @panglesd)
 
   Added css rules to hide comment delimiters and improve overall sizing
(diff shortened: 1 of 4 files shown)


tmattio:
Thanks @panglesd - this looks great! I've refactored the function to extract the general `contains` function to `Std.String` and created functions for the computations that were repeated twice (e.g. checking if the name of the package contains the pattern).

A search query returns in ~1.5s the first time for me. It seems acceptable, so let's merge and deploy as-is, and monitor how the performance looks like in production.

patricoferris:
Thanks @panglesd :)) I'm wondering if we need the the package information to be lazy anymore, iiuc we're only getting slightly faster start-up times in exchange for a bit of a bigger memory footprint and a slight indirection?

tmattio:
This would solve lots of limitations we currently have indeed (cough.. rev-deps.. cough..).

The startup time if we load everything in memory is unbearable though (~10 seconds). We can't have this kind of feedback loop when working on the frontend, where the feedback loop to validate changes visually needs to be very quick.

I'm not against having everything in memory if it makes sense from a resource perspective (i.e. memory), but I'd like us to keep a quick feedback loop during development if possible.

PR #132, opened 2021-10-05
Remove vertical scrollbar for the documentation navigation bar

TheLortex:
It rendered like this because of an overflow caused by a negative margin: 
Screenshot 2021-10-05 at 10-37-16 Cmdliner · cmdliner 1 0 4 · OCaml Packages

Now it's like this:
Screenshot 2021-10-05 at 10-37-49 Cmdliner · cmdliner 1 0 4 · OCaml Packages

Not sure why there was a negative margin though.


tmattio:
Thanks @TheLortex 🎉

PR #134, opened 2021-10-05
Rev deps experiment, continued

TheLortex:
This is based on https://github.com/ocaml/v3.ocaml.org-server/pull/109

The improvements:

- Don't block the dream loading, so the website is up immediately
- Package state (the `Info.t OpamPackage.Version.Map.t OpamPackage.Name.Map.t`) is marshalled in the cache folder for fast restarting. It stores the opam repository commit and a version number for invalidation purposes.
- `read_packages` and `Info.of_opamfiles` now return `Lwt.t` types, allowing to yield back to the scheduler in long-running operations. I've identified two heavy functions that now return lwt types: `rev_depends` and `mk_revdeps`. Therefore the 30 seconds of revdeps calculation now run in the background and the site can function as normally. There are still 10 seconds of blocking calculation somewhere that I've failed to identify.


src/ocamlorg/lib/package.mli
```diff
@@ -71,6 +76,8 @@ module Module_map = Module_map
 
 type state
 
+val get_state : state -> Info.t OpamPackage.Version.Map.t OpamPackage.Name.Map.t
```
tmattio: Do we need to expose this in the interface?

src/ocamlorg/lib/package.ml
```diff
@@ -122,60 +286,105 @@ let state_of_package_list (pkgs : t list) =
...
-        let versions = read_versions package_name in
+        let+ versions = read_versions package_name in
         OpamPackage.Name.Map.add (Name.of_string package_name) versions acc)
     OpamPackage.Name.Map.empty
     (Opam_repository.list_packages ())
 
-let update t =
-  let packages = read_packages () in
+let try_load_state () =
+  let exception Invalid_version in
```
tmattio, line 306: I didn't know you could declare exceptions in local scopes, very nice!

src/ocamlorg/lib/package.ml
```diff
@@ -58,58 +103,177 @@ module Info = struct
...
 
 type state =
-  { mutable packages :
-      Info.t lazy_t OpamPackage.Version.Map.t OpamPackage.Name.Map.t
+  { version : string
+  ; mutable opam_repository_commit : string option
+  ; mutable packages : Info.t OpamPackage.Version.Map.t OpamPackage.Name.Map.t
   }
 
+let get_state { packages; _ } = packages
```
tmattio: The name of the function is a bit confusing. We have a type `state`, and the function `get_state` returns the packages of this state.

src/ocamlorg/lib/package.ml
```diff
@@ -46,6 +51,46 @@ module Info = struct
...
+              in
+              OpamPackage.Map.add (OpamPackage.create name version) deps acc)
+            vmap
+            OpamPackage.Map.empty
+        in
+        OpamPackage.Map.union (fun a _ -> a) acc v)
+      opams
+      OpamPackage.Map.empty
+
+  let _get_conflict_set pkgs map =
```
tmattio: Why prepend the functions with `_`? If we dont use them, should we delete them?

src/ocamlorg/lib/opam_repository.ml
```diff
@@ -86,10 +85,10 @@ let list_package_versions package =
   ls_dir Fpath.(to_string (clone_path / "packages" / package))
 
 let process_opam_file f =
-  let ic = open_in (Fpath.to_string f) in
-  let result = OpamFile.OPAM.read_from_channel ic in
-  close_in ic;
-  result
+  let open Lwt.Syntax in
```
tmattio, line 88: Any reason to switch to Lwt? I'm not sure we're gaining anything and this will require all the functions that want to use it to wrap their return type in Lwt.t

src/ocamlorg/lib/package.ml
```diff
@@ -58,58 +103,177 @@ module Info = struct
...
+          Lwt_fold.package_version_map
+            (fun version opam acc ->
+              let package = OpamPackage.create name version in
+              let opam_file =
+                OpamPackage.Name.Map.find name opams
+                |> OpamPackage.Version.Map.find version
+              in
+              let+ rev_deps = mk_revdeps package packages rev_deps in
+              let t =
+                { synopsis =
```
tmattio: We could create a `make` function for this

tmattio (commented):
Looks great, thanks @TheLortex!

One thing I'm wondering is whether we'd like to keep the lazy loading if the state is not loaded yet. If I understand correctly, the state will be empty if the file isn't found in the present solution - instead, we could revert to loading the individual package when that's the case.

src/ocamlorg/lib/package.ml
patricoferris: This is an artefact from the initial work, opam2web supports things like `rev_depopts` which uses all of the same logic as `rev_deps` excepts for the sets of packages brought in see: https://github.com/ocaml-opam/opam2web/blob/master/src/o2wPackage.ml#L449 (we can drop these I don't mind... but might be nice to eventually support it)

src/ocamlorg/lib/opam_repository.ml
patricoferris: I think this is required because these changes move the package dependency reading and calculations into a "yieldable" fold over the opam files (the `Lwt.pause` in `Lwt_fold`). We don't want to do blocking I/O here... that is if I understood it correctly :))

src/ocamlorg/lib/package.ml
```diff
@@ -58,58 +103,177 @@ module Info = struct
...
+  let mk_revdeps pkg pkgs rdepends =
+    let open OpamTypes in
+    Lwt_fold.package_name_map
+      (fun name versions acc ->
+        let vf =
+          OpamFormula.formula_of_version_set
+            (OpamPackage.versions_of_name pkgs name)
+            versions
+        in
+        let _latest =
```
patricoferris: We should probably expose this `latest` to the code, this is useful for linking correctly i.e. a package that has a dependency like `A = version` currently just links to the latest version of the package which is wrong. We could do it also in a separate PR, so feels free to remove this :))

patricoferris (commented):
Looking good to me, when I have some more free time later this afternoon, I'll check it out locally :))

src/ocamlorg/lib/opam_repository.ml
TheLortex: Indeed the goal is to avoid blocking IO, because any blocking computation would make the server unresponsive for a potentially long time (like the 30 seconds of the `update` call). 
Switching to `Lwt.t` for such computations allow to interleave the 30 seconds workload with other operations from the dream server, such as answering a request.

TheLortex:
Thanks for the review!
About the lazy loading: indeed when the server starts from scratch, the state is empty and becomes populated 30 seconds after when all the informations are computed. What we could do is add an intermediate stage where packages can be individually loaded with partial information (revdeps are not known yet). I'm so sure of the benefit though:
- in production, the server wants to atomically update from one full state to another.
- in development, the situation would only occur once, when starting from scratch. Then the state would be cached.

tmattio:
> I'm so sure of the benefit though

Right, then maybe let's go without it for now and see if we hit any problem because of the 30sec initial loading.

src/ocamlorg/lib/package.mli
TheLortex: It doesn't look likes its useful for now so I've removed it

src/ocamlorg/lib/package.ml
TheLortex: Removed it as it's unused for now.

src/ocamlorg/lib/package.ml
TheLortex: As we don't use the code for now I'm in favor of removing it -- we can add support for it later

src/ocamlorg/lib/package.ml
TheLortex: For now I prefer to focus on what we have which is already quite a chunk (marshal state / switch to lwt for heavy computations / rev deps computation and display).
Let's do that in a separate PR !

TheLortex:
I've removed the unused code, extracted a `make` function for `Info.t` and added a changelog entry. It's all as separate additional commits, when the review is done I'll rebase and squash things together :)

src/ocamlorg/lib/config.ml
```diff
@@ -23,3 +23,8 @@ let topelevels_path =
   Sys.getenv_opt "OCAMLORG_TOPLEVELS_PATH"
   |> Option.map (fun x -> Result.get_ok (Fpath.of_string x))
   |> Option.value ~default:Fpath.(v "src" / "ocamlorg_toplevel" / "bin" / "js")
+
+let package_state_path =
+  Sys.getenv_opt "OCAMLORG_PKG_STATE_PATH"
```
tmattio, line 28: We should set this variable in the Dockerfile.

TheLortex: Indeed, thanks

tmattio (approved):
Thanks again for the awesome work @TheLortex! Before we merge, can you confirm that you can still build and run the docker image locally?

TheLortex:
Yes, I've just tested locally and the Docker build / run works fine

Issue #127, opened 2021-09-29
A timeout in toplevel is needed for safer browsing


hyphenrf:
*Alternatively, we should use a custom toplevel with a limited subset of the language (most Sys things seem unneeded)*

With the introduction of #106 we're able to crash the browser page by writing an infinite loop:
```ocaml
while true do () done;;
```
at toplevel. Similarly, large outputs slow the page down.
Timeout and Truncation are often employed in most online language playgrounds. We should aim at that kind of safety as well.

On a similar note, although seems benign, we're able to get directory listings and change directories via both `Sys` and toplevel directives.

tmattio:
Thanks @hyphenrf!

> Alternatively, we should use a custom toplevel with a limited subset of the language (most Sys things seem unneeded)

Very good idea, could you open a separate issue for this?

> On a similar note, although seems benign, we're able to get directory listings and change directories via both Sys and toplevel directives.

Is it a problem you think? The toplevel is running on the client, so it doesn't seem like a security issue to me.

hyphenrf:
> could you open a separate issue for this?

sure!

> Is it a problem you think? The toplevel is running on the client

seems not, I'm not a security guru haha

patricoferris:
Thanks @hyphenrf see #135 where I've split the jsoo toplevel into a web-worker which (a) means the page should always remain interactive even in the face of `while true do () done;;` and (b) we can easily terminate the worker if we don't get a timely response.

I don't think there are any security issues here, as @tmattio said it's all on the client and in fact it is a fake filesystem built into how the toplevel works :))

patricoferris:
The toplevel now has a timeout from #135 and a separate issue exists for the other suggestions #129, thanks again for the good suggestions :))

PR #135, opened 2021-10-05
split jsoo toplevel into web-worker

### LINKED ISSUE #133: Lazily load the homage toplevel
The toplevel on the homepage is ~3MB. We should not load it at every request, instead, we could provide a button for users to load it manually.

cc @TheLortex

patricoferris:
This PR aims to partially fix #133 and #127 by moving the jsoo execution to a web worker which makes the chunk of the code easier to lazy load, execution of OCaml code doesn't block the main UI thread, easy to terminate the worker if there is no response in a certain amount of time (here `10s`). It also gives the toplevel a bit of a different UI just to work in the idea of having to load the toplevel runtime code in. The initial JS loaded into the homepage is now (after `dune build --profile release`) `~80kB` gzipped (`~213kB` otherwise) which is better (we could probably go further, but perhaps in a separate PR)

**Toplevel after loading the home page**

Screenshot 2021-10-05 at 18 22 21

**Toplevel after clicking load and entering a few expressions**

Screenshot 2021-10-05 at 18 23 24

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,11 @@
 # Unreleased
 
+- Split js_of_ocaml toplevel execution into web-worker (#135, by @patricoferris)
+
+  The execution of OCaml expressions in the toplevel now takes place in a web 
+  worker which prevents the main UI thread from blocking and makes it easier to
+  terminate executions that have been running too long.
+
 - Compute and display reverse dependencies (#134, by @patricoferris and @TheLortex)
 
   Extend package information structure to add reverse dependencies. Display that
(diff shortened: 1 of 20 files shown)


tmattio (approved):
This looks so good! 🤩 

The user experience looks great and we'll really see the difference in terms of performance.

I do have a slight preference for a dark-themed top level, but we can revisit it when we come back on the design of the homepage.

Feel free to merge when you feel this is ready!

patricoferris:
macOS errors are probably from https://github.com/ocaml/opam-repository/issues/19676

PR #147, opened 2021-10-07
Add 99 problems

tmattio:
This imports the 99 problems data.

The problems are now separated into several files, where each file follows this structure:

```yaml
---
title: ...
number: "..."
difficulty: ...
tags: [ ... ]
---

# Solution

<!-- The solution body here -->

# Statement

<!-- The statement body here -->
```

The Solution and Statement heading must be the only level-one heading in the document.

image


src/ocamlorg_data/ood.mli
```diff
@@ -177,6 +177,29 @@ module Paper : sig
...
+module Problem : sig
+  type difficulty =
+    [ `Beginner
+    | `Intermediate
+    | `Advanced
+    ]
+
+  type t =
+    { title : string
+    ; number : string
```
patricoferris, line 189: Any particular reason these are strings?

patricoferris (approved):
Thank @tmattio these look great!

src/ocamlorg_data/ood.mli
tmattio: Yep - some problems have letters in their IDs, e.g. `42B`

Issue #104, opened 2021-09-21
Complete redirections from V2


tmattio:
#103 introduced a redirection module that maps the URLs from V2 to their new location in V3. While some of the URLs are already handled (the tutorials), there are still some URLs that need to be mapped.

rgrupesh:
@tmattio I would like to work on this issue. What specific redirect should I specifically looking for? Any pointers to get started?

tmattio:
Hi @rgrupesh 🙂 That sounds like a great first issue. The list of redirections to add can be seen at https://github.com/ocaml/v3.ocaml.org-server/blob/main/src/ocamlorg_web/lib/redirection.ml. The `???` are the ones that are yet to be handled. To find the equivalent route, I'd recommend looking at the router at https://github.com/ocaml/v3.ocaml.org-server/blob/main/src/ocamlorg_web/lib/router.ml, in particular the page routes at https://github.com/ocaml/v3.ocaml.org-server/blob/main/src/ocamlorg_web/lib/router.ml#L27.

Let me know if you have any questions!

rgrupesh:
Thank you for these links @tmattio . I found some of the route and added those to the `redirection.ml`. I am having trouble finding all the routes. Are all required `.md` files present in data directory? Should I be expecting to find all the routes? 

And what is the difference between `release` and `releases`? Are they two different urls?

What should be an ideal workflow? One workflow example would be great.

tmattio:
@rgrupesh Some of the routes won't be available in the new website, that's ok and you can redirect these to the homepage (i.e. `/`).

Don't hesitate to open a draft PR and I can help you to identify which page should have redirections and which one shouldn't.

> And what is the difference between release and releases? Are they two different urls?

`releases` contains the list of releases (see https://v3.ocaml.org/releases) and `release` is the content of a single release (see https://v3.ocaml.org/releases/3.12.1)

> What should be an ideal workflow? One workflow example would be great.

I'm not sure I understand the question 🙂 If you mean how to find the right redirections, I'd start with the old website URL in `redirection.ml`, see if there is an equivalent in `url.ml`, if so use this route, if not, use `/`.

You can also navigate both sites to get a good idea of what content/page is available on both.

PR #152, opened 2021-10-11
couple of redirection added

rgrupesh:
related to #104 

Couple of fixes


tmattio:
Thanks @rgrupesh!

Looks like a good start, I think the releases redirections should be relatively easy to add as well. I'll check for the others to give you some pointers 🙂 

Could you also revert the removal of the `package-lock.json` file?

rgrupesh:
@tmattio Do I need to write `release` function to route those releases Urls? Can it be routed using relative path? How should I go about this?

tmattio:
@rgrupesh To give an example, this redirection: https://github.com/ocaml/v3.ocaml.org-server/blob/main/src/ocamlorg_web/lib/redirection.ml#L2 (which is this page on the current ocaml.org: https://ocaml.org/releases/4.05.html) can redirect to `/releases/4.05.0` (this page on v3: https://v3.ocaml.org/releases/4.05.0).
And the same mapping can be done for every release.

rgrupesh:
@tmattio I am still working on this. Just to let you know, we have festival season going on that is why I was inactive lately. I will be committing fully to this PR after couple of days.

Besides, when I try to add releases URL like this:



I get this error message:



To achieve routing like above, do I need to write function like `tutorial` in `page_handler.ml`. Or, there is another way around?

P.S: I am newbie to this ecosystem. I am learning while contributing. I hope you understand.

gs0510:
@rgrupesh If you look at the error message, it tells you that the expression `Url.releases` has a type string, which is why you can't pass a string parameter to it. You need to write a function so that it can take a string parameter which can redirect the users to the correct link. :)

rgrupesh:
@gs0510 I didn't find any `.md` files related to the `/community`. Are those routes available? Or, should I redirect all of those paths to the `/`?

tmattio:
Thanks a lot for all the work @rgrupesh! I've added a few missing redirections in addition to the one you added, this is now good to be merged 🙂

rgrupesh:
> Thanks a lot for all the work @rgrupesh! I've added a few missing redirections in addition to the one you added, this is now good to be merged slightly_smiling_face

Thanks to you @tmattio . I learned a lot about the project structure and code base under your supervision.

PR #155, opened 2021-10-12
Add very simple tab-completion to the toplevel

jonludlam:
This is based on uTop's implementation, with all of the utf-8 parts
stripped out. Unfortunately compiling camomile with js_of_ocaml was
causing an error early during initialisation.

I've kept the uTop modules in the same structure they were originally
so it's easy to reinstate if camomile gets fixed or if we find a
suitable replacement.

The UI/UX/DX could do with some love!

### DIFF
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Add completion to toplevel (#155, by @jonludlam)
+
 - Add French content for academic users (#150, by @maiste)
 
 - Add 99 problems (#147, by @tmattio)
--- a/src/ocamlorg_toplevel/bin/.ocamlformat-ignore
+++ b/src/ocamlorg_toplevel/bin/.ocamlformat-ignore
@@ -1,0 +1,2 @@
+uTop.ml
+uTop_complete.ml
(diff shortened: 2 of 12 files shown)


patricoferris (approved):
LGTM (I didn't go through the vendorered utop code haha)! I'm okay to merge even without the UI work (tab completing `List.` causes a list that doesn't wrap for example). 

Looks like the linter is stuck on the cppo file, are these usually named something different than `ml` to prevent this? If not we could add them to the ocamlformat ignore file to fix CI :))

tmattio:
Looks great, thanks @jonludlam!!

Re UI: can we list the candidates for completion with a line return? Currently, they are truncated and we can't see them if there are more than 4/5. If it's too complicated, I don't mind doing this in a follow up PR either.

Same suggestion as Patrick for the CI: we can add the vendored files in `ocamlformat-ignore` to fix the lint.

jonludlam:
Yep, we've got our cppo files in odoc in `.ocamlformat-ignore` - it'd be nice to fix this!

I initially did the completions line-by-line, but that seemed to take a _lot_ of screen space. I think we can come up with something better, but I reckon that can be done in a follow up.

tmattio:
Thanks a lot @jonludlam 🎉

PR #156, opened 2021-10-13
Fix leaking file descriptor

tmattio:


Dockerfile
```diff
@@ -30,6 +30,7 @@ ENV OCAMLORG_REPO_PATH /var/opam-repository/
 ENV OCAMLORG_PKG_STATE_PATH /var/package.state
 ENV OCAMLORG_TOPLEVELS_PATH /var/toplevels/
 ENV OCAMLORG_DEBUG false
+ENV DREAM_VERBOSITY info
```
TheLortex, line 33: Is that compatible with GDPR as long as we remove the logs once in a while ?

TheLortex (approved):
Thanks, this should fix the main issue.

For the logging, I believe we will obtain source IPs so we might need to be careful about that.

Dockerfile
tmattio: I don't see why it wouldn't be compliant: there's no personal data associated with the info logs, only the requested URLs.

I'm not a GDPR expert though 🤷

TheLortex: Oh my bad, I thought the IP would be shown if logging requests was enabled. So it's ok we don't have any personal data.

PR #159, opened 2021-10-13
Toplevel rpc

jonludlam:
This is on top of the completion PR. Personally I think this looks tidier, but I appreciate there's some extra cognitive load due to the RPC definitions. The RPC library is inspired by `ctypes` so anyone who's done C bindings with that might have an easier time with the IDL file.

### DIFF
--- a/ocamlorg.opam
+++ b/ocamlorg.opam
@@ -38,6 +38,9 @@ depends: [
   "yojson"
   "lwt"
   "brr"
+  "rpclib"
+  "rpclib-lwt"
+  "ppx_deriving_rpc"
   "js_of_ocaml" {>= "3.11.0"}
   "js_of_ocaml-ppx" {>= "3.11.0"}
   "js_of_ocaml-compiler" {>= "3.11.0"}
--- a/src/ocamlorg_toplevel/bin/worker.ml
+++ b/src/ocamlorg_toplevel/bin/worker.ml
@@ -1,15 +1,12 @@
 open Js_of_ocaml_toplevel
-open Brr
-open Brr_io
-open Ocamlorg_toplevel.Toplevel
+open Ocamlorg_toplevel
+open Toplevel
 
 (* OCamlorg toplevel in a web worker
 
-   This communicates with the toplevel code via simple json schema, this allows
-   the OCaml execution to not block the "main thread" keeping the page
-   responsive. *)
-
-let jstr_of_buffer v = Jstr.v @@ Buffer.contents v
+   This communicates with the toplevel code via the API defined in
+   {!Toplevel_api}. This allows the OCaml execution to not block the "main
+   thread" keeping the page responsive. *)
 
 module Version = struct
   type t = int list
@@ -55,6 +52,7 @@ module Version = struct
       (match compint x y with 0 -> compare xs ys | n -> n)
 end
 
+
 let exec' s =
   let res : bool = JsooTop.use Format.std_formatter s in
   if not res then Format.eprintf "error while evaluating %s@." s
@@ -91,6 +89,20 @@ let stdout_buff = Buffer.create 100
 
 let stderr_buff = Buffer.create 100
 
+
+(* RPC function implementations *)
+
+module M = Idl.IdM (* Identity monad - server is synchronous *)
+module IdlM = Idl.Make (M)
+module Server = Toplevel_api.Make (IdlM.GenServer ())
+
+(* These are all required to return the appropriate value for the API
+   within the [IdlM.T] monad. The simplest way to do this is to use
+   [IdlM.ErrM.return] for the success case and [IdlM.ErrM.return_err]
+   for the failure case *)
+
+let buff_opt b = match Buffer.contents b with "" -> None | s -> Some s
+
 let execute =
   let code_buff = Buffer.create 100 in
   let res_buff = Buffer.create 100 in
@@ -100,7 +112,7 @@ let execute =
   let highlight_location loc =
     let _file1, line1, col1 = Location.get_pos_info loc.Location.loc_start in
     let _file2, line2, col2 = Location.get_pos_info loc.Location.loc_end in
-    highlighted := Some ((line1, col1), (line2, col2))
+    highlighted := Some Toplevel_api.{ line1; col1; line2; col2 }
   in
   fun phrase ->
     Buffer.clear code_buff;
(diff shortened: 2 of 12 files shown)


src/ocamlorg_toplevel/lib/dune
```diff
@@ -1,6 +1,13 @@
 (library
  (name ocamlorg_toplevel)
  (public_name ocamlorg.toplevel)
- (libraries lwt brr js_of_ocaml-tyxml)
+ (libraries
+  lwt
+  brr
+  js_of_ocaml-tyxml
+  rpclib
+  ppx_deriving_rpc
```
patricoferris: We don't need the ppx as a library dependency. IIUC this confuses jsoo and bloats the final bundle by quite a lot. When built under the `release` profile I get (all uncompressed) for `stdlib-4.13.0.js`:

 - main branch: `220KB`
 - This PR: `2.3MB`
 - This PR with `ppx_deriving_rpc` removed from `libraries`: `243KB`
 
(I think I did this check correctly, would be good if someone could double-check)

patricoferris (changes requested):
Looks really great! Definitely once you get your head around `ocaml-rpc` it tidies up a substantial amount of the code and as @tmattio pointed out offline, should make it quite portable if we ever need to change the underlying transport mechanism. One small change (with a seemingly big impact) and this should be good to go I think. Thanks :))

src/ocamlorg_toplevel/lib/dune
jonludlam: Aha good spot, thanks!

patricoferris (approved):

src/ocamlorg_toplevel/bin/worker.ml
```diff
@@ -91,6 +88,20 @@ let stdout_buff = Buffer.create 100
 
 let stderr_buff = Buffer.create 100
 
+(* RPC function implementations *)
+
+module M = Idl.IdM (* Identity monad - server is synchronous *)
+
+module IdlM = Idl.Make (M)
```
tmattio, line 95: Can we call it `Identity` or `Rpc_identity`? It'd make the code easier to read I think

ocamlorg.opam
```diff
@@ -38,6 +38,9 @@ depends: [
   "yojson"
   "lwt"
   "brr"
+  "rpclib"
+  "rpclib-lwt"
+  "ppx_deriving_rpc"
```
tmattio, line 43: We need to add these to the `dune-project` deps

src/ocamlorg_toplevel/lib/toplevel.ml
```diff
@@ -20,10 +20,21 @@
...
-open Brr_io
+
+(* open Brr_io *)
 open Lwt
 module Worker = Brr_webworkers.Worker
 
+module Toprpc = Toplevel_api.Make (Rpc_lwt.GenClient ())
+
+(* Handy infix bind-style operator for the RPCs *)
+let ( >>>= ) x f =
```
tmattio: We can define a let binding here to stay consistent with the rest of the code

src/ocamlorg_toplevel/bin/worker.ml
```diff
@@ -110,59 +121,50 @@ let execute =
...
+      ; stderr = buff_opt stderr_buff
+      ; sharp_ppf = None
+      ; caml_ppf = None
+      ; highlight = None
+      }
+
+let complete phrase =
+  let n, res = UTop_complete.complete ~phrase_terminator:";;" ~input:phrase in
+  let res =
+    List.filter (fun (l, _) -> not (Astring.String.is_infix ~affix:"__" l)) res
```
tmattio: We could define `is_infix` with `String.sub` here to avoid adding the `Astring` library (even if it's already a transitive dep)

src/ocamlorg_toplevel/bin/worker.ml
```diff
@@ -110,59 +121,50 @@ let execute =
...
+  in
+  let completions = List.map fst res in
+  IdlM.ErrM.return Toplevel_api.{ n; completions }
+
+let server process e =
+  let ( >>= ) = M.bind in
```
tmattio: Same as above, we can use a let binding

src/ocamlorg_toplevel/lib/toplevel_api.ml
```diff
@@ -0,0 +1,94 @@
+(** IDL for talking to the toplevel webworker *)
```
tmattio, line 1: Can we put the docstrings in an interface file?

src/ocamlorg_toplevel/rpc_brr/worker_rpc.ml
```diff
@@ -0,0 +1,49 @@
+(** Worker rpc *)
```
tmattio, line 1: Same as above, can we create an interface for the modules' documentation?

tmattio (approved):
Looks great, thanks a lot! It's great to see clear documentation for the protocol and the introduced modules!

Leaving some minor comments, but feel free to merge whenever you're ready.

src/ocamlorg_toplevel/rpc_brr/worker_rpc.ml
jonludlam: Actually it's better this way for the rpc things - `ocaml-rpc` actually uses docstrings. See an example [here](https://github.com/xapi-project/xapi-storage/blob/master/generator/lib/data.ml#L12-L22) which gets turned into the document [here](https://xapi-project.github.io/xapi-storage/#datapath-type-definitions)

jonludlam: Oops, you're right for `worker_rpc.ml` - it's the IDL that should keep the comments in the `.ml` file

src/ocamlorg_toplevel/bin/worker.ml
jonludlam: I'll use the function we have in odoc.

ocamlorg.opam
jonludlam: Ah yes!

src/ocamlorg_toplevel/bin/worker.ml
jonludlam: Hrm, perhaps I should just get rid of the comment - it's not strictly the identity monad ;-) (it boxes the results because of lack of injectivity annotation in the past). The important part of the comment is that we're using the synchronous module rather than the Lwt one.

jonludlam:
Thanks for the review @tmattio - I've addressed most of the comments. In adding the [.mli] for [Worker_rpc] I realised our handling of timeouts wasn't particularly good so I've improved that too.

jonludlam:
Let's get this in.

PR #170, opened 2021-10-23
refactor no_trailing_slash middleware 

### LINKED ISSUE #166: Handle trailing slash
Currently most of the pages 404 when there's a trailing slash, I think commonly web pages either still provide the implicit `index.html` file or do a redirection either to or from the trailing slash version. For example https://v3.ocaml.org/users is fine but https://v3.ocaml.org/users/ 404s, this was most noticeable for me when looking at packages like https://v3.ocaml.org/p/conan which redirects but https://v3.ocaml.org/p/conan/ does not. 

I don't know what the cleanest way to handle this, some middleware @tmattio ? The old site seems to redirect anything that doesn't end with a trailing slash to the same URL with one e.g. https://ocaml.org/docs goes to https://ocaml.org/docs/

desirekaleba:
Closes #166

### DIFF
--- a/src/ocamlorg_web/lib/ocamlorg_web.ml
+++ b/src/ocamlorg_web/lib/ocamlorg_web.ml
@@ -15,5 +15,6 @@ let run () =
   let state = Ocamlorg.Package.init () in
   Dream_cli.run ~debug:Config.debug ~interface:"0.0.0.0" ~port:Config.port
   @@ Dream.logger
+  @@ Middleware.no_trailing_slash
   @@ Router.router state
   @@ Page_handler.not_found
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -1,5 +1,7 @@
 # Unreleased
 
+- Fix handling of trailing slashes (#170, by @desirekaleba)
+
 - Demarcate include blocks (#54, by @JiaeK)
 
 - Add semicolons `;;` to 99 Problems toplevel sentences. (#176, by @Lontchi12)
--- a/src/ocamlorg_web/lib/middleware.ml
+++ b/src/ocamlorg_web/lib/middleware.ml
@@ -1,11 +1,13 @@
 let no_trailing_slash next_handler request =
   let target = Dream.target request in
-  if target = "/" then
+  match target with
+  | "/" ->
     next_handler request
-  else if String.get target (String.length target - 1) = '/' then
-    Dream.redirect request (String.sub target 0 (String.length target - 1))
-  else
-    next_handler request
+  | _ ->
+    if String.ends_with ~suffix:"/" target then
+      Dream.redirect request (String.sub target 0 (String.length target - 1))
+    else
+      next_handler request
 
 let set_locale handler req =
   let language =
--- a/src/ocamlorg_web/lib/router.ml
+++ b/src/ocamlorg_web/lib/router.ml
@@ -27,7 +27,7 @@ let redirection_routes =
 let page_routes =
   Dream.scope
     ""
-    [ Middleware.set_locale; Middleware.no_trailing_slash ]
+    [ Middleware.set_locale ]
     [ Dream.get Url.index Page_handler.index
     ; Dream.get Url.history Page_handler.history
     ; Dream.get Url.around_web Page_handler.around_web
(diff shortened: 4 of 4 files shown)


tmattio:
Hi @desirekaleba, thanks for the PR! While the code is correct and looks good, it doesn't seem to be the expected behavior. The code will make it so pages with a trailing slash will be served with the same response as without the trailing slash. What we would like, however, is to redirect (HTTP response code 3XX) users to the pages without the trailing slash when they request a page with one. So the workflow is:

- User requests `v3.ocaml.org/page/`
- Middleware sees there is a trailing slash and answers with code 3XX and location `/page` (that's the `Dream.redirect` function that was in the middleware previously)
- User requests `v3.ocaml.org/page`
- Server serves the handler matching the `/page` route.

desirekaleba:
@tmattio, I tried most of the approaches but I could not understand why the middleware is not responding when the user requests a page with a trailing slash. It only works on the index page. That's why I drafted the PR as I am figuring out what's really going on for it not to work.

desirekaleba:
@tmattio, after refactoring the middleware, I noticed that the middleware is only being called when the user hits a valid URL. For example, If the route goes to `/users/`, then the middleware will not be called as this will call our 404 handler. 

I came across [this discussion](https://github.com/aantron/dream/issues/24) on its [repository](https://github.com/aantron/dream/) which made me think that `Dream` does not have a way of handling this yet. so I used the `redirection.ml` file to add trailing slashes to the possible URLs and after, redirect them to the correct ones.

tmattio:
@desirekaleba Good catch! I guess that's because the trailing slash middleware is applied after the router middleware. To fix that, you can try to apply the middleware before the router in `ocamlorg_web.ml`. The drawback is that this'll remove the trailing slash in **every** URL before matching them, which may be incompatible with some routes, but that seems fine for now and we can change that if we hit a problem later.

desirekaleba:
Oh, I get that now.

desirekaleba:
@tmattio, This is working as expected now.

desirekaleba:
@tmattio, I need your review, please.

src/ocamlorg_web/lib/redirection.ml
```diff
@@ -211,6 +211,51 @@ let t =
   ; "/index.html", Url.index
   ; "/index.fr.html", Url.index
   ; "/contributors.html", Url.index
+  ; "/ocamllabs/index.html", "/???"
```
tmattio: I don't think these are necessary

src/ocamlorg_web/lib/ocamlorg_web.ml
```diff
@@ -15,5 +15,6 @@ let run () =
   let state = Ocamlorg.Package.init () in
   Dream_cli.run ~debug:Config.debug ~interface:"0.0.0.0" ~port:Config.port
   @@ Dream.logger
+  @@ Middleware.no_trailing_slash
```
tmattio, line 18: You'll need to remove it from the Router module if you put it here

tmattio (changes requested):
Looks good, thanks a lot! Leaving a few comments, this is good to be merged once they are addressed!

src/ocamlorg_web/lib/redirection.ml
desirekaleba: These were fetched from the main branch. I am currently trying to reference the PR which added them.

tmattio: I'm not sure what happened, but if they show in the diff, it's because these are changes when compared from the main branch, so you can delete them 🙂

desirekaleba: I am raising a new PR without them.

desirekaleba: @tmattio, Can you please rerun the checks on this PR? I don't understand why they are failing.

desirekaleba:
@tmattio, I have updated this PR. I need your review on this, please.

tmattio:
Thanks a lot @desirekaleba, I've reverted the changes on package-lock.json, this looks good to be merged 🙂

desirekaleba:
Closes #166

TheLortex:
@tmattio I tried re-using the content of this PR in mirage.io, but @reynir found issues with it (https://github.com/mirage/mirage-www/pull/788#issuecomment-1373455519). 

> This change allows you to craft URLs at mirage.io that redirect to arbitrary URLs, e.g https://mirage.io///github.com/mirage/mirage-www/pull/788/. I'm not sure what's the best way to fix that. It can also lead to too many redirect: https://mirage.io/blog///////////////////////////////////////////////////////////////

We can craft similarly problematic URLs with ocaml.org:
-  https://ocaml.org/packages///////////////////////////////////////////////////////////////
- https://ocaml.org//github.com/ocaml/ocaml.org/pull/170/

tmattio:
Good catch, thanks for the heads up @TheLortex! Would you mind opening an issue for this?

Issue #54, opened 2021-07-29
Include blocks are not visibly demarcated.


jonludlam:
Odoc usually shows a border on the right to show which sections come from included signatures. See the right-hand side of the include here:

<img width="789" alt="Screenshot 2021-07-29 at 11 28 34" src="https://user-images.githubusercontent.com/210963/127476615-43dbbdf9-22e5-4893-8893-46ffb7809a33.png">

and the bottom of the included section here:
<img width="782" alt="Screenshot 2021-07-29 at 11 29 29" src="https://user-images.githubusercontent.com/210963/127476770-32572c15-b77c-43ff-a021-451bb550f7fd.png">

JiaeK:
Hi, can I work on this issue? And can I get some details?

jonludlam:
Sure. The issue is that the CSS used for v3.ocaml.org comes partly from tailwind's `typography` plugin with some [extra rules](https://github.com/ocaml/v3.ocaml.org-server/blob/main/asset/doc.css) - and so is missing some of the rules declared in [odoc's css](https://github.com/ocaml/odoc/blob/master/src/odoc/etc/odoc.css). We need to port the rules relating to `include` blocks into v3's CSS file.

PR #175, opened 2021-10-29
Demarcate include blocks (#54)

JiaeK:
Previously include blocks are not visually demarcated.

tyxml v3 page

And after this change, 

- Include blocks are demarcated 
- The border on the right-hand side shows which sections come from included signatures
include open

(also the bottom)
include open bottom

- When hovering, the shade of include blocks changes
include hover

### DIFF
--- a/asset/doc.css
+++ b/asset/doc.css
@@ -5,7 +5,7 @@ div.odoc {
   position: relative;
 }
 
-div.odoc div.spec {
+div.odoc div.spec, div.odoc-include .spec {
   border-left-width: 4px;
   border-color: rgba(251, 146, 60, 1);
   border-radius: 0.25rem /* 4px */;
@@ -86,12 +86,13 @@ div.odoc div.spec tbody td.def {
   min-width: 40%
 }
 
-.odoc-include details {
+div.odoc-include details {
   position: relative;
 }
 
-.odoc-include details:after {
+div.odoc-include details:after {
   display: block;
+  content: " ";
   position: absolute;
   border-radius: 0 1ex 1ex 0;
   right: -20px;
--- a/asset/doc.css
+++ b/asset/doc.css
@@ -5,7 +5,7 @@ div.odoc {
   position: relative;
 }
 
-div.odoc div.spec .spec.include {
+div.odoc div.spec {
   border-left-width: 4px;
   border-color: rgba(251, 146, 60, 1);
   border-radius: 0.25rem /* 4px */;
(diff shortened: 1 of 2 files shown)


asset/doc.css
```diff
@@ -85,3 +85,24 @@ div.odoc div.spec tbody td.def {
...
   min-width: 40%
 }
+
+.odoc-include details {
+  position: relative;
+}
+
+.odoc-include details:after {
+  display: block;
+  content: " ";
```
tmattio, line 95: Why do we need a `content`?

asset/doc.css
```diff
@@ -5,7 +5,7 @@ div.odoc {
   position: relative;
 }
 
-div.odoc div.spec {
+div.odoc div.spec .spec.include {
```
tmattio: AFAICT, `div.spec` should work on `.spec.include`, unless the elements that are not `spec include` are not `div`? Why was this change necessary?

tmattio (commented):
Thanks a lot @JiaeK, this looks great! I tested locally and it's working. Leaving a few comments for things I'm not sure about and we can merge after.

asset/doc.css
JiaeK: Yeah I wasn't sure about this - the previous Tyxml page code has this snippet and it's for the pseudo-element (i.e. `:after`) after all so I thought it would need this part. I'll erase it.

asset/doc.css
JiaeK: It's mainly due to the fact that this issue inquired demarcating the include blocks, so I assumed this part doesn't work for the include blocks particularly, so it needed another CSS selector to detect. I'm sorry, I'll also redo this.

asset/doc.css
tmattio: Maybe we do need it, it was just a genuine question 🙂  Let's try removing it and see how the docs looks!

asset/doc.css
tmattio: That's a welcome change, I was just surprised we needed this, because the `.spec.include` should already be a subset of `div.spec`, but if that's not the case, let's keep this change.

JiaeK: When I simulated it, it looked like it needed and I tried something like with `div.spec`, it didn't detect well.. Shall we also remove this and see how it goes?

JiaeK:
I did the update. Can you test locally again please? @tmattio

tmattio:
The changes were indeed necessary, so fixed the CSS after the last commit, merging now 🙂

PR #177, opened 2021-11-03
Fix hover effect for include blocks

JiaeK:
From the PR for Demarcate include blocks ([https://github.com/ocaml/v3.ocaml.org-server/pull/175](url)), the hovering effect did not work properly. Now it has fixed.


tmattio:
Thanks @JiaeK 🙂

PR #178, opened 2021-11-04
Fix robots.txt url

### LINKED ISSUE #158: SEO: robots.txt, sitemap.xml and noindex
The goal is to control and optimize how the website can be accessed from search engines. There are several methods to achieve that:

### robots.txt

This file describes how the site can be crawled by search engine bots: https://developers.google.com/search/docs/advanced/robots/intro

### sitemap.xml

This can be useful to expose the list of pages if we want the search engine to know about documentation pages: https://developers.google.com/search/docs/advanced/sitemaps/overview

### noindex

`<meta name="robots" content="noindex">` can be used to tell the crawler not to index the page

## For v3.ocaml.org

At first, we don't want the website to find discoverable through search engines, so `noindex` should be used for every page.

## In the future: ocaml.org

A sitemap should be provided, eventually with the documentation pages for the _latest_ packages. `noindex` should be used for parallel universes and older versions of packages. Finally, I guess it would make sense to have a somewhat stable URL for the last version of each package, something like https://v3.ocaml.org/p/base/latest/...

desirekaleba:
Closes #158


tmattio:
Thanks @desirekaleba 🙂 

The fix won't work in production, unfortunately, where we don't serve the assets from the filesystem, but from memory.

Instead, I would suggest that we remove the `assets/` prefix in the route, and move the CSS files in a subdirectory in `asset/` (we'll also need to fix the links to the CSS files of course)

desirekaleba:
Oh, I get it now. I was trying to understand why the jobs are failing. Let me try removing the `/assets` prefix and update the imports accordingly.

desirekaleba:
@tmattio, removing `/assets/**` in the router module, it has behaved as I was expecting(assets links are broken) but I don't seem to see where I should fix that from.

desirekaleba:
@tmattio, I need your review on this, please.

tmattio (approved):
This looks great, thanks @desirekaleba 🙂 
I'm testing locally and will merge if there's no issue.

Issue #81, opened 2021-09-01
HTTP requests are not upgraded to HTTPS


tmattio:
For obvious security reasons, we should redirect HTTP requests to HTTPS.

The HTTPS handling is done with a proxy/load balancer at the moment, so the redirection should probably be handled there, but we might also want to expose the server directly: Dream should handle HTTPS nicely.

patricoferris:
Didn't see this issue, I think it is likely a Caddy server configuration problem in V3 (http://ocaml.org redirects). 

Waiting on some feedback here and then will try to fix: https://github.com/ocurrent/ocurrent-deployer/pull/76#issuecomment-915309529

avsm:
It would be nice to have the Dream server directly exposed. All that's missing is automatic LetsEncrypt certificate fetching which Caddy does. What are the chances of getting that working with a bit of Dream middleware? :-)

tmattio:
> What are the chances of getting that working with a bit of Dream middleware? :-)

Doesn't seem too complicated, assuming that Dream supports upgrading connections from HTTP to HTTPS, which I think it does. cc @aantron

I'll look into it. Do you know where the certificates are downloaded from atm? How are they protected?

aantron:
Dream currently supports establishing an HTTPS connection.

To redirect from HTTP to HTTPS, you can currently issue an ordinary redirect to the client from a helper Dream server running in-process on port 80 (or other port, if using a proxy) and not using HTTP, to a main Dream server running in the same process on port 443 and using HTTPS. This should probably be wrapped up in some kind of neat helper.

Dream doesn't have nice Let's Encrypt support, but that is planned :)

tmattio:
Closing in favor of #231

PR #182, opened 2021-11-05
https redirect and letsencrypt certificate provisioning

patricoferris:
This PR:
 - Provisions a certificate with letsencrypt (LE)
 - Adds second small HTTP server that redirects to HTTPS whenever it is enabled and also responds to the LE challenge

By default nothing should change for people developing locally. This work was possible thanks to `unipi` showing the way: https://github.com/roburio/unipi :)) We store the certificates and key so we don't perform a new LE challenge every time the server rebuilds (after merging a PR for example).

Because Dream takes the certificate/key file at startup we envisage that we'll have to force a rebuild when the certificate is no longer valid (unless there's some other way to let Dream know it is no longer valid). This is a pretty minor problem.

TODO:

- Validate that the certs on filesystem are still valid
- Create directories to store certificates


tmattio:
@patricoferris I implemented the creation of directories (recursively) and the validity check of the certificates. I couldn't test because I don't have a domain available, but hopefully, the few refactorings I've made didn't break anything.

I think this is ready to deploy now, so cc @avsm if you would like to take over and try to deploy this branch. We should be able to get rid of the nginx proxy and expose the server directly.

Also cc @aantron, since I know you've been looking into adding support for Letsencrypt in Dream. You might find this PR useful.

tmattio:
Merging now so it doesn't go out of sync with `main` and we can spot any issue in development. We can fix any deployment issue when we deploy `main` next time.

hannesm:
Hmm, I'm not entirely sure about Dream and let's encrypt integration -- a sustainable way would be if there's no change / redeploy needed when the certificate expires. To achieve that, I'd approach this as follows:
- use the ALPN challenge from let's encrypt (to free the path of #81 and just redirect all HTTP requests to HTTPS)
- somehow allow in dream that an already established TLS socket (+ALPN string) is used
- a Lwt_engine.on_timer that whenever the certificate is about to expire (5 days earlier, ..) pulls a new one from let's encrypt and directly uses it (for new connections)
- the above could avoid any restarts / termination of active connections... very similar to how unipi and tlstunnel are doing it...

Ultmately, integrating let's encrypt into dream makes very good sense. Esp. now that dream contains a http client implementatio as well :)

Issue #164, opened 2021-10-18
Versioned links from dependencies and reverse dependencies 


patricoferris:
Currently from any given package page (e.g. https://v3.ocaml.org/p/ipaddr/5.0.0) we link to both the dependencies and the reverse dependencies from the overview page.

The concrete links we use are just to the package name which _always_ redirects to the _latest_ version even when there are version constraints involved. For instance https://v3.ocaml.org/p/ipaddr/5.0.0 has a dependency on `macaddr` with the version constraint `= version` clicking on this should go to https://v3.ocaml.org/p/macaddr/5.0.0 but instead we go to https://v3.ocaml.org/p/macaddr (at the time of writing this is `5.2.0`). 

Other version constraints might be `>= "3.1" & < "3.3`. In this case we should redirect to the _latest_ version which satisfies the version constraint.

IIRC for reverse dependencies we can add the logic in 

https://github.com/ocaml/v3.ocaml.org-server/blob/f9da1e68f6d57bc62eef3837f8f1d30ff6b89753/src/ocamlorg/lib/package.ml#L129

à la opam2web https://github.com/ocaml-opam/opam2web/blob/d1f4296c62da19713d358d1792c7138102678628/src/o2wPackage.ml#L384-L394

tmattio:
Would it make sense to support version constraints in the router? I.e. handling URLs like `https://v3.ocaml.org/p/ocaml-compiler-libs/>=v0.12.0&<v0.13.0`

JiaeK:
Hi, can I work on this issue?

tmattio:
Sure @JiaeK 🙂

patricoferris:
> Would it make sense to support version constraints in the router? I.e. handling URLs like https://v3.ocaml.org/p/ocaml-compiler-libs/>=v0.12.0&<v0.13.0

Hmm I think it would be easier to do this more statically when compiling the information about the package rather than dynamically through the router. As in the rev-deps example, we already have the complexity of calculating the reverse dependencies so it's just a short opam-library call away whereas moving things into the router would require exposing the server to opam libraries (even indirectly though the `Ocamlorg` module) right? Plus the versioning can be contextual i.e. where does `https://v3.ocaml.org/p/conan/=version` go if not contextualise by being on this page https://v3.ocaml.org/p/conan-lwt/0.0.1

JiaeK:
Just wanna confirm, so the tasks are
[...]

PR #183, opened 2021-11-19
Redirect versioned links from reverse dependencies

JiaeK:
This is an attempt to fix the part of reverse dependencies for #164.

### DIFF
--- a/src/ocamlorg/lib/package.ml
+++ b/src/ocamlorg/lib/package.ml
@@ -23,7 +23,8 @@ module Info = struct
     ; homepage : string list
     ; tags : string list
     ; dependencies : (OpamPackage.Name.t * string option) list
-    ; rev_deps : (OpamPackage.Name.t * string option * OpamPackage.Version.t) list
+    ; rev_deps :
+        (OpamPackage.Name.t * string option * OpamPackage.Version.t) list
     ; depopts : (OpamPackage.Name.t * string option) list
     ; conflicts : (OpamPackage.Name.t * string option) list
     ; url : url option
@@ -135,9 +136,7 @@ module Info = struct
             (OpamPackage.versions_of_name pkgs name)
             versions
         in
-        let latest =
-          OpamPackage.Version.Set.max_elt versions
-        in
+        let latest = OpamPackage.Version.Set.max_elt versions in
         let flags = OpamStd.String.Set.empty in
         let formula =
           OpamFormula.ands
--- a/src/ocamlorg/lib/package.ml
+++ b/src/ocamlorg/lib/package.ml
@@ -23,7 +23,7 @@ module Info = struct
     ; homepage : string list
     ; tags : string list
     ; dependencies : (OpamPackage.Name.t * string option) list
-    ; rev_deps : (OpamPackage.Name.t * string option * OpamPackage.t) list
+    ; rev_deps : (OpamPackage.Name.t * string option * OpamPackage.Version.t) list
     ; depopts : (OpamPackage.Name.t * string option) list
     ; conflicts : (OpamPackage.Name.t * string option) list
     ; url : url option
--- a/src/ocamlorg/lib/package.ml
+++ b/src/ocamlorg/lib/package.ml
@@ -23,7 +23,8 @@ module Info = struct
     ; homepage : string list
     ; tags : string list
     ; dependencies : (OpamPackage.Name.t * string option) list
-    ; rev_deps : (OpamPackage.Name.t * string option) list
+    ; rev_deps :
+        (OpamPackage.Name.t * string option * OpamPackage.Version.t) list
     ; depopts : (OpamPackage.Name.t * string option) list
     ; conflicts : (OpamPackage.Name.t * string option) list
     ; url : url option
@@ -135,6 +136,9 @@ module Info = struct
             (OpamPackage.versions_of_name pkgs name)
             versions
         in
+        let latest =
+          OpamPackage.create name (OpamPackage.Version.Map.max_binding versions)
+        in
         let flags = OpamStd.String.Set.empty in
         let formula =
           OpamFormula.ands
@@ -162,6 +166,7 @@ module Info = struct
   let make ~package ~packages ~rev_deps opam =
     let open Lwt.Syntax in
     let+ rev_deps = mk_revdeps package packages rev_deps in
+    let latest = OpamPackage.Version.Map.max_binding version_map in
     let open OpamFile.OPAM in
     { synopsis = synopsis opam |> Option.value ~default:"No synopsis"
     ; authors =
--- a/src/ocamlorg/lib/package.ml
+++ b/src/ocamlorg/lib/package.ml
@@ -136,7 +136,7 @@ module Info = struct
             versions
         in
         let latest =
-          OpamPackage.create name (OpamPackage.Version.Set.max_elt versions)
+          OpamPackage.Version.Set.max_elt versions
         in
         let flags = OpamStd.String.Set.empty in
         let formula =
--- a/src/ocamlorg/lib/package.ml
+++ b/src/ocamlorg/lib/package.ml
@@ -24,7 +24,7 @@ module Info = struct
     ; tags : string list
     ; dependencies : (OpamPackage.Name.t * string option) list
     ; rev_deps :
-        (OpamPackage.Name.t * string option * OpamPackage.Version.t) list
+        (OpamPackage.Name.t * string option * OpamPackage.t) list
     ; depopts : (OpamPackage.Name.t * string option) list
     ; conflicts : (OpamPackage.Name.t * string option) list
     ; url : url option
@@ -137,7 +137,7 @@ module Info = struct
             versions
         in
         let latest =
-          OpamPackage.create name (OpamPackage.Version.Map.max_binding versions)
+          OpamPackage.create name (OpamPackage.Version.Set.max_elt versions)
         in
         let flags = OpamStd.String.Set.empty in
         let formula =
@@ -156,7 +156,7 @@ module Info = struct
                    vf
                ]
         in
-        Lwt.return @@ ((name, string_of_formula formula) :: acc))
+        Lwt.return @@ ((name, string_of_formula formula, latest) :: acc))
       (OpamPackage.to_map
       @@ OpamStd.Option.default OpamPackage.Set.empty
       @@ OpamPackage.Map.find_opt pkg rdepends)
@@ -166,7 +166,6 @@ module Info = struct
   let make ~package ~packages ~rev_deps opam =
     let open Lwt.Syntax in
     let+ rev_deps = mk_revdeps package packages rev_deps in
-    let latest = OpamPackage.Version.Map.max_binding version_map in
     let open OpamFile.OPAM in
     { synopsis = synopsis opam |> Option.value ~default:"No synopsis"
     ; authors =
--- a/src/ocamlorg/lib/package.ml
+++ b/src/ocamlorg/lib/package.ml
@@ -23,8 +23,7 @@ module Info = struct
     ; homepage : string list
     ; tags : string list
     ; dependencies : (OpamPackage.Name.t * string option) list
-    ; rev_deps :
-        (OpamPackage.Name.t * string option * OpamPackage.t) list
+    ; rev_deps : (OpamPackage.Name.t * string option * OpamPackage.t) list
     ; depopts : (OpamPackage.Name.t * string option) list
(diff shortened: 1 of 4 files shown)


src/ocamlorg/lib/package.ml
```diff
@@ -135,6 +136,9 @@ module Info = struct
             (OpamPackage.versions_of_name pkgs name)
             versions
         in
+        let latest =
+          OpamPackage.create name (OpamPackage.Version.Map.max_binding versions)
```
JiaeK: `versions` is a wrong type to put but I don't know what to put instead.

src/ocamlorg/lib/package.ml
```diff
@@ -162,6 +166,7 @@ module Info = struct
   let make ~package ~packages ~rev_deps opam =
     let open Lwt.Syntax in
     let+ rev_deps = mk_revdeps package packages rev_deps in
+    let latest = OpamPackage.Version.Map.max_binding version_map in
```
JiaeK: Also I'm not confident with this `version_map`

src/ocamlorg/lib/package.ml
patricoferris: Yep, if we look at the error coming from the compiler we see:

```
File "src/ocamlorg/lib/package.ml", line 140, characters 71-79:
140 |           OpamPackage.create name (OpamPackage.Version.Map.max_binding versions)
                                                                             ^^^^^^^^
Error: This expression has type version_set
       but an expression was expected of type 'a Version.Map.t
```

The type `version_set` is just an alias to `Version.Set.t` (see https://v3.ocaml.org/p/opam-format/2.1.0/doc/OpamTypes/index.html#type-version_set). So we need to find the latest version from a set of versions not a map. If we then look at `OpamPackage.Version.Set` for an appropriate function for this, a little digging gives [OpamPackage.Version.Set.max_elt](https://v3.ocaml.org/p/opam-format/2.1.0/doc/OpamPackage/Version/Set/index.html#val-max_elt).

Also note that you have changed the type of `rev_deps` to be `(OpamPackage.Name.t * string option * OpamPackage.Version.t) list` where I assume the third element is the latest version which makes sense but if you were to use this value of `latest` it is a `OpamPackage.t` because you create one, you shouldn't need to create one.

src/ocamlorg/lib/package.ml
```diff
@@ -162,6 +166,7 @@ module Info = struct
   let make ~package ~packages ~rev_deps opam =
     let open Lwt.Syntax in
     let+ rev_deps = mk_revdeps package packages rev_deps in
```
patricoferris: Note also that you don't actually add the `latest` to the accumulation in the `mk_revdeps` function. It has: 

```ocaml
Lwt.return @@ ((name, string_of_formula formula) :: acc))
```

When it should add in the latest you calculated

```ocaml
Lwt.return @@ ((name, string_of_formula formula, latest) :: acc))
```

src/ocamlorg/lib/package.ml
patricoferris: This `latest` is unneeded and won't be used so you can delete it.

src/ocamlorg/lib/package.ml
```diff
@@ -23,7 +23,8 @@ module Info = struct
     ; homepage : string list
     ; tags : string list
     ; dependencies : (OpamPackage.Name.t * string option) list
-    ; rev_deps : (OpamPackage.Name.t * string option) list
+    ; rev_deps :
```
patricoferris, line 26: You'll need to expose this type change in the `package.mli` file too.

patricoferris (commented):
Thanks @JiaeK, all the pieces are here just needs some work :))

Once this part of the code is correct we'll then need to use the `latest` value in 

https://github.com/ocaml/v3.ocaml.org-server/blob/2135fc8215ea5bbb189d5285ae857b2cf3b1b9ec/src/ocamlorg_web/lib/templates/pages/package_template.eml#L37

src/ocamlorg/lib/package.ml
JiaeK: I see, I was sure about to put `OpamPackage.Version.Set.max_elt` but then since I put the type t of `(OpamPackage.Name.t * string option * OpamPackage.Version.t) list`, the compiler said we needed 'Version_map.key' so I got messed up :sweat_smile:

src/ocamlorg/lib/package.ml
JiaeK: That's why you said 'return' it. Noted!

JiaeK:
Thanks for the feedback, @patricoferris !
 
> Once this part of the code is correct we'll then need to use the `latest` value in
> 
> https://github.com/ocaml/v3.ocaml.org-server/blob/2135fc8215ea5bbb189d5285ae857b2cf3b1b9ec/src/ocamlorg_web/lib/templates/pages/package_template.eml#L37

According to compiler, `type` needs to change?

src/ocamlorg_web/lib/templates/pages/package_template.eml
```diff
@@ -34,8 +34,8 @@ let render ~readme ~license:_ package =
     | "" -> gettext "No License"
     | l -> l
   in
-  let rev_deps = package_info.Ocamlorg.Package.Info.rev_deps |> List.map (fun (name, x) ->
-      (Ocamlorg.Package.Name.to_string name, x)
+  let rev_deps = package_info.Ocamlorg.Package.Info.rev_deps |> List.map (fun (name, version) ->
+      (Ocamlorg.Package.Name.to_string name, Ocamlorg.Package.t)
```
JiaeK: It's just an attempt - I assumed it would be something like this?

patricoferris: You've turned `rev_deps` into a list of triples to accommodate the new latest version information, so when mapping over it we will need to at least do something with each. 

```ocaml
let rev_deps = package_info.Ocamlorg.Package.Info.rev_deps |> List.map (fun (name, x, version) ->
      (Ocamlorg.Package.Name.to_string name, x, (* Convert version to a string here *))
```

Note you will also have to change the code further below that uses the `rev_deps`:

```ocaml
<% rev_deps |> List.iter begin fun (name, cstr, version) -> %>
```

src/ocamlorg/lib/package.mli
```diff
@@ -46,7 +46,7 @@ module Info : sig
     ; homepage : string list
     ; tags : string list
     ; dependencies : (Name.t * string option) list
-    ; rev_deps : (Name.t * string option) list
+    ; rev_deps : (Name.t * string option * OpamPackage.t) list
```
patricoferris: I would just store the `Version.t` as an `OpamPackage.t` is a name and version but we already know the name from the first element of the triple.

src/ocamlorg_web/lib/templates/pages/package_template.eml
JiaeK: I see, thanks, and can I ask what does this 'x' mean? Using this when something is uncertain?

src/ocamlorg/lib/package.mli
JiaeK: Oh, I should have taken more time to notice that. Sorry.

src/ocamlorg_web/lib/templates/pages/package_template.eml
patricoferris: `x` here is just a (pretty undescriptive 😅 )  variable name for the second element of the triple. The second element being the optional string describing the version constraints (e.g. `">= 4.3.0"`) -- later you see it given the slightly more descriptive `cstr` name.

src/ocamlorg/lib/package.mli
JiaeK: After I changed it to `Version.t`, I got this error:

i3 p1

src/ocamlorg_web/lib/templates/pages/package_template.eml
JiaeK: Great, thank you :)

src/ocamlorg/lib/package.mli
patricoferris: Yep the types in `ml` and `mli` need to be the same :)) There's some explanation of `ml` and `mli` here: https://dev.realworldocaml.org/files-modules-and-programs.html#scrollNav-3

JiaeK: Thanks for the reference :) So that means we're taking back to `OpamPackage.t` ?

patricoferris: Not quite, it means updating the implementation to return just the version and not create the package. Concretely changing the type to match the `mli` https://github.com/JiaeK/v3.ocaml.org-server/blob/217f4a8173314f7cd307504d138d7080b42a830a/src/ocamlorg/lib/package.ml#L26 and update the `latest` calculation to only return the latest version https://github.com/JiaeK/v3.ocaml.org-server/blob/217f4a8173314f7cd307504d138d7080b42a830a/src/ocamlorg/lib/package.ml#L139

src/ocamlorg_web/lib/templates/pages/package_template.eml
```diff
@@ -161,14 +161,19 @@ let render ~readme ~license:_ package =
...
-                <% rev_deps |> List.iter begin fun (name, cstr) -> %>
+                <% rev_deps |> List.iter begin fun (name, cstr, version) -> %>
                 <li class="py-2">
                   <a href="/p/<%s name %>" class="mt-1 text-sm text-orange-600 hover:text-orange-500"><%s name %></a>
                   <% match cstr with None -> () | Some cstr -> %>
                   <span class="inline-flex items-center px-2.5 py-0.5 rounded-full text-xs font-medium bg-white text-gray-800">
                     <pre><%s cstr %></pre>
                   </span>
                   <% ; %>
+                  <% match version with None -> () | Some version -> %>
```
JiaeK: How can I write 'matching version'?

src/ocamlorg/lib/package.ml
```diff
@@ -23,7 +23,7 @@ module Info = struct
     ; homepage : string list
     ; tags : string list
     ; dependencies : (OpamPackage.Name.t * string option) list
-    ; rev_deps : (OpamPackage.Name.t * string option) list
+    ; rev_deps : (OpamPackage.Name.t * string option * OpamPackage.t) list
```
patricoferris: Note this probably still needs to be changed

```suggestion
    ; rev_deps : (OpamPackage.Name.t * string option * OpamPackage.Version.t) list
```

src/ocamlorg/lib/package.ml
```diff
@@ -135,6 +135,9 @@ module Info = struct
             (OpamPackage.versions_of_name pkgs name)
             versions
         in
+        let latest =
+          OpamPackage.create name (OpamPackage.Version.Set.max_elt versions)
```
patricoferris: And with the above change we only need the version

```suggestion
          OpamPackage.Version.Set.max_elt versions
```

src/ocamlorg_web/lib/templates/pages/package_template.eml
patricoferris: You shouldn't need to -- `version` at this point is a `string`. The match for `cstr` is because it is a `string option`.

JiaeK:
Thank you very much for your guidance @patricoferris. For the changelog, is it ok to put both of our IDs? I feel like it'd better to put your id :sweat_smile:

patricoferris:
No worries, great work (hopefully your learnt some things too :)) -- one final thing to do is use the version in the link rather than just adding the version number. Something like 

```
<a href="/p/<%s name ^ "/" ^ version %>" class="mt-1 text-sm text-orange-600 hover:text-orange-500"><%s name %></a>
```

should do the trick. You can test this locally for https://v3.ocaml.org/p/lwt/5.4.2 where the reverse dependency for `bap-server` is  `< "0.2.0"` so when we click we should go to `0.1.0` (if you try it now we get `bap-server.0.3.0`!)

>  is it ok to put both of our IDs?

I'm happy if you just put your own ID -- sometimes changelogs have a `reviewed by` section but we don't -- this is your work and PR :))

JiaeK:
That's very kind. And this indeed helped me to learn a lot about the rev_deps, thanks again. 

> You can test this locally for https://v3.ocaml.org/p/lwt/5.4.2 where the reverse dependency for `bap-server` is `< "0.2.0"` so when we click we should go to `0.1.0` (if you try it now we get `bap-server.0.3.0`!)
> 
Yeah, now it redirects to '0.1.0' :tada:

patricoferris (approved):
Great, lgtm!

PR #185, opened 2021-12-01
Fix versioned links from dependencies

JiaeK:
This PR fixes the dependencies part of #164.

There are still a few of uncleared spots, especially inside of `graphql.ml`, but I believe opening this PR draft would accelerate to fix this issue.

### DIFF
--- a/src/ocamlorg/lib/package.ml
+++ b/src/ocamlorg/lib/package.ml
@@ -22,7 +22,8 @@ module Info = struct
     ; license : string
     ; homepage : string list
     ; tags : string list
-    ; dependencies : (OpamPackage.Name.t * string option) list
+    ; dependencies :
+        (OpamPackage.Name.t * string option * OpamPackage.Version.t) list
     ; rev_deps :
         (OpamPackage.Name.t * string option * OpamPackage.Version.t) list
     ; depopts : (OpamPackage.Name.t * string option) list
@@ -49,7 +50,8 @@ module Info = struct
 
   let parse_formula =
     OpamFormula.fold_left
-      (fun lst (name, cstr) -> (name, string_of_formula cstr) :: lst)
+      (fun lst (name, cstr, latest) ->
+        (name, string_of_formula cstr, latest) :: lst)
       []
 
   let depends packages opams f =
@@ -68,6 +70,7 @@ module Info = struct
                 | _ ->
                   None
               in
+              let latest = OpamPackage.Version.Set.max_elt version_set in
               let deps =
                 OpamFormula.packages packages
                 @@ OpamFilter.filter_formula ~default:true env (f opam)
--- a/src/ocamlorg/lib/package.ml
+++ b/src/ocamlorg/lib/package.ml
@@ -69,7 +69,7 @@ module Info = struct
                 | _ ->
                   None
               in
-              
+              let latest = OpamPackage.Version.Set.max_elt versions in
               let deps =
                 OpamFormula.packages packages
                 @@ OpamFilter.filter_formula ~default:true env (f opam)
--- a/src/ocamlorg/lib/package.ml
+++ b/src/ocamlorg/lib/package.ml
@@ -69,7 +69,7 @@ module Info = struct
                 | _ ->
                   None
               in
-              let latest = OpamPackage.Version.Set.max_elt versions in
+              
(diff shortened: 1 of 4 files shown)


patricoferris (commented):
So this issue is quite a bit trickier than I first made out. But here's a rough plan of the pieces that need to be changed to support this:

#### In package.ml

You can leave `parse_formula` as is, for calculating `depopts` and friends it will need to be as it was, you can then get rid of `get_dependencies` because we can actually reuse `mk_revdeps` for dependencies now. The code in `mk_revdeps` is not specific to reverse dependencies because it does the calculation parameterised by the `rdepends` argument (of type `package_set package_map`). For doing the calculating for normal dependencies we just need to give it a `package_set package_map` for dependencies which we have already calculated here:

https://github.com/ocaml/v3.ocaml.org-server/blob/bb31c2835db299c89143b28743a2eddc20c22f0e/src/ocamlorg/lib/package.ml#L220

Which we can then pass in to `make` 

```ocaml
let+ t = make ~dependencies ~rev_deps ~packages ~package opam in
```

And use like `rev_deps` 

```ocaml
let+ rev_deps = mk_revdeps package packages rev_deps 
and+ dependencies = mk_revdeps package packages dependencies in
```

Feel free to rename that function. 

#### In graphql.ml

It looks like we'll have to extend `package_info` to include the latest version too, but make it optional so we can still pass `None` to functions that don't need it.

```ocaml
type package_info =
  { name : string
  ; constraints : string option
  ; latest : string option
  }
```

Let me know if that makes any sense :)

JiaeK:
Thanks for the response, @patricoferris :)

This does makes sense - especially in `graphql.ml`, I tried that before but after also adjusting `parse_formula`, as you mentioned, `depopts` and friends needed something else.


 &
> And use like `rev_deps`
> 
> ```ocaml
> let+ rev_deps = mk_revdeps package packages rev_deps 
> and+ dependencies = mk_revdeps package packages dependencies in
> ```
> 
> Feel free to rename that function.

I guess `and+` should be `let+`? And by renaming, you meant renaming `let+ rev_deps` and `and+ dependencies`?

And thanks for the explanation for `version`, `versions`, `version_set` - so now I think versions would be a better choice, somehow this doesn't still work.

patricoferris:
Sorry for the very slow response, completely missed your reply. 

> I guess and+ should be let+?

Not quite. See [these docs](https://v3.ocaml.org/p/lwt/5.5.0/doc/Lwt/Syntax/index.html). `and+` is syntax for `Lwt.both` whereas `let+` is syntax for `Lwt.map`. Twos binds are not the same as a bind and a both. For example:

```ocaml
let f () = 
  let+ a = Lwt.return 1 in
  let+ b = Lwt.return 2 in
  a + b

let g () =
  let+ a = Lwt.return 1
  and+ b = Lwt.return 2 in
  a + b
```

The function `f` has type `unit -> int Lwt.t Lwt.t` whereas `g` has type `unit -> int Lwt.t`. In `g` we're saying wait until promises `a` and `b` resolve before moving on. In `f` we have two `map`s which is functionally equivalent (I think) to:

```ocaml
let f () = 
  Lwt.map (fun a -> (Lwt.map (fun b -> a + b)) (Lwt.return 2)) (Lwt.return 1)
```

Hence we get this double-nested `Lwt.t`. 

> And by renaming, you meant renaming let+ rev_deps and and+ dependencies?

I think I meant renaming `mk_revdeps` since it could be used for more than just reverse dependencies, although I don't quite remember 😅.

> so now I think versions would be a better choice, somehow this doesn't still work.

What's going wrong?

JiaeK:
>Sorry for the very slow response, completely missed your reply. 

That's no problem at all & thanks for the reply! I'll study more about the syntax and semantics, and get back to you soon.

JiaeK:
Hi @patricoferris,
Sorry it took quite some time to get back - obviously I couldn't cover all the syntax but the memory of this PR seems like fading away so! 
Thanks for the explanation of `let+` & `and+` I adjusted along.

The current conflicts are about the two `versions` - `Error: Unbound value versions`
and it'd be great if I can get some tips to solve.

tmattio:
I'll close this for now. Feel free to re-open if you make progress on this @JiaeK 🙂

PR #188, opened 2021-12-15
New home toplevel

patricoferris:
This PR reintegrates the home page toplevel with the new design:

Screenshot 2021-12-15 at 10 28 11

It's not perfect and the simple regexp syntax highlighting could be expanded but it may be worth merging as an initial piece of work.


tmattio:
Thanks a lot @patricoferris, this is looking great!

I've just tried it and pushing the "Try" button doesn't create a textarea where I can type OCaml sentences, so I'm not sure how I can test it?

patricoferris:
Thanks for taking a look @tmattio! 

Hmm, seems to work for me 😅 -- did you run `dune clean && make build` I don't think the jsoo gets rebuilt with `make start` or `make watch` iirc? Perhaps it's a browser issue?

tmattio:
You're right, had to rebuild the toplevel manually, thanks! 🙂 

This is working great! I'll have a look at the code but sharing some thoughts in the meantime:

- We could add syntax highlighting for strings and module (For the later, everything starting with a capital letter should be pretty close I think?) if it's not too complicated
- We should remove the try button once we've clicked on it
- As you pointed out previously, syntax completion doesn't trigger a scroll down if the content overflows the content div. Is there any easy fix for this? Maybe we can call [`scroll`](https://developer.mozilla.org/en-US/docs/Web/API/Element/scroll) on the content div every time we write something to it? 

Also, I'm sure you had a good reason, but I'm curious to know why you had to remove the line numbers?

patricoferris:
> We could add syntax highlighting for strings and module

Yep, highlighting anything with a capital letter sounds fine given it'll be a module or a variant value constructor.

> We should remove the try button once we've clicked on it

Yep

> As you pointed out previously, syntax completion doesn't trigger a scroll down if the content overflows the content div. Is there any easy fix for this? Maybe we can call scroll on the content div every time we write something to it?

Yeah I thought this would fix it: https://github.com/ocaml/v3.ocaml.org-server/pull/188/commits/cdb753787fa535716f6ffe84814235ccb78a60f2 -- clearly not haha, I'll look into it :))

> Also, I'm sure you had a good reason, but I'm curious to know why you had to remove the line numbers?

Haha you give me too much credit! I couldn't figure out how to do it nicely, originally they were two separate `div`s so scrolling would be a pain. We could prefix anything we add to the output with a number but I'm not convinced a REPL like this needs numbers -- happy to try adding that though if you think it would look nice ?

patricoferris:
Haven't quite figured out the appropriate regex for capital identifiers atm, but strings are done, scrolling works and the button is hidden after clicking it.

Screenshot 2021-12-15 at 15 22 23

EDIT: looks like operators in strings are broken
EDIT EDIT: Fixed in https://github.com/ocaml/v3.ocaml.org-server/pull/188/commits/b3743065cc5f86df195887e72bb16ed51edc5cd9

tmattio (approved):
I quite liked the line numbers, but nothing blocking, we can always bring them back later. This LGTM whenever you're ready to merge 🙂

patricoferris:
Thanks @tmattio, a follow-up issue is #190

PR #189, opened 2021-12-15
Update dependencies

Julow:
also the build was broken

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -56,7 +56,7 @@
   fpath
   fmt
   logs
-  omd
+  (omd (= "2.0.0~alpha2"))
   opam-format
   timedesc
   yojson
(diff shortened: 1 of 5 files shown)


dune-project
```diff
@@ -77,4 +77,11 @@
    (>= 3.11.0))
   (js_of_ocaml-toplevel
    (>= 3.11.0))
-  (alcotest :with-test)))
+  (alcotest :with-test)
+  ; hilite
+  textmate-language
+  ; tools/ood-gen
```
tmattio, line 83: These are listed in the Makefile in the `DEV_DEPS` variable. The reason for that is: they are not dependencies of the project, they are only used to re-generate the data modules. So typically, we don't want them to be installed in Docker.

Julow: Perhaps these dependencies can be marked `{dev}` then.

Julow: I think `with-test` makes more sense, I've updated the PR.

tmattio: Cool, happy to have them with the `with-test` flag 🙂 
Can you also add the complete list of dependencies:
```
ppx_deriving_yaml
ezjsonm
ptime
fmt
fpath
piaf
lambdasoup
cmdliner
crunch
xmlm
uri
```

Julow: Updated!

tmattio:
Thanks! 🙂

PR #193, opened 2021-12-20
docs(workflows): bootstrapping

cdaringe:
# Problem

No problem per se, but the content could be organized more in [inverted pyramid](https://en.wikipedia.org/wiki/Inverted_pyramid_(journalism)) style, however that ports from journalism into technical writing :)

# Discussion

now the content flows with macro topics in this order:

1. use `dune init`
1. here's useful info for `dune init`
1. here's how to get dune installed
1. consider adding a dune-project file (note, the next version of dune init will emit dune-project files)
1. consider spin

i know spin is the original author's personal project, and it is _indeed_ a cool project. i did observe that it's not formally part of the [platform](https://ocaml.org/platform) yet. with respect to the author 🙇 , i'd at least _consider_ pulling `spin` out of this particular doc as a platform formal recommendation. if it is recommended formally, then by all means, please disregard.


tmattio:
Thank you! I agree that the best method's documentation could be improved, so the PR is very much welcome and a clear improvement 🙂 

It's indeed unfortunate to recommend a tool in the official best practices when it's not in the Platform. I've thought about that when writing the best practices, and the thought process was that there was no beginner-friendly way to bootstrap a project in the Platform (`dune-init` generated projects are not setting you up to be able to work on the project in your IDE for instance, so that can be a pain point for beginners).

Any thoughts @avsm?

I'll also take the occasion to mention that Spin's looking for co-maintainers. That's a requirement to be incubated in the Platform, so that's the only thing we're waiting for to land an RFC 😉

data/workflows/en/01_bootstrap_project.md
```diff
@@ -2,33 +2,38 @@
...
 
-Once you have successfully installed opam, you can install Dune with `opam install dune`.
+> dune init --help
+>
+> **dune init {library,executable,test,project} NAME [PATH]** initialize a
+> new dune component of the specified kind, named NAME, with fields
+> determined by the supplied options.
 
-This will install Dune's binary in your current Opam switch, so `dune` should now be in your `PATH`. If it not, you probably need to run `eval $(opam env)` to configure your current terminal with Opam environment.
+Once you have [installed opam](https://opam.ocaml.org/doc/Install.html), you can install Dune with `opam install dune`. This installs the Dune binary into your `opam` switch, so `dune` should now be in your `PATH`. If it is not, you probably need to run `eval $(opam env)` to configure your current terminal with your opam environment.
```
patricoferris: We should definitely link to installing opam, perhaps we should direct people to our documentation page i.e. https://v3.ocaml.org/tutorials/up-and-running-with-ocaml?

data/workflows/en/01_bootstrap_project.md
```diff
@@ -39,7 +44,6 @@ dune exec bin/main.exe
 
 Or you may be looking for the best way to get started with a specific kind of project:
 
-- A library
```
patricoferris, line 42: What was the reason for removing this? The other bullet points are more end-user applications (executables) -- do you think we need something more specific?

patricoferris (commented):
I agree with most of the points here :)) Thanks @cdaringe for taking the time to improve the docs 🚀

I would argue `dune init` does not "Bootstrap a Project" as @tmattio pointed out. In particular the "here's how to get dune installed" vs. what is possible with `spin` is miles apart w.r.t to a developer's experience getting started.

Personally, in the future, I'd like to see `dune init` pulled from the documentation and replaced with spin as the recommended way to get up and running :)) But as is pointed out, officially it's not in the platform just yet, so `dune init` remains.

avsm:
This seems like an excellent opportunity to come up with a consensus-driven public process for incubating a project in the Platform.  In my [2020 Platform talk](https://speakerdeck.com/avsm/ocaml-platform-2020?slide=18) I outlined a lightweight process and some criteria. To take spin, we would need to:

- to incubate it, check that it meets some minimum criteria for a community project (>1 maintainer, desire to increase adoption, willingness to merge good features from other similar projects that wish to join forces).
-  to graduate to active, needs to demonstrate it doesnt overlap with other active projects. In this case, once incubated we could request `dune init` be deprecated in favour of spin.

For now though, I agree that it's important to be consistent in the very first iteration of the website and not include non-platform tools. However, I'm entirely for noting an "at your own risk" set of bootstrap projects to try out here (such as spin, drom, carcass and other boilerplate generators through the years).

data/workflows/en/01_bootstrap_project.md
cdaringe: that's a great idea. how should we deal with the inner-linking URLs? pathnames only? get pathnames from a `Module`? `protocol:host:pathname`? you may already have prior art :)

data/workflows/en/01_bootstrap_project.md
cdaringe: ya, sure thing. my thoughts are as follows:

- the content above discuss using `dune init`
- the content shows `dune init library NAME PATH` from the `--help` dump

to state later in the document "if you need a library ..." right after mentioning tooling that can setup a library, I perceive as conflicting information.

counter point, the examples above used the `project` `kind` for `dune init`, so perhaps it's not super explicit. nonetheless, the `dune init` flow for a library kind is capable.

cdaringe:
hey @patricoferris

> I would argue dune init does not "Bootstrap a Project" as @tmattio pointed out

Perhaps we could refine terminology. What _is_ bootstrapping?

From my perspective, creating a subjectively-reasonable number of assets and asset-types to help get a user from zero to execution.

_After_ `dune init project NAME`, I can then run a new executable with a new linked library with about one-ish command. Not bad, IMHO. See output below.

```sh
~/src
❯ dune init project radness
Success: initialized project component named radness


~/src
❯ tree radness
radness
├── bin
│   ├── dune
│   └── main.ml
├── lib
│   └── dune
├── radness.opam
└── test
    ├── dune
    └── radness.ml

~/src
✦ ❯ cd radness && dune exe bin/main.exe
Info: Creating file dune-project with this contents:
| (lang dune 2.9)
| (name radness)
Hello, World!
```

It's certainly not as comprehensive as something `spin` may offer, ...but I'd suggest that it is indeed a non-trivial amount of bootstrapping.  Perhaps we make clear that this is somewhat minimal, and then as Anil suggested, articulate that more comprehensive solutions are available? That's pretty much what @tmattio already did :), perhaps without platform/non-platform delineation :)

data/workflows/en/01_bootstrap_project.md
patricoferris: Nothing that fancy yet :)) A pathname should suffice for the moment thanks

patricoferris:
@cdaringe well put, completely agree -- thanks for taking the time to explain. I think with a nod towards alternatives somewhere in the BKM this should be good to go?

Re:Platform, thanks @avsm for the ideas :)) I agree this would be a good opportunity to try and iron out the process for the platform. Probably some combination of porting over [what we have from v2](https://ocaml.org/platform/) and [other places about the platform](https://v3-explore-ocaml.netlify.app/platform/), an issue on this repo and a discuss post would help garner some community support and get the ball rolling.

cdaringe:
Took a third pass, based on above feedback

data/workflows/en/01_bootstrap_project.md
```diff
@@ -2,61 +2,59 @@
...
 
-`dune init proj` does not generate a `dune-project` for you, so you need to go in the generated project and create one:
+- "project" as the _kind_
+- "hello" as the _name_, and
+- "~/src/ocaml-projects" as the _path_ to generate the content in
 
-```
-echo "(lang dune 2.0)" > dune-project
-```
+The the `project` _kind_ creates a `library` in `./lib`, an `executable` in `./bin`, and links them together in `bin/dune`. Additionally, the command creates a test executable and an opam file.
```
patricoferris:
```suggestion
The `project` _kind_ creates a `library` in `./lib`, an `executable` in `./bin`, and links them together in `bin/dune`. Additionally, the command creates a test executable and an opam file.
```

data/workflows/en/01_bootstrap_project.md
```diff
@@ -2,61 +2,59 @@
...
 
+Hello, world!
 ```
-spin new bin my_project/
-```
-
-This will take some time, because Spin will install all of the dependencies in a new opam local switch (a.k.a local sandbox), which needs to compile an OCaml compiler.
 
-Once the project is generated, you can run the executable with `make start`.
+Thus, `dune init` can rapidly scaffold new projects, with minimal content. It can also be used to add incremental components (kinds) to existing projects.
```
patricoferris: If I understand this correctly should it be: 

```suggestion
Thus, `dune init` can rapidly scaffold new projects, with minimal content. It can also be used to add components (kinds) incrementally to existing projects.
```

patricoferris (commented):
Thanks @cdaringe this looks great, just a few nits and should be good to go :))

tmattio:
Thanks @cdaringe!

PR #194, opened 2021-12-20
Search by tag, authors (and synopsis, description...)

panglesd:
This PR implements a more expressive search for packages.

It adds support for `tag:`, `author:`, `name:`, `synopsis:`, `description:`. By defaults, it now does not take into account for the words order, except inside quoted strings.

For instance 
```
author:"alpha beta" tag:int tag:float "quoted string" html encoding
```

Clicking on an author name or on a tag redirects to the corresponding search.

Finally, the search query is put inside the input box to allow for editing the query.

### DIFF
--- a/src/ocamlorg_frontend/pages/packages_search.eml
+++ b/src/ocamlorg_frontend/pages/packages_search.eml
@@ -1,4 +1,4 @@
-let render ~total (packages : Package_intf.meta list)  (query : string) =
+let render ~total ~search (packages : Package_intf.meta list) =
 Layout.render
 ~turbo_full_reload:true
 ~title:"OCaml Packages · Search Result"
@@ -17,7 +17,7 @@ Layout.render
                         <div class="relative w-full text-gray-400 focus-within:text-gray-600">
                             <input name="q" id="q"
                                 class="w-full appearance-none outline-none focus:border-primary-600 text-body-600 border rounded-md py-3 pr-6 border-gray-200 pl-14 h-full"
-                                placeholder="Search packages" type="search" value="<%s Dream.from_percent_encoded query %>">
+                                placeholder="Search packages" type="search" value="<%s search %>">
                             <div class="absolute h-full flex items-center pl-6 opacity-60 left-0 top-0">
                                 <svg xmlns="http://www.w3.org/2000/svg" class="h-5 w-5 text-body-400" fill="none"
                                     viewBox="0 0 24 24" stroke="rgba(26, 32, 44, 1)">
--- a/src/ocamlorg_frontend/pages/packages_search.eml
+++ b/src/ocamlorg_frontend/pages/packages_search.eml
@@ -1,4 +1,4 @@
-let render ~total (packages : Package_intf.meta list) =
+let render ~total (packages : Package_intf.meta list)  (query : string) =
 Layout.render
 ~turbo_full_reload:true
 ~title:"OCaml Packages · Search Result"
@@ -17,7 +17,7 @@ Layout.render
                         <div class="relative w-full text-gray-400 focus-within:text-gray-600">
                             <input name="q" id="q"
                                 class="w-full appearance-none outline-none focus:border-primary-600 text-body-600 border rounded-md py-3 pr-6 border-gray-200 pl-14 h-full"
-                                placeholder="Search packages" type="search">
+                                placeholder="Search packages" type="search" value="<%s Dream.from_percent_encoded query %>">
                             <div class="absolute h-full flex items-center pl-6 opacity-60 left-0 top-0">
                                 <svg xmlns="http://www.w3.org/2000/svg" class="h-5 w-5 text-body-400" fill="none"
                                     viewBox="0 0 24 24" stroke="rgba(26, 32, 44, 1)">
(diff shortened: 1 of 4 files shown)


src/ocamlorg_frontend/pages/packages_search.eml
```diff
@@ -17,7 +17,7 @@ Layout.render
                         <div class="relative w-full text-gray-400 focus-within:text-gray-600">
                             <input name="q" id="q"
                                 class="w-full appearance-none outline-none focus:border-primary-600 text-body-600 border rounded-md py-3 pr-6 border-gray-200 pl-14 h-full"
-                                placeholder="Search packages" type="search">
+                                placeholder="Search packages" type="search" value="<%s Dream.from_percent_encoded query %>">
```
tmattio: Let's not depend on Dream here (we do through dream_eml of course, but let's assume that these two are separated).

The reason for that is that we want to stay as close as possible to HTML in the `ocamlfrontend_package` to ease the collaboration with frontend engineers who may not be comfortable with OCaml.

There are some exceptions to that (e.g. the package doc table of content is heavy on the OCaml code), but in this instance, we could pass the percent-encoded query to as an argument directly.

src/ocamlorg_frontend/pages/packages_search.eml
```diff
@@ -1,4 +1,4 @@
-let render ~total (packages : Package_intf.meta list) =
+let render ~total (packages : Package_intf.meta list)  (query : string) =
```
tmattio: We can make `query` a labeled argument

src/ocamlorg_frontend/pages/packages_search.eml
```diff
@@ -42,8 +42,7 @@ Layout.render
                         </div>
                         <div class="flex flex-wrap">
                             <% package.tags |> List.iter (fun (tag : string) -> %>
-                                <!-- TODO(tmattio): Support searching for tags -->
-                                <a href=""
+                                <a href="?q=tag:<%s Dream.to_percent_encoded ("\""^tag^"\"") %>"
```
tmattio, line 45: That's unfortunate that we have to use Dream here as well, but I don't see a reasonable alternative..

src/ocamlorg_package/lib/ocamlorg_package.ml
```diff
@@ -614,46 +614,183 @@ let module_map ~kind t =
...
+end = struct
+  type search_constraint =
+    | Tag of string
+    | Name of string
+    | Synopsis of string
+    | Description of string
+    | Author of string
+    | Any of string
+
+  type search_request = search_constraint list
```
tmattio, line 638:
```suggestion
  type t = search_constraint list
```
Maybe?

src/ocamlorg_package/lib/ocamlorg_package.ml
```diff
@@ -614,46 +614,183 @@ let module_map ~kind t =
...
+
+  let re =
+    let ascii_no_quote = Re.rep1 @@ Re.diff Re.ascii @@ Re.set "\"" in
```
tmattio, line 641: Were we depending on `Re` before? I don't see it being added in the opam file or the dune file.

src/ocamlorg_package/lib/ocamlorg_package.ml
```diff
@@ -614,46 +614,183 @@ let module_map ~kind t =
...
+    let unquoted = Re.group @@ Re.rep1 @@ Re.diff Re.ascii @@ Re.set " \"" in
+    let quoted = Re.seq [ Re.str "\""; Re.group ascii_no_quote; Re.str "\"" ] in
+    let option name =
+      Re.seq [ Re.group @@ Re.str name; Re.alt [ quoted; unquoted ] ]
+    in
+    let tag = option "tag:"
```
tmattio, line 647: Don't we need to list the other fields here (`description`, `synopsis`, etc.)?

tmattio (commented):
Thanks a lot for doing this!!

Apart from the few comments I left, one comment on the general approach I'd have is that we should maybe use exact matches for the special search query. I think it does that for the `tag` now, but we should be doing this for all the special queries. If we do this, we may not want to support the description and synopsis queries.

panglesd:
> we should maybe use exact matches for the special search query. I think it does that for the tag now, but we should be doing this for all the special queries. If we do this, we may not want to support the description and synopsis queries.

I don't think tags use "exact" (as opposed to "substring") matching, and I would say it's most of the time fortunate:
- we want `tag:mirage` to match `org:mirage` tags (there is no plain `mirage` tags)
- we want `tag:html` to match package `vg`, which has no `html` tag but an `html-canvas` tag
- we want `author:mattio` to match package with author `Thibaut Mattio`...

I agree that `description` and `synopsis` are not super useful and might never be used, I'm ok to remove them if you think it's better.

src/ocamlorg_package/lib/ocamlorg_package.ml
panglesd: Thanks! Somehow it shows that `description` and `synopsis` might not be very useful ;)

tmattio:
> I don't think tags use "exact" (as opposed to "substring") matching, and I would say it's most of the time fortunate:
>
> - we want tag:mirage to match org:mirage tags (there is no plain mirage tags)
> - we want tag:html to match package vg, which has no html tag but an html-canvas tag
> - we want author:mattio to match package with author Thibaut Mattio...

Mmh, I see where you're coming from. I don't agree, but I guess there are trade-offs to both approaches. FWIW, GitHub special queries use the exact matches approach, but one differentiating factor is the number of available results. Given the number of packages on opam-repository, I'm fine going with your approach if you think it's best.

> I agree that description and synopsis are not super useful and might never be used, I'm ok to remove them if you think it's better.

Yup, let's try to keep the special queries to a minimum until we know what we want. It's harder to remove them once they are used than adding them later.

panglesd:
I removed description and synopsis special queries.

I'm not that opposed to making special queries "exact matches". If you want, I'll let you have the last word on this! :slightly_smiling_face:

tmattio:
This looks very good like this, we can change the behavior later if the benefits become more apparent.

Merging now, thank you! 🎉

PR #195, opened 2021-12-22
Use leaflet.js for academic users map

patricoferris:
This PR vendors [leaflet.js]() and provides partial bindings using js_of_ocaml to the library in order to add an interactive map to the Academic Users page. There's a little library (`Carte`) that makes a simple map when given a bunch of markers, we might be able to reuse for things like events or industrial users?

We could also think about in the future providing a more interactive experience using the Leaflet controls to provide some form of "experience" to explore the different users. 

Some questions:
 - Should we use a different map, I'm not sure how many different ones there are from OSM?
 - Are the icons for the markers clear enough?
 - Should I have just created this using SVGs instead of pulling in Leaflet?

The initial map you see on the page:

Screenshot 2021-12-22 at 11 29 35

The markers have the information:

Screenshot 2021-12-22 at 11 30 34

Zoomed into Paris:

Screenshot 2021-12-22 at 11 30 52


patricoferris:
I forgot to remove the uses of the unreleased ppx, will do that shortly 

EDIT: fixed

tmattio:
Thanks a lot for the work on this @patricoferris! 🙂 

The implementation in the PR looks good, but we're losing the nice design of the map that we currently have. Is there a solution to use a custom map with leaflet.js?

If not, I think @asaadmahmood also had a solution to make the map interactive with plain JS (or rather AlpineJS). The dots on the map can also be made dynamic easily by generating the SVG from the locations of the academic users under the root `svg` element.

patricoferris:
Thanks for taking a look @tmattio :))

> but we're losing the nice design of the map that we currently have. Is there a solution to use a custom map with leaflet.js?

There are different places to get maps from (e.g. https://leafletjs.com/examples/choropleth/ is similar to the design, but requires an account). One thing I'd be wary of is the design looks great currently, but as you can see from these images the dots won't be as nicely distributed. For example, how do we show the three different institutions in Paris without reinventing leaflet?

patricoferris:
Putting on hold until later to see if we actually need the full functionality of (and dependency on) leaflet.

PR #196, opened 2021-12-24
Integrate blog and news

tmattio:
This PR integrates the blog and news pages.

The blog uses two types of data:

- The blog posts that we fetch from the different RSS feeds of the community (at the moment JaneStreet, Ahrefs and Tarides)
- News created by the community members, this includes releases of Platform tools, the OCaml newsletter, Multicore monthly, and OCaml Weekly News.

The OCaml newsletter and Multicore monthlies should be fetched from Discourse. At the moment, they have been manually added to the data, so this PR focuses on providing the content on Ood and the UI. The automation can come later.
The Platform news and OCaml Weekly News are largely incomplete and have been added here as examples. We'll need to integrate all the Platform blog once this PR is merged, and reach out to Alan Schmitt to see if and how we want to integrate the CWN in ocaml.org.

There still are some work to be done before we can merge this:

- [ ] Rework the UI of the blog, news and news post pages.
- [x] Handle pagination
- [ ] Handle listing blog posts and news by tags/categories
- [x] Fetch images from Discourse
- [x] Make the dates Human readable


tmattio:
There's some UI work remaining, but this can be done with follow-up PRs. This is ready to merge I think.

tool/ood-gen/lib/news.ml
```diff
@@ -1,142 +1,38 @@
...
       { title = metadata.title
-      ; slug = Utils.slugify metadata.title
+      ; slug
       ; description = metadata.description
-      ; url = metadata.url
       ; date = metadata.date
-      ; preview_image = metadata.preview_image
-      ; body_html = String.trim body
+      ; tags = metadata.tags
+      ; body_html = Omd.to_html (Omd.of_string (String.trim body))
```
patricoferris: Might be worth a `Hilite` call here to get the OCaml code blocks (which I assume would be fairly frequent). Alternatively, it might be better to do this when saving the markdown file since it will replace code blocks with HTML which is still valid markdown? Up to you, we can leave it as is too :))

patricoferris (approved):
Thanks @tmattio this looks great ! 

As you noted we should maybe do some UI work, like the pagination going from `1 - 12` at the moment :)) But that can be done in a follow-up PR 👍

tool/ood-gen/lib/news.ml
tmattio: Good idea, will do that!

PR #197, opened 2021-12-24
feat: compress by default

cdaringe:
# Problem

Page and assets are large

# Solution

Compress by default.

An unscientific diff shows there are hundreds of kBs of savings just on the homepage:

image


cdaringe:
The file failing lint is unchanged

patricoferris:
Thanks @cdaringe -- is there any noticeable extra time added to serving the page? Afaict there's no caching of compressed assets in `dream-encoding`, I was wondering if we should instead compress things at build time so we don't do this on every single request?

tmattio:
Thanks @cdaringe, this is great and the analysis provided is very useful!

Even if we compress everything, let's keep the middlewares in the router, so we can just apply the middleware to every route.

> Afaict there's no caching of compressed assets in dream-encoding, I was wondering if we should instead compress things at build time so we don't do this on every single request?

That would be great! I think that's something we could implement in `crunch` if we wanted (cc @hannesm). The PR does seem like an improvement even without it though, so I think we could do it separately.

I'd be very surprised if we ended up with a longer TTI (Time to Interactive), it'd be nice to check though. Could we generate a lighthouse report before and after the PR @cdaringe?

hannesm:
@tmattio sure, go ahead. I've not much stake in crunch.

cdaringe:
likely no different. i suspect the different numbers are just noise:

before:

image

after:

image

tmattio:
Thanks a lot @cdaringe 🙂

PR #199, opened 2021-12-25
feat: local switch

cdaringe:
# Problem

fresh cloners may not be usin' the right switch!

# Solution

- default to the local switch!
- fix lint issue broken in main

# Discussion

Not required :) would be nice to either accept it or perhaps add `.vscode` to gitignore

### DIFF
--- a/.vscode/settings.json
+++ b/.vscode/settings.json
@@ -1,0 +1,6 @@
+{
+  "ocaml.sandbox": {
+    "kind": "opam",
+    "switch": "${workspaceFolder:v3.ocaml.org-server}"
+  }
+}
--- a/CHANGES.md
+++ b/CHANGES.md
@@ -2,6 +2,8 @@
 
 - Adding search by authors, tag and other package fields (#194, by @panglesd)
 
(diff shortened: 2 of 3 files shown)


tmattio:
Maybe I'm old school, but I tend to like not having IDE-specific files in the repo (no VSCode config, and the gitignore rules can be set at the user level, rather than at the repo level).

Not a very strong opinion though, I'll let others chime in 🙂

cdaringe:
do what you see fit. i won't be upset either way :)

my take is that repositories are for collaboration, and i'm all for making collaboration as easy as possible. IDE assets are _prettty_ painless. having the editor ready to go (vscode, .idea, .editorconfig, whatever, any/all of 'em) reduces friction

tmattio:
IDE setup changes for everyone. For opam users especially: some opam users use global switches, some use opam link, and other use local switches. I can imagine other people using esy sandboxes as well, since it now support opam files.

I agree the recommended installation method should be to create a local switch, and this is documented in the CONTRIBUTING guide, but I don't think we should add IDE configuration files in the repo.

I'm closing the PR, but feel free to continue the discussion if you disagree.

PR #200, opened 2021-12-26
feat: locked dependencies

### LINKED ISSUE #198: format fails as default cmds do not produce reproducible switch
# Problem

- `make switch`
- _...do work..._
- `make fmt`

```sh
ocamlformat: Error while parsing /Users/cdaringe/src/v3.ocaml.org-server/_build/default/.ocamlformat:
             For option "version": expecting "0.20.1" but got "0.20.0"
```

# Discussion

- opam really needs to [just use lockfiles by default](https://discuss.ocaml.org/t/should-opam-use-a-lockfile-by-default/4251/16)
- `opam lock .`?

cdaringe:
# Problem

Closes #198 

# Solution

- use `opam switch import/export`
- `opam lock . -y` doesn't actually produce a pinned opam lockfile :/ . i'll probably open an issue over there


tmattio:
Thanks for the PR @cdaringe 🙂 

IIUC, this is fixing the dependencies installation by having `make deps` install a locked version of the dependencies. I think a simpler solution is to install the correct version of the dependencies in the `opam` file or the `Makefile`. So running `make deps` should install `ocamlformat.0.20.0` instead of the latest version.

Locking dependencies in itself could be useful to ensure reproducible builds in production (or at least builds that use the same version of the dependencies), but I don't think this is what this PR does (the Dockerfile doesn't use the locked file)

cdaringe:
> this is fixing the dependencies installation by having make deps install a locked version of the dependencies

correct

>  I think a simpler solution is to install the correct version of the dependencies in the opam file or the Makefile

Yes, at least until this (or similar) happens again. 

Aside, I spend an inordinate amount of time in ocaml projects trying to get switches aligned with compatible dependencies. lockfiles fix the issue near permanently, ~99%. i really wish opam did what Cargo.lock, package-lock.json, Pipfile.lock, gradle.lock, yarn.lock, etc etc do, right off the bat. It's been demotivating, for me personally, contributing to opam based projects.

> Locking dependencies in itself could be useful to ensure reproducible builds in production

true. id be willing to patch the dockerfile as well.

seems like our community is pretty fractured on lockfile alignment still.

tmattio:
> seems like our community is pretty fractured on lockfile alignment still.

Indeed, but to be clear, I'm on the side of using lockfiles by default 😉 

I just think the issue you're fixing here should be addressed in another way. Taking the example of npm for instance, even with the lock file, you need to specify the correct version of your dependencies in the `package.json` files. And most users will run `npm install` and not `npm ci` (the name of the command hint at the main use cases!) to install the deps, so the default developer workflow is to not use the lockfile.

> Aside, I spend an inordinate amount of time in ocaml projects trying to get switches aligned with compatible dependencies. lockfiles fix the issue near permanently, ~99%. i really wish opam did what Cargo.lock, package-lock.json, Pipfile.lock, gradle.lock, yarn.lock, etc etc do, right off the bat. It's been demotivating, for me personally, contributing to opam based projects.

Your experience is very unfortunate. I'd be curious to see where the problems you're encountering come from. The opam package manager runs matrix builds with all lower and upper bounds of every dependencies and reverse dependencies when a package is updated, so I'm expecting that installing the dependencies for a project should be successful a large majority of the time, and indeed it's been my experience. Are you mainly contributing to projects not released on the opam repository (like this one)? In any case, I'd love to get an idea of the pain points you're encountering. As someone working on improving developer experience with OCaml, this would be very valuable insights. Let's continue this discussion offline maybe 🙂

tmattio:
I've opened #263 which, apart from formatting the code, pins ocamlformat in the Makefile.
This should address the specific issue you are concerned about with the PR.

While I agree there's a lot of improvement that can be done with reproducible builds and locking dependencies in opam, I'm not sure using export/import to work around that is a good solution. In particular, locked dependencies isn't something you use as a default installation source with any other package managers I can think of (npm, mix, cargo all use the deps constraints and re-generate the lock files when installing the deps)

I'm closing the PR for now, but let's reconsider if we have more data points indicating that we should push a lock file to the repo.

Issue #201, opened 2021-12-30
A11y: low contrast orange


patricoferris:
Through out the site we use an orange background with white text (e.g. for buttons) or the inverse (e.g. headings). As with the V2 version of the site (see https://github.com/ocaml/ocaml.org/issues/1199), this has a low contrast and fails web accessibility standards tests. 

More concretely the "Get Started" button on the top right has a background colour of `#ee6a1a` with white text. Using something like the [WebAIM contrast checker it scores 3.12](https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=EE6A1A.) (ideally it would be greater than or equal to 7 or at least 4.5). Using the tool this gives two colours:

 - Contrast ratio 4.5: `#C2540F` (see [test](https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=C2540F))
 - Contrast ratio 7: `#943F0A` (see [test](https://webaim.org/resources/contrastchecker/?fcolor=FFFFFF&bcolor=943F0A))

Which are heading towards brown more than orange. Any thoughts on a design to improve this (note this often applies to the links in text too which are orange on a pale background)?

sabine:
Resolved in the meantime.

PR #202, opened 2021-12-30
Accessibility improvements

patricoferris:
This PR fixes a few accessibility issues on a couple of the main pages, primarily it:

 - Fixes some `<hX>` orderings (see: https://dequeuniversity.com/rules/axe/3.3/heading-order)
 - Increases the contrast on some text (see: https://www.w3.org/WAI/WCAG21/Understanding/contrast-minimum.html)
 - Fixes duplicated input ids (see: https://www.w3.org/WAI/WCAG21/Understanding/parsing.html)
 - Discernible names for links (see: https://www.w3.org/WAI/WCAG21/Understanding/link-purpose-in-context.html)

One problem is the low contrast of the orange and white used throughout the site, see #201 for thoughts.

### DIFF
--- a/src/ocamlorg_frontend/layouts/layout.eml
+++ b/src/ocamlorg_frontend/layouts/layout.eml
@@ -144,7 +144,7 @@ inner =
         <li>
 
           <div class="space-x-6 text-2xl flex items-center">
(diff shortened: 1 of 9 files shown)


tmattio:
@asaadmahmood feel free to review when you have time, but I'll merge now, this LGTM.

PR #204, opened 2022-01-06
Add number stats on the Packages page

Julow:
The stats are:

- Total number of packages (not counting every versions)
- New packages added this month (basically new directories in `packages/<new>`)
- Updates this week (new directories in `packages/<every packages>/<new>`)

There is some refactoring so should be reviewed commit by commit.

It looks like this:
image

The "packages" page isn't done yet, missing (in following PRs):
- "Listing" stats. We thought about these stats with @tmattio :
  + Packages with the most number of rev-deps (except Dune and virtual packages)
  + Newest packages
  + Recently updated packages

- Featured packages
  I'm not sure what that mean yet, perhaps a static list of packages ?


tmattio:
> Featured packages
> I'm not sure what that mean yet, perhaps a static list of packages ?

Yup, that was the idea. These can be added to the data as a yml file, and we'll be able to rotate the featured packages frequently (with PR from the community, or when we see an announcement on Discuss, etc.)

src/ocamlorg_web/lib/handler.ml
```diff
@@ -305,7 +305,19 @@ type package_kind =
...
-let packages _req = Dream.html (Ocamlorg_frontend.packages [])
+let packages state _req =
+  let stats =
+    match Ocamlorg_package.packages_stats state with
+    | Some
+        { Ocamlorg_package.Packages_stats.nb_packages
+        ; nb_update_week
+        ; nb_packages_month
+        } ->
+      Some { Ocamlorg_frontend.nb_packages; nb_update_week; nb_packages_month }
```
tmattio, line 316: We don't have to create a new structure here I think, we can just pass the arguments to the function.

The less OCaml code in the frontend, the better, and using multiple arguments instead of a new type definition and a record is a bit lighter.

Julow: That's two different records with the same fields. This is needed because the `ocamlorg_frontend` and `ocamlorg_package` libraries don't communicate. The same thing happens for the `package` type, this is probably done on purpose.
Making the libraries depend on each other might be worse because we might loose the ability to build the values. To really avoid the conversion, the best solution is to move the types in a separate library.

The record is needed because it is wrapped in an `option`. I think it's also easier to read that way.

tmattio:
Thanks a lot @Julow 🙂

PR #213, opened 2022-01-13
Packages: Add listing stats

Julow:
The stats are:

- Newest packages
  Computed from the Git history.

- Most recently updated packages
  Computed from the Git history too.

- Packages with the biggest number of revdeps
  The rev-deps for each packages were already computed, so this one is easy.

There's a data point rendered next to each package with an icon (number of rev-deps, package addition date, lastest version).

Computing these statistics isn't that time-consuming, it takes 200ms on my laptop, compared to 35s for the other package computations.

### DIFF
--- a/src/ocamlorg_frontend/components/icons.eml
+++ b/src/ocamlorg_frontend/components/icons.eml
@@ -1,0 +1,21 @@
+(** Using icons from {:https://heroicons.com/} *)
+
+let downloads =
+  <svg xmlns="http://www.w3.org/2000/svg" class="h-5 w-5 pr-1" fill="none" viewBox="0 0 24 24" stroke="currentColor">
+    <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M4 16v1a3 3 0 003 3h10a3 3 0 003-3v-1m-4-4l-4 4m0 0l-4-4m4 4V4" />
+  </svg>
+
+let calendar =
+  <svg xmlns="http://www.w3.org/2000/svg" class="h-5 w-5 pr-1" fill="none" viewBox="0 0 24 24" stroke="currentColor">
+    <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M8 7V3m8 4V3m-9 8h10M5 21h14a2 2 0 002-2V7a2 2 0 00-2-2H5a2 2 0 00-2 2v12a2 2 0 002 2z" />
+  </svg>
+
+let refresh =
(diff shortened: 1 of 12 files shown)


src/ocamlorg_frontend/components/icons.eml
```diff
@@ -0,0 +1,21 @@
+(** Using icons from {:https://heroicons.com/} *)
+
+let downloads =
+  <svg xmlns="http://www.w3.org/2000/svg" class="h-5 w-5 pr-1" fill="none" viewBox="0 0 24 24" stroke="currentColor">
```
tmattio, line 4: One problem with extracting the icons (and the main reason we haven't done that) is that they aren't reusable if you hardcode the size here.

It should be possible to make them w-full and h-full and encode the size on the parent element though. I'd try this approach if you want to extract the icons.

tmattio:
This is great, thanks @Julow 🙂 
Left a comment for the extraction of icons, but this LGTM otherwise

src/ocamlorg_frontend/components/icons.eml
Julow: I thought about this and decided that it'll be better when we have a second place we want to the icon.
I think we don't want too many different sizes across the website so we probably want the sizes defined in `icons.eml`. Not sure how to allow different sizes for now

tmattio:
> I thought about this and decided that it'll be better when we have a second place we want to the icon.
I think we don't want too many different sizes across the website so we probably want the sizes defined in icons.eml. Not sure how to allow different sizes for now

Ok, let's merge now then. I think the structure the eml doc could be updated slightly to make it easier to follow for frontend developers (i.e. working on it without knowledge of OCaml), but let's merge now and I'll take care of this if needed.

Thanks again, this is great 🎉

patricoferris:
This is great, thanks @Julow :)))!

src/ocamlorg_frontend/pages/packages.eml
```diff
@@ -115,171 +116,52 @@ Layout.render
     </div>
 </div>
 <div class="bg-white py-20 lg:py-32">
+% let render_package_stat pkg render_extra_stats =
+    <a href=""
```
patricoferris, line 120: I wonder if for all of them we could just link to `https://v3.ocaml.org/p/<package>` ?

patricoferris (commented):
I know it's merged haha, just a thought?

tmattio:
> I know it's merged haha, just a thought?

Sorry @patricoferris, got excited and merged very hastily 😄

src/ocamlorg_frontend/pages/packages.eml
Julow: Good catch!

PR #221, opened 2022-01-14
Home: Fix curated resources

Julow:
- Write the synopsis for the cards.
- The "Application Guides" card becomes "99 Problems" because we don't have that page yet
- Fix the manual and book cards


tmattio (commented):
That's great, thanks!

Can we replace the lorem ipsum in the academic users and industrial pages as well? This should be the same content

Julow:
Done !

tmattio:
Looks great, thanks a lot @Julow 🎉

PR #225, opened 2022-01-14
Fix a few typos

patricoferris:
WIP PR as I go through fixing some typos


data/workflows/en/03_update_deps.md
```diff
@@ -58,4 +58,4 @@ build: [
 ]
 ```
 
-Either way, once you have added your dependency in the appropriate file, you can run `opam install . --deps-only` to update your current switche's dependencies.
+Either way, once you have added your dependency in the appropriate file, you can run `opam install . --deps-only` to update your current switch's dependencies.
```
avsm:
```suggestion
Either way, once you have added your dependency in the appropriate file, you can run `opam install . --deps-only` to update your current switch dependencies.
```

I think just 'switch dependencies' reads better?

PR #226, opened 2022-01-14
Tutorials: Fix containers complexity

Julow:
- Remove uninteresting or not well defined complexities (length, finding
  an element)
- Rename "accessing cell `i`" into "random access"
- ~~The "returning the number of elements" complexity for Set and Map was
  wrong~~
- The "adding an element" complexity for Hashtbl and Buffer was wrong


Julow:
Actually the complexity for Map and Set wasn't wrong.

christinerose (approved):

PR #239, opened 2022-01-16
Adding compiler video

asaadmahmood:


tmattio:
Thanks @asaadmahmood!

Is there a way to convert the mp4 to a GIF? 1.3MB is rather large, so I'm wondering if converting to a GIF could help with the asset size.

tmattio:
I've tried converting the video to GIF, and couldn't find a good balance between the quality and the size of the asset.

We definitely can't add a 1.4MB video to the generated OCaml modules, so I don't think having a demo as a GIF or a video is going to be a viable option.

Another idea to replace the image for the Fast build system would be to add some benchmark that shows how fast it is. I'll try to see if we can find something for this. Closing the PR in the meantime.

PR #243, opened 2022-01-17
Update ahrefs logo

Khady:
Ahrefs' logo has been changed. Latest version can be obtained at https://ahrefs.com/media-kit. This PR updates all occurrences I could find.


Khady:
is it possible to use the svg logo everywhere? It would make things easier

tmattio:
Thanks a lot @Khady! I've cleaned up the unused assets and we're now using the SVG for Ahrefs logo in both the cloud logo and the success story.

PR #245, opened 2022-01-17
Add ahrefs to jobs

Khady:


tmattio:
Thanks @Khady 🙂 

I've regenerated the static content and fixed a bug with the line breaks in the yml files.

I've also changed the location of the job to "Remote", as we don't have a way to specify multiple locations for a job at the moment (PR welcome though!), and we use it to group opportunities by locations.

Let me know if you prefer to use Singapore instead, I'll update that.

Khady:
remote is fine. Thanks!

src/ocamlorg_data/job.ml
```diff
@@ -13,7 +13,7 @@ type t =
   
 let all = 
 [
-  { id = 1
+  { id = 2
```
patricoferris, line 16: @tmattio looks like the `id`s are not stable and they probably should be, if someone shares a link to a job opportunity (e.g. https://v3.ocaml.org/opportunities/1) it now points to the wrong job.

If `jobs.yml` is an append-only list then we can just number them in order and add a field to indicate the opening is no longer available?

tmattio: > If jobs.yml is an append-only list then we can just number them in order and add a field to indicate the opening is no longer available?

That's how it's supposed to work, will have a look 🙂

tmattio: Oh I see what happened there! It was expected that users will add job posts at the top of the list: we reverse the list to populate the IDs, and reverse it back once populated.

I doubt that there are any links to the job posts at the moment, so I think we can keep the IDs currently in `main`, and maybe add a comment to the YAML file to ask users to add the posts at the top of the list.

PR #248, opened 2022-01-19
Name mix fix

dinakajoy:
Fixed a JavaScript reference error: Java was used instead of JavaScript


patricoferris:
Thanks @dinakajoy your use of camel-case in `JavaScript` looks like the correct way to do it so I fixed the `Javascript` above :)) Once CI is happy I'll merge

PR #250, opened 2022-01-24
Toplevel Playground

patricoferris:
This is a draft PR to add support for a toplevel playground with a [codemirror 6](https://codemirror.net/6/) editor. It is still very much a WIP.

### Preview

The editor uses the same logic as the homepage editor for running examples of code. This PR also extends the RPC API to include a `typecheck` method. This combined with codemirror's lint library allows us to do real-time typechecking giving a merlin-like experience.

typing

It is actually pretty quick, the lag you see in the gif before the message pops up is me trying to find my mouse 😅  The gif is also a little outdated, I have subsequently removed the `Problems` section as they are now reported to the linter.

### Tasks 

 - [x] Vendor the codemirror libraries we use rather than via `npm` -- The JS libraries are vendored statically in https://github.com/patricoferris/jsoo-code-mirror/tree/static
 - [x] We might want to re-add the base64 logic to share code snippets.
 - [x] Switch to using https://github.com/jonludlam/js_top_worker via a pin and upstream type-checking logic.
 
 Follow up PR:
  - [ ] The old UI code had to separate editors for mobile and desktop, it would be a lot more convenient to only have one editor I think which I think should be possible with the right tailwind mangling.


patricoferris:
autocompleting

This PR now also supports code-mirror+UTOP powered auto-completion. Previously it was just what [mllike](https://github.com/codemirror/legacy-modes/blob/main/mode/mllike.js) had, but now it is whatever the worker has (so if other libraries are loaded it should autocomplete for those too).

patricoferris:
Apologies for the size of this PR, but it should now be ready for review. There are two main changes:

 - The library for running the toplevel in a worker and communicating with OCaml RPC now exists separately at https://github.com/jonludlam/js_top_worker and we use it pinned here.
 - The playground uses Codemirror 6 -- I'm maintaining a `static` branch of the [OCaml bindings that vendors/bundles](https://github.com/patricoferris/jsoo-code-mirror/tree/static) the Javascript libraries so we need only add the library here not constantly using the npm packages --- this should in theory be safer and less likely to randomly break. For the curious it uses browserify to bundle and then babel to translate the JS into something jsoo is happy with.

src/ocamlorg_frontend/layouts/layout.eml
```diff
@@ -51,8 +52,141 @@ inner =
 
 <body x-data="{ open: false, sidebar: window.innerWidth > 1024 && true, showOnMobile: false}"
   class="bg-global-default" @resize.window="sidebar = window.innerWidth > 1024">
+<% if playground then ( %>
+<header class="h-20 flex items-center header-black">
```
tmattio, line 56: The HTML structure for the playground and the rest of the pages should be the same. The difference should be in the classes.

Actually, tailwind has a `dark:` variant for this, so the classes that differ from the light theme should be added in the same HTML elements with the `dark:` variant.

We want to force the playground to be dark though for now, as we don't have a light theme for it. But I'm sure Alpine or Tailwind have something to do this.

src/ocamlorg_frontend/layouts/layout.eml
```diff
@@ -51,8 +52,141 @@ inner =
...
+      </div>
+      <div class="flex items-center space-x-8">
+        <div class="hidden lg:flex">
+          <a href="<%s Url.getting_started %>"><button class="btn btn-sm">Get started</button></a>
+        </div>
+
+        <div
+          class="hamburger lg:hidden flex h-12 w-12 hover:bg-primary-100 flex items-center justify-center rounded-full"
+          x-on:click="open = ! open">
+          <dib class="text-gray-100">
```
tmattio, line 79:
```suggestion
          <div class="text-gray-100">
```

src/ocamlorg_frontend/layouts/layout.eml
```diff
@@ -193,9 +331,13 @@ inner =
     </div>
     </div>
   </header>
+<% ); %>
+
   <main>
     <%s! inner %>
   </main>
+  <% if playground then ( %>
+  <% ) else ( %>
```
tmattio, line 340:
```suggestion
  <% if not playground then ( %>
```

src/ocamlorg_frontend/pages/playground.eml
```diff
@@ -0,0 +1,53 @@
+let render () =
+Layout.render
+~turbo_full_reload:true
```
tmattio, line 3: Makes me think: we should remove turbolink for now, until we have a better integration. Out of scope for this PR of course.

src/ocamlorg_frontend/url.ml
```diff
@@ -38,6 +38,7 @@ let news_post v = "/news/" ^ v
 let opportunities = "/opportunities"
 let opportunity v = "/opportunities/" ^ v
 let carbon_footprint = "/carbon-footprint"
+let playground = "/playground"
```
tmattio, line 41:
```suggestion
let playground = "/play"
```

src/ocamlorg_toplevel/bin/dune
```diff
@@ -104,14 +75,27 @@
...
   (deps
+   ../../playground/index.bc.js
    ../toplevel_stdlib.bc.js
    ../toplevel_base.bc.js
    ../worker.js
    ../worker_base.js)
   (action
    (progn
+    (run
+     npx
```
tmattio, line 93: I'm surprised you managed to run `npx` there. I always had trouble running it from `dune` 🤷

src/ocamlorg_toplevel/lib/toplevel.ml
```diff
@@ -144,18 +135,27 @@ let timeout_container worker () =
...
-  let worker =
-    try Worker.create (Jstr.v s)
-    with Jv.Error _ -> failwith "Failed to created worker"
-  in
+let or_raise = function
+  | Ok v -> v
+  | Error (Toplevel_api.InternalError e) -> failwith e
+
+let initialise s callback =
+  let rpc = Js_top_worker_client.start s 100000 callback in
```
tmattio, line 143: What is `100000` here?

src/ocamlorg_toplevel/playground/edit.ml
```diff
@@ -0,0 +1,34 @@
+open Code_mirror
+open Brr
+
+let basic_setup = Jv.get Jv.global "__CM__basic_setup" |> Extension.of_jv
```
tmattio, line 4: What is this variable?

tmattio (commented):
Thanks a lot for the fantastic work @patricoferris 🎉 

I've reviewed everything except `playground/index.ml`, which I'll try to review more carefully.
Everything else LGTM overall and I've left a few comments for the minor things I noticed.

I tried to run the playground, and I'm seeing an empty editor with the following error in the console:

```
Error: <svg> attribute height: Expected length, "auto".
playground:146 Error: <svg> attribute height: Expected length, "auto".
playground:153 Error: <svg> attribute height: Expected length, "auto".
DevTools failed to load source map: Could not load content for chrome-extension://ljngjbnaijcbncmcnjfhigebomdlkcjo/content-script/browser-polyfill.min.js.map: HTTP error: status code 404, net::ERR_UNKNOWN_URL_SCHEME
playground:83 GET http://localhost:8080/img/logo-footer.svg 404 (Not Found)
playground.js:54 Uncaught TypeError: Cannot read properties of undefined
    at playground.js:54
    at playground.js:54
    at playground.js:1
    at playground.js:152
    at playground.js:152
(anonymous) @ playground.js:54
(anonymous) @ playground.js:54
(anonymous) @ playground.js:1
(anonymous) @ playground.js:152
(anonymous) @ playground.js:152
playground.js:54 Uncaught TypeError: Cannot read properties of undefined
    at playground.js:54
    at playground.js:54
    at playground.js:1
    at playground.js:152
    at playground.js:152
```

Perhaps something went wrong during the rebase? (I've rebased on top of `main` again and pushed to `playground` if you want)

IIRC, when I tried it a few days ago, I noticed these:

- Typing Cmd+Enter doesn't run the code. We either need to handle this or change the Run button to remove the keyboard shortcut. The former would be nicer, but I'm not sure how complicated it is with Code Mirror
- The code-sharing didn't work. Copying the generated URL in another tab didn't create an editor with the new code.
- When sharing the code, we should add some visual indicators and copy the URL to the user's clipboard. We can take inspiration from the UX on https://play.tailwindcss.com/

tmattio:
Superseded by #411, I rebased and addressed the review comments there.

PR #256, opened 2022-01-26
packages: Fix crash when package is removed

Julow:
The query for detecting new files has been changed to using git log, which has the disadvantage of returning paths to file that have been added then removed.

### DIFF
--- a/src/ocamlorg_package/lib/ocamlorg_package.ml
+++ b/src/ocamlorg_package/lib/ocamlorg_package.ml
@@ -48,9 +48,8 @@ let state_of_package_list (pkgs : t list) =
   ; featured_packages = None
   }
 
-let read_versions package_name =
+let read_versions package_name versions =
   let open Lwt.Syntax in
-  let versions = Opam_repository.list_package_versions package_name in
   Lwt_list.fold_left_s
     (fun acc package_version ->
       match OpamPackage.of_string_opt package_version with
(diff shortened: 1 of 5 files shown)


src/ocamlorg_package/lib/ocamlorg_package.ml
```diff
@@ -73,7 +72,15 @@ let read_packages () =
...
         Lwt.return acc
       | _name ->
-        let+ versions = read_versions package_name in
+        let versions =
+          match Opam_repository.list_package_versions package_name with
+          | Some v ->
+            v
+          | None ->
+            []
+          (* TODO: Opam_repository.list_packages_and_versions *)
```
tmattio, line 81: Do we want to keep this TODO? We may want something more descriptive if that's the case.

tmattio (commented):
LGTM, thanks a lot for fixing this!

I just had a small comment for the TODO in `ocamlorg_package.ml`

tmattio:
Merging the PR now since the package documentation is broken now. Feel free to comment for the TODO after the merge.

src/ocamlorg_package/lib/ocamlorg_package.ml
Julow: I think I wrote that TODO for myself, I intended to implement it before pushing.

Julow: Fixed here: https://github.com/ocaml/v3.ocaml.org-server/pull/257

PR #257, opened 2022-01-26
packages: Minor refactoring of listing functions

Julow:
I left a TODO in a previous PR and this what was supposed to added instead.

### DIFF
--- a/src/ocamlorg_package/lib/opam_repository.ml
+++ b/src/ocamlorg_package/lib/opam_repository.ml
@@ -72,23 +72,38 @@ let last_commit () =
   in
   output
 
-let ls_dir directory =
+let fold_dir f acc directory =
   match Sys.readdir directory with
   | exception Sys_error _ ->
     None
   | entries ->
-    let entry_is_dir x = Sys.is_directory (Filename.concat directory x) in
-    Some (List.filter entry_is_dir (Array.to_list entries))
-
-let list_packages () =
-  match ls_dir Fpath.(to_string (clone_path / "packages")) with
-  | Some pkgs ->
-    pkgs
-  | None ->
-    []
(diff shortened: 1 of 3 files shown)


src/ocamlorg_package/lib/opam_repository.ml
```diff
@@ -72,23 +72,38 @@ let last_commit () =
...
-    Some (List.filter entry_is_dir (Array.to_list entries))
-
-let list_packages () =
-  match ls_dir Fpath.(to_string (clone_path / "packages")) with
-  | Some pkgs ->
-    pkgs
-  | None ->
-    []
+    (* Sort to remove a source of non-determinism. *)
+    Array.sort String.compare entries;
```
Julow, line 81: This is basically useless because these values are either inserted in a `Map` or thrown away immediately. But I really don't like leaving this unsorted.

src/ocamlorg_package/lib/opam_repository.mli
```diff
@@ -7,8 +7,9 @@ val pull : unit -> unit Lwt.t
 val last_commit : unit -> string Lwt.t
 (** Get the latest commit of the opam repository. *)
 
-val list_packages : unit -> string list
```
tmattio, line 10: I think we can keep this function, it seems useful, even if we have a function `list_packages_and_versions`.

Julow: I removed it because it's not used anymore. I don't believe it can be ever useful because we want to use the cached and processed list of packages we already keep in memory. These private functions are only used to refresh this internal cache and won't be used outside of `Ocamlorg_package` and `Packages_stats`.
I've added it back but suggest reverting.

PR #259, opened 2022-01-26
Make the manual and standard library easily discoverable

tmattio:
Addresses some feedback from @glennsl on Discuss.

This PR:

- Adds links to the manual and Stdlib API from the learn page's navigation bar
- Change the curated resources on the homepage, the industrial users and academic users to make it clear what manual is, and add a link to the standard library API.

@glennsl - thanks for the fantastic feedback on Discuss 🙂 Let me know what you think of these changes, do they make the access to the Standard library easier for power users? Note that this isn't linking to the API from the manual (https://ocaml.org/api/index.html), but to the one generated by the package documentation (https://v3.ocaml.org/p/ocaml-base-compiler/4.13.1/doc/index.html), which may need some polishing as well, happy to hear your toughts on this as well!

On Learn:
image

On the Homepage:
image


glennsl:
Awesome, thank you! While I still think learning resources and reference documentation serve completely different uses and should not be conflated, this does go most of the way to addressing the problem. And once you learn that "Documentation" = "Learn" (somewhat ironically :grin:) it's easy enough to ignore whatever else is in there.

> Note that this isn't linking to the API from the manual (https://ocaml.org/api/index.html), but to the one generated by the package documentation (https://v3.ocaml.org/p/ocaml-base-compiler/4.13.1/doc/index.html), which may need some polishing as well, happy to hear your toughts on this as well!

Ah, yeah this is unfortunately quite a step backwards. While I much prefer this _style_ of documentation, there's just way too much unordered information here. The TOC on the left is almost completely unusable because of that, and the first page is almost as bad. There's (somewhat confusingly) a TOC on the right too however, which on the surface is much more usable. Unfortunately it's just internal page navigation, so clicking `stdlib` there just scrolls you down the page to that section, and you have to click on `Stdlib` again to get to the actual documentation.

But even there you have a way to go to get to the actual module you want documentation for. The modules are listed in the TOC on the left, but again that's obscured by a mountain of irrelevant information. The TOC on the right again lists the sections of the current page, but unfortunately there are too many sections here for it to fit on my desktop screen, and because it's sticky I'm not able to scroll down to it either. Even if I could, however, that would be another two clicks to get to the module I'm interested in.

In total, that's 4 clicks (once I've figured out where to click) to get to the documentation for the `List` module, for example, compared to a single click in the old API documentation. And there's also no search functionality.

It's definitely something to build on, but I think it needs quite some heavy polishing before I'd use this over the old API documentation, which functionally works quite well despite its quirks.

glennsl:
Actually, if the library nodes in the left TOC were initially collapsed, that alone would make it quite usable. It's still three clicks, but it's three easy clicks in rapid succession, no page loading or moving the mouse cursor to different parts of the page between clicks.

Also I really like that items show anchor links on hover. That's something that has annoyed me for a long time when I have to link to something in the old API documentation, having to go into the source to find the `id` of an item and manually adding it to the URL.

Octachron:
I don't think that linking to the compiler-base-package documentation is a good idea in its current state: there are far too much emphasis on the compilerlibs library  (which is only useful for few expert users). Even worse, even in the standard library section, six of the seven toplevel modules are internal modules that should not be that visible.

tmattio:
Thanks for the review @glennsl and @Octachron 🙂 

> I don't think that linking to the compiler-base-package documentation is a good idea in its current state

Do you recommend serving https://ocaml.org/api/index.html for now then? Or should we link to the `Stdlib` module directly: https://v3.ocaml.org/p/ocaml-base-compiler/4.13.1/doc/Stdlib/index.html?

Here's how the page looks if we collapse the package TOC as @glennsl suggested:

image

Octachron:
The current page  https://ocaml.org/api/index.html at least presents a list of modules (even if I am not sure why the order is that random, but that can be fixed). I fear that the  current version of https://v3.ocaml.org/p/ocaml-base-compiler/4.13.1/doc/Stdlib/index.html would be mostly noise for non-expert users: the libraries in the toc are mostly unhelpful (half of them are compiler libraries), and the list of standard library modules is at the bottom of the page. People will leave the page before finding this list.

I think that the standard library will require a hand-written entry point. If I remember correctly, the documentation builder will pick an `index.mld` file as a custom entry point, doesn't it?

tmattio:
I've updated the PR to link to the current API documentation (https://ocaml.org/api/index.html) for now, as it seems more usable than the one on the package doc 🙂 

> I think that the standard library will require a hand-written entry point. If I remember correctly, the documentation builder will pick an index.mld file as a custom entry point, doesn't it?

That's right!

tmattio:
Merging now as I think this is already an improvement on the discoverability of the manual and standard library.
There's still a lot of room for improvement, of course, so we'll keep experimenting on this while we work on the documentation after launch.

In the meantime, please keep the feedback coming, this is really helpful! 🙂

PR #274, opened 2022-02-04
Add a footer in the tutorials to contribute or ask questions

tmattio:
This adds a footer to the documentation pages (heavily inspired by GitHub's documentation) to contribute or ask questions to the community.

cc @yawaramin @samoht

image

### DIFF
--- a/src/ocamlorg_data/tutorial.ml
+++ b/src/ocamlorg_data/tutorial.ml
@@ -7,6 +7,7 @@ type difficulty =
 
 type t =
   { title : string
+  ; fpath : string
   ; slug : string
   ; description : string
   ; date : string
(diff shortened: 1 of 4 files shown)


src/ocamlorg_data/tutorial.ml
```diff
@@ -20,6 +21,7 @@ type t =
 let all = 
 [
   { title = {js|Up and Running with OCaml|js}
+  ; fpath = {js|tutorials/en/000_up_and_running.md|js}
```
yawaramin, line 24: Is `{js|...|js}` a vestige of when the frontend was in ReScript? I think it doesn't have any meaning in standard OCaml?

yawaramin (approved):

src/ocamlorg_data/tutorial.ml
tmattio: That was indeed introduced for compatibility with ReScript. When trying to remove the `js` part last time, I found that some strings contained the `|}` character, so the strings were escaped unintentionally.

Keeping the `js` namespace seemed like a good workaround 🙂 

Thanks for the review!

samoht: It's indeed a bit odd to not just use `".."` here. But not super important and could be done in another PR.

PR #279, opened 2022-02-07
Syndicated RSS feed

tmattio:
This PR updates the way we scrape the data, and in particular the RSS feeds for the blog.

As reported in https://github.com/ocaml/v3.ocaml.org-server/issues/242, maintaining a white list of IDs to use from the different RSS feeds creates a lot of friction and we can't expect users to open a PR whenever they publish a new blog post.

Instead, we now fetch all the articles, and we will be reaching out to the different RSS feeds maintainers to ask to create a feed specific to OCaml if we see that they are publishing articles unrelated to OCaml.

This also adds https://hannes.robur.coop/atom as a new source for the blog. This addition wasn't immediately possible because ocamlrss, the library we were using to parse the RSS feeds, doesn't support the Atom format.
To work around this, we update river (https://github.com/kayceesrk/river), a library that builds on top of `syndic` and support both RSS and Atom feeds).
We vendor the updated version of river in this PR.

We also add a new GitHub action that runs every day to scrape the content. By doing so, users can simply add their feed url in `data/rss-sources.yml`, and every new article they publish will trigger a PR on OCaml.org to add the new content.

This workflow is a bit heavy, but the alternative to fetch the RSS feeds at runtime and cache them, didn't seem better.

To recap, for a user to add their RSS or Atom feed:
- They can open a PR adding their URL in `data/rss-sources.yml`
- Once merged, the CI will check every day for new content from the feed and will open a PR with the new scraped content if anything changed
- When we merge the PR coming from the CI, the website is published with the new content

@hannesm let me know what you think of this workflow

TODO:
- [x] Add back the featured blog posts
- [ ] List unrelated blog posts, filter out, and reach out to feed maintainers


hannesm:
Thanks for your work on that, this looks great. I'd appreciate to have the workflow, as you describe in the PR comment above, on the website (so that anyone who wants to contribute RSS sources knows how to proceeed).

tmattio:
>  I'd appreciate to have the workflow, as you describe in the PR comment above, on the website (so that anyone who wants to contribute RSS sources knows how to proceeed).

Makes sense! I've created #288 to track this. I'll work on this in a follow-up PR 🙂

PR #299, opened 2022-02-08
Make package search more compact

tmattio:


tmattio:
@MirzaBabarBaig Thanks a lot for the fix 🙂 

The density looks much better, but I'm seeing a few regressions:

- The content goes beyond the container on mobile
image

- The search results don't look very nice when we have a lot of authors or a long license name
image

For the later, perhaps we can limit the number of authors to list, or truncate the content

Issue #232, opened 2022-01-14
docs broken link: module index


avsm:
https://v3.ocaml.org/p/tezos-rpc/11.0/doc/index.html

The 'RPC_description'  module link in the bar on the left is broken - it seems to be missing the module prefix in the url.

PR #302, opened 2022-02-09
Fix 404 on absent docs

### LINKED ISSUE #262: Tezos documentation is not generated

jonludlam:
If the status of the docs is anything other than 'success', then don't have the documentation tab link anywhere.

Fixes #262 (sort of)

### DIFF
--- a/src/ocamlorg_frontend/layouts/package_layout.eml
+++ b/src/ocamlorg_frontend/layouts/package_layout.eml
@@ -53,7 +53,7 @@ Layout.render
               Overview
             </button>
           </a>
-          <a href="<%s Url.package_doc package.name package.version "index.html" %>">
+          <a href="<%s match documentation_status with | `Success -> Url.package_doc package.name package.version "index.html" | _ -> "#" %>">
             <button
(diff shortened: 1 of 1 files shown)


tmattio:
Thanks @jonludlam 🙂 

I think it may be better for user experience to disable the tab instead of removing it completely.
If the tab is disabled, we can add a tooltip saying the documentation is not available for some reason.
If the tab is not there, it may be confusing for users: why doesn't this package have a documentation tab?

Any idea why the tezos doc wasn't generated in the first place?

jonludlam:
Sorry, my comment wasn't clear - I meant that the tab is still there, but clicking on it doesn't take you anywhere else. We retain the red, yellow and green dots indicating the presence of the docs.

I'm not sure why the docs haven't been created - it looks like a pipeline weirdness to me. @pitag-ha and I noticed a similar issue earlier when dealing with issue #232 .

tmattio:
I think showing the error when the doc page is unavailable is quite useful (e.g. https://v3.ocaml.org/p/ocaml-top/1.1.5/doc/index.html)

Closing this for now, but let me know if you think we shouldn't show the CI errors in the documentation.

PR #304, opened 2022-02-09
Make release links absolute to v2.ocaml.org

### LINKED ISSUE #165: Releases links are broken
The releases have been imported as-is and the relative links are currently broken. We need to update the links to redirect to the correct URL.

tmattio:
Fixes #165 

This redirects to v2.ocaml.org. So let's wait for the domain to be available before merging.


tmattio:
> This redirects to v2.ocaml.org. So let's wait for the domain to be available before merging.

The links were already broken. Let's merge now.

PR #313, opened 2022-02-13
Enable odoc 2.1's new formatting

jonludlam:
Odoc 2.1 is more careful in its formatting of long signatures and
types. To enable this we need to use the following css rule:

```css
code {
  white-space: pre-wrap;
}
```

This assumes a width of 80 characters, which is close to the maximum
width of the doc pane (on my laptop). The switch to mobile-style layout
happens well after the width goes below 80 and we get unintentional
wrapping.

One possibility to address this is to remove this rule below a
particular width and fall back to the original formatting. Other
possibilities exist!

### DIFF
--- a/src/ocamlorg_frontend/css/other/doc.css
+++ b/src/ocamlorg_frontend/css/other/doc.css
@@ -22,6 +22,10 @@ div.odoc div.spec code::after {
   content: "";
 }
 
+div.odoc div.spec code {
+  white-space: pre-wrap;
+}
+
 div.odoc a.anchor {
   position: absolute;
   left: 0px;


jonludlam:
Before:
Screenshot 2022-02-13 at 10 14 53
After:
Screenshot 2022-02-13 at 10 14 25

jonludlam:
Also before:
Screenshot 2022-02-13 at 10 24 00
After:
Screenshot 2022-02-13 at 10 24 10

tmattio:
This looks great, thank you!

> This assumes a width of 80 characters, which is close to the maximum
> width of the doc pane (on my laptop). The switch to mobile-style layout
> happens well after the width goes below 80 and we get unintentional
> wrapping.
> 
> One possibility to address this is to remove this rule below a
> particular width and fall back to the original formatting. Other
> possibilities exist!

Would you mind opening an issue for this, with a link to a page where you are able to reproduce the issue?

PR #316, opened 2022-02-14
Migrate HTTP client to Hyper

tmattio:
This PR refactors the different libraries to use Hyper as an Http client.

The initial motivation was to add a new endpoint for the GitHub avatars that would act as a proxy to the GitHub servers so that clients don't have to query the GitHub servers themselves and be exposed to eventual tracking and data collection.

As hyper uses the same types as Dream, it seemed to be a good candidate to implement this proxy easily.

The PR is not ready to merge though. I'm seeing some runtime errors when querying the documentation servers.
```
14.02.22 11:09:36.050       dream.log  INFO REQ 1 GET /p/dream/1.0.0~alpha3 127.0.0.1:56737 Mozilla/5.0 (Macintosh; Intel Mac OS X 10_16
14.02.22 11:09:36.050                  INFO REQ 1 GET https://docs-data.ocaml.org/live/p/dream/1.0.0~alpha3/doc/README.md.html
14.02.22 11:09:36.102       dream.log  WARN REQ 1 Aborted by: Unix.Unix_error(Unix.ECONNREFUSED, "connect", "")
14.02.22 11:09:36.102       dream.log  WARN Raised at Lwt_unix.connect.(fun) in file "src/unix/lwt_unix.cppo.ml", line 1832, characters3
14.02.22 11:09:36.102       dream.log  WARN Called from Lwt_unix.retry_syscall in file "src/unix/lwt_unix.cppo.ml", line 509, character4
14.02.22 11:09:36.102       dream.log  WARN Re-raised at Hyper__http__Connect.alpn_https_tcp.(fun) in file "src/http/connect.ml", line 0
14.02.22 11:09:36.102       dream.log  WARN Re-raised at Hyper__http__Connect.no_pool.(fun) in file "src/http/connect.ml", line 156, ch5
14.02.22 11:09:36.102       dream.log  WARN Re-raised at Dream_encoding.compress.(fun) in file "lib/dream_encoding.ml", line 185, chara1
14.02.22 11:09:36.102      dream.http ERROR REQ 1 Unix.Unix_error(Unix.ECONNREFUSED, "connect", "")
14.02.22 11:09:36.103      dream.http ERROR REQ 1 Raised at Stdlib__Map.Make.find in file "map.ml", line 137, characters 10-25
14.02.22 11:09:36.103      dream.http ERROR REQ 1 Called from Logs.Tag.find in file "src/logs.ml", line 154, characters 14-32
```

I'll investigate further.

cc @aantron - the diff should speak for itself, but this is simplifying a lot of the code that needed to make HTTP requests. The Hyper API is simple, intuitive and flexible enough for what we needed here with the `request/run` pair 🙂


aantron:
Great!

I want to mention immediately that Hyper had a "pre-release release," mainly to:

1. Unblock some experimental projects (like this PR, if it can be considered experimental!).
2. Release Dream and Hyper in their first, minimal, reasonably stable and mutually-consistent state, because I was getting very burdened by the huge divergence between Dream development branch and the last Dream release, which was last May.

I wouldn't suggest Hyper for "real-world" use yet, or announce it. Most notably, it doesn't check server certificates yet, which is of course a critical security issue (though relatively easy to implement code-wise, and something I intend to address soon).

Speaking to the actual issue, the output looks like what I would expect if the connection was refused (if you think the output needs to be cleaner, please let me know). Nothing jumps out at me from your code, and I'm not sure if connection refused was a legitimate outcome or not.

I would suggest not to use `Hyper.request`, `Hyper.run` directly for now, but stick to `Hyper.get`, `Hyper.post`, mainly because I will likely change the types of the former values.

aantron:
I also see that your current client appears to follow redirects. `Hyper.get` [does that as well](https://github.com/aantron/hyper/blob/a16193c5369d379772765b549a393090d639373e/src/hyper.ml#L56-L65) (it composes a [redirect-following middleware](https://github.com/aantron/hyper/blob/a16193c5369d379772765b549a393090d639373e/src/logic/redirect.ml#L22) onto `Hyper.run`).

aantron:
Try specifying the port explicitly (`:443`). Looking at the Hyper code now, it looks like I lost port 443 in one of the refactorings, and Hyper is likely trying port 80 (did not confirm).

tmattio:
> I wouldn't suggest Hyper for "real-world" use yet, or announce it. Most notably, it doesn't check server certificates yet, which is of course a critical security issue (though relatively easy to implement code-wise, and something I intend to address soon).

CoHttp doesn't check the certificates either AFAICT, so this is on par with the security level we had using CoHttp. But that seems fine as we currently have only two uses cases for HTTP clients:
- Scrape the RSS feeds and watch videos. Both of these are done offline, and are only there to update the files in `data/`. This doesn't have any impact on the users.
- Query the documentation. It was on the same server until very recently, and there is a plan to have the documentation server on the same machine hosting OCaml.org again. So the security risk here is minimal (the HTTP request never leaves the machine).

I'm feeling rather confident that Hyper is more robust than our custom HTTP client with a custom follow-redirect implemented on top of CoHttp. I'm happy to not merge this PR if you feel like Hyper is not ready, but I'm not feeling too concerned myself given the above and compared to what we had before.

> Try specifying the port explicitly (:443). Looking at the Hyper code now, it looks like I lost port 443 in one of the refactorings, and Hyper is likely trying port 80 (did not confirm).

Adding the port in the URL does the trick 🙂 

> I would suggest not to use Hyper.request, Hyper.run directly for now, but stick to Hyper.get, Hyper.post, mainly because I will likely change the types of the former values.

I needed that in this PR because I needed to handle non-successful return status. There wasn't any way to do this with the `get/post` API I think. However, I'm happy to pin Hyper to a specific commit (or tag) and fix compilation errors in case of breaking changes in future versions.

dune-project
```diff
@@ -45,12 +45,11 @@
...
   (ocamlorg-data
    (= :version))
-  cohttp
-  cohttp-lwt-unix
   bos
   crunch
   mirage-kv-mem
-  dream
+  (dream
+   (>= ~alpha3))
```
tmattio: Not the correct version

ocamlorg.opam
```diff
@@ -83,9 +77,7 @@ build: [
   ]
 ]
 pin-depends: [
+  ["hyper.dev" "git+https://github.com/aantron/hyper#master"]
```
tmattio: Remove now that hyper is on opam

aantron:
Fair enough. Your comment addressed all my warnings and concerns. I'll fix the https 443 issue (https://github.com/aantron/hyper/issues/2) in the immediate future. As for getting the actual response, that is very valid. I think the top-level `Hyper.get` will eventually have a custom exception it can throw that can carry a non-`200 OK` response. At the same time, from the other direction, all the helpers that `Hyper.get` will eventually use will also be exposed for you to call as you call `Hyper.run` now, so you will be able to use either style.

tmattio:
I'll keep the branch around, but closing this for now. I'll bring it back to life when https://github.com/aantron/hyper/issues/5 is fixed

PR #319, opened 2022-02-15
Add back Bloomberg

code-ghalib:
Adding back the Bloomberg logo per discussion [over here](https://github.com/ocaml/ocaml.org/discussions/312#discussioncomment-2166677)


tmattio:
Thanks @code-ghalib this looks good to me!

As per your suggestion in the discussion you linked to, I agree we should have better heuristics to decide which logos are displayed on the homepage, but hopefully, there won't be any controversy to add back previously removed logos, so merging now 🙂

PR #325, opened 2022-02-21
Fix "Up and running" tutorial

shonfeder:
The section on project initialization was using the wrong command for project generation (`dune init exec` is only for generating a single component, not for generating a project), and is doesn't work as written anymore, since dune devs changed the behavior of `build` so that it no longer crates a `dune-project` file if one isn't found.

I've also tried to fill in some of the context around project structure, without getting too into the weeds.


data/tutorials/en/000_up_and_running.md
```diff
@@ -127,53 +127,82 @@ Let's begin the simplest project with Dune and OCaml. We create a new directory
...
 $ ls
-_build		dune		helloworld.ml
+bin  _build  dune-project  helloworld.opam  lib  test
 ```
 
-The `helloworld.exe` executable is stored inside the `_build/default` subdirectory, so
+All the build outputs generated by dune go in the `_build` directory. The 
+`helloworld.exe` executable is stored inside the `_build/default/bin/` subdirectory, so
 it's easier to run with `dune exec`. To ship the executable, we can just copy
 it from inside `_build/default` to somewhere else.
```
shonfeder:
```suggestion
it from inside `_build/default/bin` to somewhere else.
```

gasche:
Thanks for working on this! Could we get just the minimal fix, and discuss the other additions as another PR? It's not obvious what the "minimal" fix is because `dune init project` generates a very different layout from `dune init exec`, so the explanations after that need invasive changes. But I would say "not adding entirely new content".

shonfeder:
The current docs give the wrong way to create projects with dune. This gives the right way. 

Could you please be specific about which parts you’d like removed from this PR? 

PR feedback that amounts to “no, not like that” is not very helpful or encouraging to people doing free labor.

shonfeder:
Maybe this whole section should just be removed and replaced with a link to https://github.com/ocaml/ocaml.org/blob/5e6174ca3fd106c8bf4baf8286ac775ddb0b0a85/data/workflows/en/01_bootstrap_project.md? That one has the correct instructions.

gasche:
My thoughts:
- the "Up and Running" piece should stay short and to the point. I don't know about the idea of linking to another page in the middle of it, I think it disrupts the flow. (I'm also not sure about showing the Dune file, etc.; too technical for the audience.)
- After more discussion with @shonfeder I'm now convinced that we don't want to use `dune init exec` here, it was not meant for this and its behavior, today or tomorrow, is not good enough. `dune init project` sounds okay (as long as it's clear it does not work like `git init`, it should not be run from the directory itself)
- If @cdaringe and @patricoferris and @tmattio and @avsm worked on good bootstrapping instructions over at #193, maybe a subset of them could lend a hand to do something minimal and good here? I'm not fond of @shonfeder's current proposal, but I agree that something needs fixing.

shonfeder:
I can remove the dune project file dump. That was only meant to follow the existing example (which showed the dune file content already). I don’t know what else you don’t like about these changes tho, so I don’t know how else I’d improve it.

But it sounds like your preference is to have someone else make the changes? Should I just close the PR then, or image there some concrete changes that would make it acceptable to you?

gasche:
> But it sounds like your preference is to have someone else make the changes?

Sorry if that came out wrong, I thought that maybe you preferred to work just on fixing the ` dune init exec` issue, and not get into a long discussion on the rest of the changes.

I agree that the previous version would show the `dune` file, but that was 2 lines in the section. With your version, the `dune-project` file is 25 lines, it refers to a specific Dune version so it may need an update from time to time, and you felt the need to add another 12 lines of explanation for its content. I don't know for sure if the changes are good or bad, but that's what I call "not minimal", now we have to think about the change in context to understand if they are "worth it".

gasche:
Another thing: I'm not sure about the change from `dune exec ./helloworld.exe` (now `dune exec ./bin/helloworld.exe`) to `dune exec helloworld`. The latter feels more magical to me, it goes in the direction of teaching the user about the local layout versus the installed/public names, etc. I think I would show `dune exec ./bin/helloworld.exe` first, which is also an opportunity to let the user guess where to find the code, and `dune exec helloworld` later or not at all (in the context of a quick tutorial).

gasche:
So: on first approach, my recommendation would be to have things roughly as follows:

```
$ dune init project helloworld

$ dune build

$ dune exec ./bin/helloworld.exe

$ ls
```

with explanations along the way, and that's it, and a link to other documents for more. (The idea is that the users follows along, walking through the basic steps only, and that it will read the other documents *later*.)

I think that this is more in spirit with the current content. If you want to expand the content significantly, as in your PR, that could be a separate discussion. (And I don't feel particularly competent to give my opinion on that part.)

shonfeder:
I I think stripped out everything mentioned here.

data/tutorials/en/000_up_and_running.md
```diff
@@ -123,57 +123,54 @@ OCaml compilation. We installed it with opam, above.
...
-file, which tells dune how to build the program, and a `_build` subdirectory,
-which is dune's working space.
+Or, alternatively,
 
 ```
-$ ls
-_build		dune		helloworld.ml
+$ dune exec hellworld
+Hello, World!
 ```
```
gasche, line 154: I'm not convinced by this "alternatively", because I think it is too magic (where does the mapping from `helloworld` to `./bin/main.exe` come from? Does Dune implement the convention that `<dir>/bin/main.ml` are also runnable as `dune exec <dir>`?)

gasche (commented):
The new state looks very reasonable to me. Thanks!

gasche:
@tmattio I wonder if you would be interested in merging this; the tutorial as currently formulated is incorrect in its `dune` usage instructions and beginners get confused.

data/tutorials/en/000_up_and_running.md
shonfeder: My own feeling is that this command is much more user friendly, especially for people just getting started. Executing the program by giving a path to *a file that doesn't exist* is quite strange, in my experience, and iirc, I've seen new learners get thrown off by it. However, other tool chains that users are likely familiar with (e.g., npm, cargo, sbt, etc.) all have this convention of being able to run the executable under development by name.

I would ask too, why is running an executable by name (which is what anyone who uses terminal shell's is used to any how) more magical than running an executable by a non-existent path? :)

Note that this point is not altogether unrelated to correcting the command: previously, the generated file had the same name as the executable's `public_name` so there was somewhat less room for confusion.

What do you think?  Do you find my reasoning here compelling? Should we solicit input (perhaps from some people who are just learning OCaml)? Or are you firm on your position and I should just remove this part?

gasche: I think that your experience is at least as valid as mine (you know the dune UI better and possibly have had more recent interactions with dune beginners), and in addition you are the one who wrote the PR, so I would suggest to follow your gut feeling here instead of mine.

tmattio:
That looks very good, thank you so much @shonfeder for your work and @gasche for the thorough review 🙂

shonfeder:
Thanks all, for the reviews and the ongoing work to maintain these resources.

haesbaert:
Hi, I just had a newcomer hitting the same problem, I think this fix should be backported to the current ocaml.org website, as it's still showing the outdated instructions.

https://ocaml.org/learn/tutorials/up_and_running.html

SoumenDass:
I just started to learn Ocaml and got bit by this today - the tutorial instructions don't work. Then I tried the instructions listed above and they don't work either:

```
$ dune init project helloworld
Success: initialized project component named helloworld
$ cd helloworld/
helloworld $ dune build
helloworld $
helloworld $ dune exec ./bin/helloworld.exe
Error: Program "./bin/helloworld.exe" not found!

```

PS: this worked:
```
$ dune init project helloworld
Success: initialized project component named helloworld
$ cd helloworld/
helloworld $ dune build
helloworld $
helloworld $ dune exec helloworld
Hello, World!
```

hmackiernan:
I am probably adding minimal value by commenting but I just wanted to record the fact that I, too, just now fell into this same hole while trying to follow the tutorial as written.

shonfeder:
@hmackiernan Could you share the link for the problematic tutorial material you are referring to? Or perhaps which hole you fill into? (There are a couple in this thread.) Looking at https://ocaml.org/learn/tutorials/up_and_running.html, things seem to be in order afaict.

hmackiernan:
@shonfeder Sorry about that I didn't see that there were multiples. I think the one I tried to follow was this one:
https://v2.ocaml.org/learn/tutorials/up_and_running.html

cuihtlauac:
@hmackiernan Sorry you had a bad experience. As @shonfeder accurately pointed out, the document you read is obsolete. It should be marked as deprecated or no longer be online

shonfeder:
@hmackiernan no need to apologize! The problem here is definitely some stale docs that should be removed or marked as deprecated as @cuihtlauac noted. Thank you very much for taking the time to point out the problem. It is adding non-minimal value :)

PR #340, opened 2022-03-05
Improve homepage content

tmattio:
This improves the homepage copy as per @jsomers's suggestions.

@jsomers I didn't include the changes on the success page, as I thought it would be better done in the separate PR that will merge the success stories and industrial users pages. If this looks good to you, we can merge the changes on the homepage independently.


jsomers:
@tmattio Thanks for doing this! Lgtm.

PR #348, opened 2022-03-10
Closes #341, show and install the latest version

### LINKED ISSUE #341: tutorial is out of date with latest release.
In data/tutorials/en/000_up_and_running.md

it says to install 4.11.1: `opam switch create 4.11.1` which is not the current release.  Having to manually update this every time a new release comes out seems annoying.

Looking through the documentation, it looks like to see the latest version one would run:

`opam switch list-available ocaml-base-compiler`

and then they can run `opam switch create 4.13.1` or whatever the latest version is at the time the user reads it.

Am I correct this is the right set of commands, and is this a good change?  If so, I can try my hand at submitting a PR.

birlorg:
This fixed #341 and teaches the user how to show the latest version of the compiler and install it.
This will also help them upgrade in the future if they desire.

### DIFF
--- a/data/tutorials/en/000_up_and_running.md
+++ b/data/tutorials/en/000_up_and_running.md
@@ -42,8 +42,11 @@ Then, we install an OCaml compiler and some basic dev tools:
 opam init
 eval `opam env`
 
-# install given version of the compiler
-opam switch create 4.11.1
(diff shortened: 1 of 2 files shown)


tmattio:
Thanks @birlorg 🙂

PR #352, opened 2022-03-11
Rework documentation structure

tmattio:
This PR reworks the documentation to:

- Expose the documentation pages (aka tutorials) under `/docs/<document-id>`
- Include the title and page description in the markdown content, instead of using the metadata
- Update the documentation category to be "Getting Started", "Language", "Data Structures", "Tutorials", "Guides" and "Resources".

The structure update is inspired by https://diataxis.fr/ and aims at providing a structure that is easier to navigate for users. It also prepares the ground for the new documentation content we want to write (such as developer guides, more resources, language doc and tutorials)

- Getting Started is the documentation on how to install OCaml and get up and running. It is a How-to Guides section in the Diataxis framework.
- Language is the documentation to learn the language. It is an Explanation section in the Diataxis framework.
- Data Structure is the documentation of the data structures provided in the standard library. It is an Explanation section in the Diataxis framework.
- Tutorials maps to Tutorials section in the Diataxis framework.
- Guides aims at documenting the common developer workflows in OCaml. It now contains Compiling Projects, Debugging and Profiling from the existing documentation, and we want to add guides such as Bootstrapping a Project, Generating Documentation, etc. It maps to the How to Guides in the Diataxis framework.
- Resources lists general resources such as common errors messages. We want to add things like an index of the commonly used operators (e.g. `|>`, `@@`, `>>=`, `>>|`, etc.). It maps to the Reference section of the Diataxis framework.

cc @gasche as I saw your post on Discuss about the Diataxis framework (https://discuss.ocaml.org/t/diataxis-documentation-structure/7750). I'm curious to have your feedback on this new structure.

The full structure is as follow:

```
- GETTING STARTED
  - Up and Running with OCaml
  - A First Hour with OCaml
  - OCaml on Windows
- LANGUAGE
  - Data Types and Matching
  - Functional Programming
  - If Statements, Loops and Recursions
  - Modules
  - Labels
  - Pointers in OCaml
  - Functors
  - Objects
  - Error Handling
- DATA STRUCTURES
  - Lists
  - Map
  - Sets
  - Hash Tables
  - Streams
- TUTORIALS
  - Formatting and Wrapping Text
  - Command-line Arguments
  - File Manipulation
  - Calling C Libraries
  - Calling Fortran Libraries
- GUIDES
  - Compiling OCaml Projects
  - Debugging
  - Profiling
- RESOURCES
  - OCaml Programming Guidelines
  - Common Error Messages
  - Comparison of Standard Containers
  - Best Practices
  - OCaml Platform
```


gasche:
This generally looks good to me. I'm by no means a documentation expert or a Diataxis expert, and I am not familiar enough with the previous organization. But I think that it did lack structure, and that having clear general organization principles is excellent -- it is going to make it easier to structure things and give more coherent result.

One nitpick: shouldn't the "Comparison of Standard Containers" piece move to the Data Structures section? (Possibly it could become "Comparison of standard data structures", containers is now an overloaded word with Docker etc.)

I'll also note that "Data structures" stands out among the toplevel categories in your organization as something that is not, strictly speaking, "necessary". It's clear that the other categories could be around no matter what the rest of the content, but "Data Structures" feels a bit more arbitrary: we had tutorials about those contributed, enough to have a full section, but we could have more "Explanations" that are about something else and then the focus would change. I'm not suggesting any change here, just a remark.

christinerose:
My partner asked if there were other containers in OCaml like packages, map or hashmap, or a dictionary. Are there something that contains keys & values, or any other type of container other than lists and set that need to be under data structures?

I think the First Hour with OCaml is a good place to start, but the current document would take far more than an hour! Even though it's long, it's more lightweight than a tutorial, as it's more of an intro into how the language is used, but if people are coming here to learn OCaml, wouldn't tutorials come before Language and Data Structures? Perhaps an Introductory tutorial earlier before the series of tutorials?

Also, there are no OCaml tutorials anywhere on the Internet (that I could find) that go past installing OCaml and a little bit about how the language is used. I think we really need to create proper tutorials for OCaml (and perhaps for elsewhere) later this year, if the goal is to bring new people to OCaml. 

On that topic, I really believe there needs to be at least a short introduction accessible from the Learn navigation panel, even if it's a link to "Why OCaml." As someone new to OCaml, it's not clear why anyone would want to learn OCaml and use it instead of more common languages. Most programmers get an introduction to functional programming at university, so they have some contact, but it was always in an academic setting to solve academic problems. So first it must be clear why someone would want to use a functional language, and second why it would be OCaml over Haskell, the better known one.

tmattio:
> Are there something that contains keys & values

These would be hashtbl and map, which both have a doc page already.

> or any other type of container other than lists and set that need to be under data structures?

I guess we could add `Array`, `Stack`, `Queue` and `Seq` (the latest should replace `Streams` I think)

> Perhaps an Introductory tutorial earlier before the series of tutorials?

Is the First Hour with OCaml more of a tutorial, or an explanation in your opinion? We could move it to the Tutorials and extend it a little bit.

> On that topic, I really believe there needs to be at least a short introduction accessible from the Learn navigation panel, even if it's a link to "Why OCaml."

Right, let's continue this discussion in #349 🙂

christinerose:
> Is the First Hour with OCaml more of a tutorial, or an explanation in your opinion? We could move it to the Tutorials and extend it a little bit.
> 
I feels more like an explanation to me. 
A tutorial in my mind would have clear steps to follow, each leading to the next, while you build something, like perhaps start with 'Hello World' and then have a second and maybe third exercise that are a little more complicated, so the few truly only take an hour or less to complete, just as a little taster of OCaml and how it works. 

The other parts covered in "first hour" seem to be covered in more detail later, like Data Types, Expressions, Recursive Functions, etc. If feels that information would be better suited in that section rather than these short introductions. 

I think the first section (A First Hour with OCaml) and the second section (Running OCaml Programs) work in the "first hour" document. The explanation of Imperative OCaml (perhaps more brief ?) would fit well here, as those who are new to OCaml likely use imperative languages, and that would be followed by 'Hello World' and one or two others shorter, simple exercises.

The Standard Library and a Module from Opam could be in their own document, perhaps. 

At the end of this page, it says "this quick tour" gives a taste of OCaml, and that's a perfect way to describe what's on this page. Rather than a tutorial, it's a tour. To someone who doesn't know OCaml, it feels like a straight forward, introductory tutorial would be more beneficial, and then perhaps another document called "A Quick Tour of OCaml" or just "A Quick Tour" that introduces those other things. 

What do you think?

---
Another observation: under Language, the "Functional Programming" document seems like it would be a better fit under Getting Started as part of the *what is this and why should we use it* information, as it's partly an introduction. ...or if you decide to put in an intro to OCaml here, it could be part of that section.

tmattio:
Thanks a lot for the feedback @gasche!

> shouldn't the "Comparison of Standard Containers" piece move to the Data Structures section? (Possibly it could become "Comparison of standard data structures", containers is now an overloaded word with Docker etc.)

I've moved the file under Data Structures and renamed it to "Comparison of Standard Data Structures".

> It's clear that the other categories could be around no matter what the rest of the content, but "Data Structures" feels a bit more arbitrary

I agree! To me, these should be under the "Language" category. The problem is that it will be confusing to put them at the same level as the other documentation pages in Languages. We don't have a good way to present multiple levels of documentation (e.g. chapters in the language documentation).
@christinerose and @joanmh123 are currently working on something related though, so we might have a better way to organize the "Explanations" (aka the language documentation) soon.

tmattio:
> At the end of this page, it says "this quick tour" gives a taste of OCaml, and that's a perfect way to describe what's on this page. Rather than a tutorial, it's a tour. To someone who doesn't know OCaml, it feels like a straight forward, introductory tutorial would be more beneficial, and then perhaps another document called "A Quick Tour of OCaml" or just "A Quick Tour" that introduces those other things.

My main concern with this, and also with an introduction document that covers some aspects of "Why OCaml" is that there are multiple ways to "get started":

- You can land on Up and Running, then go to the quick tour? Or maybe the intro?
- You can land on the intro, then should you go on the Quick tour?
- You can land on that tutorial, then go in up and running?

I can imagine it becoming very confusing for people trying to get up and running quickly.
IMO, we should have a unique, very clear user path to get started with OCaml.

tmattio:
I'll merge now, to unblock #350, but of course, this is a first step, and the structure is still far from perfect. Further feedback is welcome, and we will land other PRs to discuss how we can continue to improve it.

gasche:
> My main concern with this, and also with an introduction document that covers some aspects of "Why OCaml" is that there are multiple ways to "get started":
>
> - You can land on Up and Running, then go to the quick tour? Or maybe the intro?
> - You can land on the intro, then should you go on the Quick tour?
> - You can land on that tutorial, then go in up and running?
>
> I can imagine it becoming very confusing for people trying to get up and running quickly.
IMO, we should have a unique, very clear user path to get started with OCaml.

I'm sure there are improvements to be done in this area, but we should also be careful that different use-cases have different needs. For example, if my question is "I already know how to use language ..., should I consider learning OCaml as well?", I may want to be exposed to the language itself, and not care about the tooling so much. But then if my question is "I've decided to write OCaml code, how do I actually compile and run my first program?", the focus is on the tooling rather than the language features. So I think that "Up and Running" and "A first hour with OCaml" both serve useful purposes, and I'm not sure that merging them would give a document that is as efficient for those purposes.

gasche:
One use-case would be: I'm a teacher at some university, I want to direct my students at an "official" webpage for how to install OCaml on their system. (I could write my own page but those quickly get out of date as the ecosystem evolves.)

gasche:
Another use-case (somewhat similar to the "teacher/student" scenario): I organize an OCaml-related hackaton at some free software events, and some people in the audience have never used OCaml before. What's the first link I give for how to setup a development environment on their machine?

Issue #286, opened 2022-02-08
Distinguish a package description from README on absence of the latter


hyphenrf:
Right now, package overview puts the `description` field text right under `README.md` header. A separation of those two fields seems to make sense, and (perhaps) a way to detect that both fields are the same when a package uses its `README` for the `description` field.

tmattio:
Can you give an example of that? We're not adding the description above the README AFAICT

hyphenrf:
sure, I noticed it first on this package: https://v3.ocaml.org/p/ppx_seq/0.2.0
the `description` field appears under the `README.md` header as it shows

hyphenrf:
Oh, I just checked the source code:

https://github.com/ocaml/v3.ocaml.org-server/blob/a8eb00767d46f983356d287d9b2ce55b21ec684c/src/ocamlorg_web/lib/handler.ml#L423-L427
and it seems descriptions are used as fallback when `README.md` file "doesn't exist".. So I guess the problem is just that. description being under the wrong header on fallback:

https://github.com/ocaml/v3.ocaml.org-server/blob/a8eb00767d46f983356d287d9b2ce55b21ec684c/src/ocamlorg_frontend/pages/package_overview.eml#L18-L20

hyphenrf:
Would you like me to try fixing this myself? ~~I'm thinking of adding a conditional to the template in order to render the correct title depending on whether I'm getting `Some` or `None` (so `Option.map` instead of `Option.value`)~~

hyphenrf:
scratch that, I do need to pass something either way.. so perhaps an extra polyvar/boolean param for title is the right way to approach this. Maybe a title string should be passed along with the discovered filename even. If I do it this way, it could allow you to work out the solution for (#30) with less friction.

What do you think?

tmattio:
I think replacing the description with "This package doesn't have a README" or something similar should be good enough?

hyphenrf:
yeah that sounds alright, I'm onto it

hyphenrf:
.... I don't have enough memory on my laptop to install the deps and build the website. I tried and it eventually crashes midway through and I have to clean up the switch broken package by broken package instead of nuking it and starting fresh.

Anyway, I wrote a patch and unfortunately I can't test it. Someone with a beefier machine can take it for a run and submit the PR 😅 

```patch
diff --git a/src/ocamlorg_frontend/pages/package_overview.eml b/src/ocamlorg_frontend/pages/package_overview.eml
[...]

PR #427, opened 2022-04-17
distinguish a package description from README

tatchi:
This PR applies https://github.com/ocaml/ocaml.org/issues/286#issuecomment-1034184814 as suggested by @hyphenrf

Example with the `base` package:

image

### DIFF
--- a/src/ocamlorg_frontend/ocamlorg_frontend.ml
+++ b/src/ocamlorg_frontend/ocamlorg_frontend.ml
@@ -26,10 +26,10 @@ let news_post news = News_post.render news
 let opportunities ?search ?location ~locations opportunities =
   Opportunities.render ?search ?location ~locations opportunities
 
-let package_overview ~readme ~dependencies ~rev_dependencies ~homepages ~source
-    package =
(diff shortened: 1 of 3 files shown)


tatchi:
What about showing the description on the right side instead? And keeping the center of the page for the readme (if it exists). If the readme doesn't exist, we can display a fallback message "No README file found" or something similar?

image

hyphenrf:
this seems like a great idea, I definitely prefer it to my own patch.

tmattio:
Thanks @tatchi!

> What about showing the description on the right side instead? And keeping the center of the page for the readme (if it exists). If the readme doesn't exist, we can display a fallback message "No README file found" or something similar?

I'm a bit wary that the length of the description will be too long in a lot of cases. (e.g. https://github.com/aantron/dream/blob/master/dream.opam#L6).

For now, the solution implemented in the PR already improves the situation and there doesn't seem to be any downside, so let's merge it, and feel free to open a new PR with the alternative solution where we can discuss the pro and cons, and find a solution for the long descriptions.

PR #251, opened 2022-01-25
Streamline opam package installs and REPL usage

yawaramin:
Install all opam dev tool packages in one shot, instead of piecemeal. This is
more efficient with the user's time.

Use utop directly instead of spending paragraphs showing the `ocaml` toplevel,
then talking about `rlwrap`. For beginners, `ocaml` toplevel and `rlwrap` are
almost certainly irrelevant when utop exists.

PR #442, opened 2022-04-29
Fix install and toplevel instructions

yawaramin:
Somehow the instructions in this document got mixed up with recent
changes from https://github.com/ocaml/ocaml.org/pull/251 . Fix up the
install and toplevel usage instructions.


tmattio:
Thanks @yawaramin 🙂

PR #448, opened 2022-05-01
Fix some 404s on ocaml.org

edwintorok:
There are more, these are just the ones that are found first by `linkchecker` and were obvious how to fix.


tmattio:
Thanks a lot @edwintorok!

PR #451, opened 2022-05-02
Fix syntax coloring for home page snippet

thomas-huet:
The code snippet on the home page has the syntax coloring wrong: `let rec fac x` is colored as if `rec` was the name of the function and `fac x` its arguments. I'm changing it to color `rec` as a keyword instead.


tmattio:
Thanks @thomas-huet 🙂

PR #452, opened 2022-05-04
Add working example of mutual recursion

grawlixes:
At the very bottom of [this doc](https://ocaml.org/docs/if-statements-and-loops), it's explained to the reader that it's possible to write mutually recursive functions in OCaml. 

The doc seems to want to show the reader a working example of mutual recursion, but instead it shows a repeat of a failed example from earlier in the doc. This PR will update that example to one that actually does work, so the reader understands how to do it themselves.


tmattio:
Thanks @grawlixes!

PR #463, opened 2022-05-15
Fixed formatting in OCaml programming guidelines

axrwl:
In the page [https://ocaml.org/docs/guidelines](https://ocaml.org/docs/guidelines), some parts of the text were unintentionally improperly formatted.

### DIFF
--- a/data/tutorials/rs_00_guidelines.md
+++ b/data/tutorials/rs_00_guidelines.md
@@ -134,11 +134,11 @@ In conclusion, copy-pasting code leads to programs that are more
 difficult to read and more difficult to maintain: it has to be banished.
 
 ###  How to comment programs
-####  Don't hesitate to comment when there's a difficulty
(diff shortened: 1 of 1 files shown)


tmattio:
Thanks for the PR @kodwx!

Unfortunately, I don't think the PR is correct, the use of heading 4 for the items of the "How to comment programs" is intentional and consistent with the rest of the document.
I agree this section could be improved, though: we can add some content under the headings; we can merge the first and second items, etc.

I'll close this PR for now, but feel free to open a new one if you'd like to suggest improvements to the `How to comment programs` section.

axrwl:
The current website renders like this @tmattio 
<center>Screen Shot 2022-05-24 at 7 19 54 PM</center>

tmattio:
Which I concur is not great.. To fix it, I think we should:

- Update the CSS for the `<h4>` tags to make them smaller than `<h3>`
- Probably add some content in each section

PR #466, opened 2022-05-24
Fix pinned code-mirror dependency

JiaeK:
Current pinned `code-mirror` dependency for up and running is pinned to a branch which gives an environment setting error.
This PR fixes the dependency to the appropriate commit hash.

### DIFF
--- a/ocamlorg.opam
+++ b/ocamlorg.opam
@@ -87,7 +87,7 @@ dev-repo: "git+https://github.com/ocaml/ocaml.org.git"
 pin-depends: [
   ["esbuild.dev" "git+https://github.com/tmattio/opam-esbuild#main"]
   ["tailwindcss.dev" "git+https://github.com/tmattio/opam-tailwindcss#main"]
-  ["code-mirror.dev" "git+https://github.com/patricoferris/jsoo-code-mirror#static"]
+  ["code-mirror.dev" "git+https://github.com/patricoferris/jsoo-code-mirror#3e47f324456aafd84dece3dec4e7be70c2696f11"]
   ["js_top_worker-rpc.dev" "git+https://github.com/tmattio/js_top_worker#614c4b45b9a6924496f5569650ebb0676dba1526"]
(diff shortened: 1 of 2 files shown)


tmattio:
Thanks for the PR @JiaeK!

I think the pin is correct though, you can see the branch here: https://github.com/patricoferris/jsoo-code-mirror/tree/static

patricoferris:
Ah this was my suggestion. We were seeing `git fetch` failing for the branch name version so I suggested pinning to a specific commit hash (the head of the `static` branch instead). But perhaps this was just a one off issue ?

EDIT: I was also misremembering that the "branch name in opam pin" issue I was thinking about was actually nothing to do with opam, but instead ocaml-ci https://github.com/ocurrent/ocaml-ci/issues/384

tmattio:
Ah! Thanks for the details, did you have any issue with the other pins that use the branch names then? e.g. `git+https://github.com/tmattio/opam-tailwindcss#main`

JiaeK:
No, it was only `code-mirror` one. So, to give a little more context, this was my error message


code-mirror error

tmattio:
That's strange, the issue above indicates that using the `main` branch doesn't work either for ocaml-ci. Perhaps you were only seeing the first pin that failed?

I can merge the PR as-is as it doesn't have an impact on ocaml.org, but do you want to confirm that you don't have any other issue using this PR before?

JiaeK:
Yeah it is strange. And it's possible that I was the only one that's having the issue 😅 

The only thing I can say is: after that, I tried to install `code-mirror` alone, I got pretty much the same error message. So I asked @patricoferris and got a suggestion that I should run `opam pin -y git+https://github.com/patricoferris/jsoo-code-mirror#3e47f324456aafd84dece3dec4e7be70c2696f11` and then everything worked.

tmattio:
Alright, if that fixes your issue, let's merge as-is. Thanks a lot for the PR @JiaeK and apologizes for the delays.

PR #470, opened 2022-05-31
use 4.14.0 as latest ocaml version

tcoopman:
To get up and running it's confusion if an older version is shown. I've also added an instruction to show all possible switches but as that returns a huge list, I'm wondering if that might be too much for the up and running. On the other hand, it would be good for new people to find a list of possible releases they can use.

see for example this question: https://discuss.ocaml.org/t/are-ocaml-install-instructions-up-to-date/9941

### DIFF
--- a/data/tutorials/gs_00_up_and_running.md
+++ b/data/tutorials/gs_00_up_and_running.md
@@ -47,7 +47,7 @@ $ opam init
 $ eval `opam env`
 
 # install a specific version of the OCaml base compiler
-$ opam switch create 4.11.1
+$ opam switch create 4.14.1
 $ eval `opam env`
 
 # install dev tools, hit Enter to confirm at Y/n prompt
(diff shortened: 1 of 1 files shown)


Octachron:
The latest version of OCaml is 4.14.0 .

tcoopman:
Maybe the instructions should point to https://github.com/ocaml/ocaml/releases?

tmattio:
Thanks for the PR @tcoopman and for the review @Octachron!

This is making the doc better, so let's merge it now, but I agree it would be nice to have a solution so that the documentation doesn't get outdated after every OCaml release. A PR that updates the doc to point to the release page would be welcome 🙂

Octachron:
A link would always be a little less ergonomic that a plain text that can be read directly. Maybe the best solution would be to allow templating the last version of OCaml in `md` files too?

PR #471, opened 2022-05-31
Update lg_08_error_handling.md

jonahbeckford:
Switch from `lazy_t` to `Lazy.t`

Guidance at https://v2.ocaml.org/api/Lazy.html says:

> Note: lazy_t is the built-in type constructor used by the compiler for the lazy keyword. You should not use it directly. Always use Lazy.t instead.


tmattio:
Thanks a lot @jonahbeckford!

PR #476, opened 2022-06-03
Typo in Modified run-length encoding

fccm:


tmattio:
This doesn't seem to be a typo, this is the type of ` encode ["a"; "a"; "a"; "a"; "b"; "c"; "c"; "a"; "a"; "d"; "e"; "e"; "e"; "e"];;`. The `rle` type is defined above.

fccm:
Yes the type `rle` is defined above.
The typo is `rle/2`. It should be just `rle`.
Please read the diff/patch carefully.

tmattio:
The `rle/2` type is correct, it's actually generated by `mdx`. To remove it, we should avoid defining it twice. It's already defined in the solution above the statement, so we can add `<!-- $MDX skip -->` to the code block that redefines it.

fccm:
I'm not able to reproduce this type `string rle/2 list` in my toplevel.
My toplevel just give me `string rle list`.
What is this type exactly ?

tmattio:
I can reproduce this with OCaml 4.14's toplevel:

```
OCaml version 4.14.0
Enter #help;; for help.

# type 'a rle =
  | One of 'a
  | Many of int * 'a;;    
type 'a rle = One of 'a | Many of int * 'a
# let encode l =
    let create_tuple cnt elem =
          if cnt = 1 then One elem
      else Many (cnt,   elem) in
    let rec aux count acc = function
          | [] -> []
      | [x] -> (create_tuple (count +   1) x) :: acc
      | hd :: (snd :: _ as tl) ->
              if hd = snd then aux (count + 1) acc tl
            else aux 0 ((create_tuple (count + 1) hd) :: acc) tl in
      List.rev (aux 0 [] l);;  
val encode : 'a list -> 'a rle list = <fun>
# type 'a rle =
  | One of 'a
  | Many of int * 'a    ;;
type 'a rle = One of 'a | Many of int * 'a
# encode ["a"; "a"; "a"; "a"; "b"; "c"; "c"; "a"; "a"; "d"; "e"; "e"; "e"; "e"];;  
- : string rle/2 list =
[Many (4, "a"); One "b"; Many (2, "c"); Many (2, "a"); One "d";
 Many (4, "e")]
```

Octachron:
The issue is that there is two types named `rle` in the REPL session. And encoding works on the first `rle` type which is no longer in scope after the definition of the second one. A simpler illustration of the issue might be
```ocaml
type t = A;;
type t = B
let y = A;;
```
which prints `y` as:
```ocaml
val y : t/2 = A
```

fccm:
Ok, this is something new then.
I'm still using 4.13.1

Octachron:
This is not a new feature, this disambiguation of printed types is here since OCaml 4.08 .

fccm:
Yes, I probably tested the code in an old switch, or maybe in a toplevel session where I only defined only one `rle` type.
I was not aware of this new feature, sorry.

fccm:
This type is defined 2 times in exactly the same way.
I think the second is there for clarity.
Does it make sense to leave `rle/2` ?

fccm:
The ocaml provided in a recent chrome-book is still a 4.05

tmattio:
> Does it make sense to leave rle/2 ?

Nope, we can avoid redefining the type twice by adding `<!-- $MDX skip -->` above the code block that defines the type for the second time 🙂

PR #477, opened 2022-06-03
typo in nonograms.md

fccm:


tmattio:
This is not a typo, the inclusion of `;;` at the end of OCaml phrases is recommended by `mdx`.

fccm:
Yes this is a typo because the function expects 2 arguments, and there is a `;;` after each arguments.
The second argument is given on the second line, and you can see that the `;;` is there.
If you put an additional `;;` in the middle of the argements, the function application will be partial, and the second argument that is after the first `;;` will not be passed to the function.
I tested in the toplevel to be sure.
(it's true that it's not that easy to read)

tmattio:
You're right, apologies!

We should be checking this with `mdx`, but I guess we skipped it because we didn't provide a solution for the `solve` function. Do you happen to have the solution implemented somewhere? (reading between the line that you may as you've tested in the top-level)

PR #490, opened 2022-06-12
Correct small spelling mistake on https://ocaml.org/problems#3

joshniec:
While working through the problems on https://ocaml.org/problems to learn OCaml I noticed an ever so small spelling mistake, thought I would do my part in correcting it rather than letting it sit (I know, I know, it's extremely nit picky) 

Also, when attempting to write a solution for problem # 3, it references the implementation of `List.nth` as the problem statement, which uses zero based numbering, however, the solution uses one based numbering. To align with the problem statement, should the solution be updated to:

```ocaml
# let rec at k = function
    | [] -> None
    | h :: t -> if k = 0 then Some h else at (k - 1) t;;

(* passes using zero based numbering *)
# let%test "zero based numbering" = at 2 [ 6; 5; 4; 3; 2; 1; 0 ] = Some ( List.nth [ 6; 5; 4; 3; 2; 1; 0 ] 2 ) (* 4 *)
```

(FWIW my solution was much uglier but something I noticed when writing tests for my solution and against the proposed solution)

### DIFF
--- a/data/problems/003_nth_element.md
+++ b/data/problems/003_nth_element.md
@@ -10,13 +10,13 @@ tags: [ "list" ]
 ```ocaml
 # let rec at k = function
     | [] -> None
-    | h :: t -> if k = 1 then Some h else at (k - 1) t;;
+    | h :: t -> if k = 0 then Some h else at (k - 1) t;;
 val at : int -> 'a list -> 'a option = <fun>
 ```
 
 # Statement
 
-Find the K'th element of a list.
+Find the N'th element of a list.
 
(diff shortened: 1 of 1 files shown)


tmattio:
Thanks a lot @joshniec 🙂 

The solution didn't have to follow the behavior of `List.nth`, but I agree it's better to follow the same conventions as the standard library. I pushed your suggestion with a small fix in the exercise statement as well.

joshniec:
Thanks @tmattio!

PR #493, opened 2022-06-14
Change CPU usage color

JiaeK:
This PR fixes the color of the CPU usage bar blue to green, as it was planned. 
Before the fix, the color was just like the memory usage bar (blue). And after it would look like the below screenshot. 

cpu color after


tmattio:
Thank you @JiaeK 🙂

Issue #486, opened 2022-06-10
Package redirect not fully functional


Leonidas-from-XIV:
If I visit [ocaml.org/p/yojson](https://ocaml.org/p/yojson) I get redirected to the last release, which I think is a great feature to ensure stable URLs, I really enjoy this.

Unfortunately though if I visit [ocaml.org/p/yojson/](https://ocaml.org/p/yojson/) I get.. a page. Not a 404 page, not a redirect, just nothing.

I know that the URLs are not equivalent but it would be nice if the version with the slash worked too.

koonwen:
Hello I'm a new contributor!

I took a look into this issue and it seems like there is already middleware that was set up to drop the trailing slash.  https://github.com/ocaml/ocaml.org/blob/8808e5c348efb4c49d67ba7771f490b360bb36b2/src/ocamlorg_web/lib/router.ml#L60-L70 
However, after closer inspection of how Dream works, it seems that middleware is **only** run if there is a route in the scope that matches the request exactly. This means that `Middleware.no_trailing_slash` is never actually run for its intended purpose because none of the routes are registered with a trailing slash. A simple way to make this redirect behavior work is just to register routes with it's corresponding "**/" pair.

For example
```ocaml
[
Dream.get (Url.package ":name") (Handler.package t); 
Dream.get (Url.package ":name/") (Handler.package t)
]
```

But this solution is probably not ideal because the codebase will be littered a lot of extra routes with dummy handlers.

Other than this, I also tried to put the redirect logic into a handler which acts as a catch-all case when none of the routes match. The problem with this is that the wildcard `Dream.get "/**" ...` is already used for getting static files. To work around this, I can suggest that for the scopes that have `Middleware.no_trailing_slash` registered (graphql_route/page_routes/package_route), we can

1. Add prefixes to these scopes
2. Register the catch-all route handler in these scopes

What do you think about this proposal @tmattio ?

tmattio:
Fixed in #514

PR #514, opened 2022-07-04
Fix trailing slash redirection

tmattio:
The middleware was only applied when a match was found for the route in the defined scope, which defeats the purpose of handling unknown routes ending with a slash to the correct routes.

This commit makes the trailing slash middleware global to every request.


tmattio:
@koonwen thanks a lot for investigating the issue in #486, do you want to have a look at this PR?

koonwen:
Sure! Let me have a look through

koonwen:
This solution is great! I didn't realize you could just install the Middleware to run globally. Thank you for sharing it! I'll snoop around and see if there are more issues that I can help out with!

PR #518, opened 2022-07-05
compilation time graph fix

### LINKED ISSUE #407: compilation time graph on the landing page is useless
the plot of compilation times https://v3.ocaml.org/img/home/workfaster.svg is lacking units on the y axis. that should be added as well as a rough spec of the processor(s) used to get the timing. this issue has been pointed out before in the discuss thread but has not been fixed. in its current form the plot feels like dishonest advertisement.

MirzaBabarBaig:
Fixes #407


tmattio:
Thanks @MirzaBabarBaig!

Issue #453, opened 2022-05-04
RSS feed for OCaml Blog


mndrix:
The [OCaml Blog](https://ocaml.org/blog) has some excellent content, however it appears not to publish an RSS feed of its own articles.  I searched the page for "rss" and "atom". I also examined the HTML looking for `<link rel="alternate" ...>`. If the blog has its own RSS feed, I couldn't find it.

Having an RSS or Atom feed that lists recent OCaml Blog articles, would make it easier for those using feed readers to follow the OCaml Blog.

tmattio:
There's no RSS feed for the blog at the moment, but it could be implemented relatively easily I think. We could take example from Mirage's website which has a similar tech stack and provides an RSS feed: https://github.com/mirage/mirage-www/blob/master/lib/mirageio.ml#L73

koonwen:
Hi! I'd like to work on this if nobody has started on it!

tmattio:
Thanks @koonwen! I think @MirzaBabarBaig started working on it?

tmattio:
@koonwen you can take over the issue if you want 🙂

koonwen:
Okay thanks! I'll start working on it.

tmattio:
Done in #524 and live at https://ocaml.org/feed.xml 🎉

brabalan:
Thank you! I confirm it works with my tiny tool to populate the OCaml Weekly News.

PR #524, opened 2022-07-07
Add rss

koonwen:
This PR addresses #453 

I didn't quite follow the mirage example because I couldn't get their RSS feed working. 

Instead, I added some logic inside the scraping function to also generate an atom.xml file that is served directly from the filesystem.

**On a side note, I seem to have a dependency issue when I build the project, wondering if anybody else is facing this and has a fix. The conflict is between the `ocaml-lsp-server` and `hilite` package. I can only have one or the other in the switch at a time. This is quite troublesome because if I run `ocaml-lsp-server` during development, I will have to recompile the switch everytime I want to perform a build to install `hilite`
```bash
❯ opam install ocaml-lsp-server
The following actions will be performed:
  ⊘ remove    hilite           0.1.0                 [conflicts with omd]
  ↘ downgrade omd              2.0.0~alpha2 to 1.3.1 [required by ocaml-lsp-server]
  ∗ install   ocaml-lsp-server 1.12.1
```


src/ocamlorg_frontend/url.ml
```diff
@@ -35,6 +35,7 @@ let releases = "/releases"
 let release v = "/releases/" ^ v
 let workshop v = "/workshops/" ^ v
 let blog = "/blog"
+let atom = "/atom"
```
tmattio: Perhaps a more standard URL for RSS feed is `/feed.xml`. But also if you generate the file in `asset`, you don't need to add a route for it. 
```suggestion
```

tool/ood-gen/lib/rss.ml
```diff
@@ -31,6 +31,12 @@ let feeds () =
 
 let scrape () =
   let feeds = feeds () in
+  let id = Uri.of_string "https://ocaml.org/atom" in
+  let title : Syndic.Atom.title = Text "OCaml.org blog" in
+  let updated = Ptime.of_float_s (Unix.gettimeofday ()) |> Option.get in
+  let entries = feeds |> River.posts |> River.create_atom_entries in
+  let feed = Syndic.Atom.feed ~id ~title ~updated entries in
+  Syndic.Atom.write feed "data/rss/atom.xml";
```
tmattio: You can generate the file in `asset/feed.xml`, and the file will be available directly from https://ocaml.org/feed.xml
```suggestion
Syndic.Atom.write feed "asset/feed.xml";
```

src/ocamlorg_web/lib/router.ml
```diff
@@ -40,6 +40,7 @@ let page_routes =
       Dream.get (Url.release ":id") Handler.release;
       Dream.get (Url.workshop ":id") Handler.workshop;
       Dream.get Url.blog Handler.blog;
+      Dream.get Url.atom Handler.atom;
```
tmattio: You don't need a new route if you generate the file in the asset directory.
```suggestion
```

src/ocamlorg_web/lib/handler.ml
```diff
@@ -179,6 +179,9 @@ let blog req =
     (Ocamlorg_frontend.blog ~featured ~rss:current_items ~rss_page:page
        ~rss_pages_number:number_of_pages ~news)
 
+let atom =
+  Dream.from_filesystem "data/rss/" "atom.xml"
+
```
tmattio: We can't depend on the filesystem in production, instead we [crunch](https://github.com/mirage/ocaml-crunch) the static file into the server binary.
```suggestion
```

tmattio (commented):
Thanks a lot for getting the ball rolling on this @koonwen!
The approach looks great, the only thing I can see is that we don't need all the machinery to serve the file if you generate it in `asset/` directly. It would then be crunched at compile time and available as a static asset at https://ocaml.org/feed.xml

tmattio:
> This is quite troublesome because if I run ocaml-lsp-server during development, I will have to recompile the switch everytime I want to perform a build to install hilite

You can try using the (very experimental and unreleased) OCaml Platform installer: https://github.com/tarides/ocaml-platform-installer. Don't hesitate to leave feedback on the repo if you use it 🙂

koonwen:
Thanks for the pointers @tmattio, they are super helpful! I've made the changes you've suggested so that we are serving the /feed.xml directly from assets/. Let me know if there's anything I missed out!

src/ocamlorg_web/lib/handler.ml
koonwen: I'm not sure I fully understand this, why is the filesystem less reliable than a server binary?

tmattio (approved):
Looks great! Thanks again @koonwen 🎉 

I'll merge in staging to test it on the live environment and if everything works, I'll push to main.

One small note: linting the feed.xml file with https://validator.w3.org/feed/#validate_by_input, it looks like it's not entirely valid. It's not blocking the PR I think since RSS readers should still be able to parse it, but it would be a nice follow-up work to make it valid if you're interested 🙂

koonwen:
> One small note: linting the feed.xml file with https://validator.w3.org/feed/#validate_by_input, it looks like it's not entirely valid. It's not blocking the PR I think since RSS readers should still be able to parse it, but it would be a nice follow-up work to make it valid if you're interested 🙂

Sure thing! I'll have another look at it!

kuchta:
Thank you @koonwen ❤️. Unfortunatelly it's not linked on the blog (well even on main) page so RSS readers won't find it.

koonwen:
Hi @kuchta, thanks for your comment! From my quick search, it seems like not pointing out the RSS link on sites is becoming common. https://zapier.com/blog/how-to-find-rss-feed-url/. I suppose this is the direction the ocaml.org is going for as well. In my experience, my RSS reader managed to find the feed just by providing the ocaml.org url. Maybe some readers haven't implemented the search heuristic to find hidden rss feeds yet. In any case, it could be useful to create a new issue and suggest this as a new feature!

cuihtlauac:
Hi @koonwen, thanks for your feedback. Do you think https://github.com/ocaml/ocaml.org/pull/719 will help solve https://github.com/ocaml/ocaml.org/issues/718 ? I must say I don't use RSS.

koonwen:
@cuihtlauac yup looks good to me! Thanks for the initiative!

PR #526, opened 2022-07-08
Add validation pass over entries

koonwen:
@tmattio This is a follow up from the previous RSS PR #524

As suggested, I have added some validation over the feed entries so that it passes https://validator.w3.org/feed/#validate_by_input. 

The validator still recommends some changes in the entry content (dangerous HTML tags) but I'm not sure we want to change this in case it affects the structure of the blog post. Do let me know if we need to clean the content as well!


tmattio:
Brilliant! Thank you @koonwen!

I wanted to suggest that a better place for this could be in [Syndic](https://github.com/Cumulus/Syndic) itself, as it seems like something you'll want to do for every Atom feeds. But seeing that the last commit on Syndic was 3 years ago, let's merge this here for now. You can still open a PR on Syndic if you want and when it gets merged and released, we can remove this validation on OCaml.org's side.

Issue #229, opened 2022-01-14
for opam releases, expose changelog and release date


avsm:
This makes it convenient to browse details of a given release.

Via @hannesm in https://twitter.com/h4nnes/status/1482019842030981121

hannesm:
what I meant is -- apart from OCaml releases -- this information would be nice to have for individual opam packages :)

cuihtlauac:
I believe these PRs:
* https://github.com/ocaml/ocaml.org/pull/530
* https://github.com/ocaml/ocaml.org/pull/564

have fixed this. Let's close it.

PR #530, opened 2022-07-14
Add publication dates to opam packages

### LINKED ISSUE #40: Add package publication date to overview, search results and package statistics

tmattio:
This PR adds support for displaying publication dates in OCaml packages.

The challenge is that opam doesn't provide any metadata for the publication dates. The publication date of a package is then the date of the first commit of the opam file of the package.

To get this information, we can run `git --no-pager log --format="%at" --reverse -- <filename>| head -1` on the file. However, this takes time, running it on the 26K files in the opam-repository takes more or less 3 hours on my computer. Running that operation at startup time like we do for other operations on the opam-repository is not an option.

Instead, this PR proposes to add a data file to the repository that contains the result of
```
find packages -type f | grep "/opam" | sort | while read filename; do
  echo "$filename $(git --no-pager log --format="%at" --reverse -- $filename | head -1)"
done
```

The file looks like
```
packages/0install-gtk/0install-gtk.2.15.2/opam 1585479891
packages/0install-gtk/0install-gtk.2.16/opam 1585854647
packages/0install-gtk/0install-gtk.2.17/opam 1588602382
packages/0install-solver/0install-solver.2.17/opam 1588602382
```

So each line is an opam package with the file path and the timestamp separated by a whitespace.

We can then parse this file and generate an OCaml module at build time that will contain the list of `(filepath, timestamp)`.

As part of this PR, we add a CI that will regenerate this file on a weekly basis.

Fixes https://github.com/ocaml/ocaml.org/issues/40
Related to https://github.com/ocaml/ocaml.org/issues/229

**TODO**

- [x] Generate the complete file
- [x] Add the publication date in the package metadata in the frontend
- [x] Add the information to the GraphQL API


tmattio:
@koonwen I see that you've been quite active on the repository recently. Are you interested in reviewing this PR? 🙂

koonwen:
Sure thing! I'll have a look over the weekend 👍🏼

dra27:
Even on my Windows machine, this command is quite fast (15-25 seconds for the entire history). It's worth ensuring `diff.renamelimit` is at least 20000 so that it can cope with the OPAM 1.2.2 -> 2.0 migration in the history:

```
git log --diff-filter A --name-status --format="%ct" --first-parent --reverse
```

I then cheekily processed this output with awk:

```awk
/^[0-9]+/ {
  stamp = $0;
}

/^A\t/ {
  printf("%s %s\n", stamp, substr($0, 3));
}
```

but that should obviously be done in OCaml. A few notes:
- I'd suggest using `%ct` rather than `%at` - publication is when the file was committed, not authored. That said, the commit's commit time is still inaccurate - what's really wanted is the commit stamp of the _merge commit_
- There are duplicates (e.g. if a release gets reverted while revdeps are fixed - it's happened for alcotest and a few others IIRC). It's probably better to have the most recent publication date, hence using `--reverse` - the idea is that a fold down the lines will then give a map with the most recent publication date
- On my machine, adding `-- packages/*/*/opam` to the `git log` considerably slows the command down. Omitting it means that information is returned for invalid files, so that scanning of the lines should filter for that regex
- The `--first-parent` (used by `opam2web`) causes the commit time to be shown for the _merge commit_ rather than the author's commit (e.g. picked at random, `ocurl.0.7.0` which would show as 13 June 2017 for the author but the PR itself was merged on 14 June 2017)

tool/ood-gen/lib/opam_repository_timestamps.ml
```diff
@@ -0,0 +1,19 @@
+let all () =
+  let content = Data.read "opam-repository-timestamps.txt" |> Option.get in
+  let lines = String.split_on_char '\n' (String.trim content) in
```
dra27: If the file is altered to have the stamp first, then this can be done much more simply with `String.index_opt content ' '` (shouldn't trim either - we may not care, but the filename is perfectly permitted to end in spaces). That hardens against any possibility of a file with spaces (which can then be ignored by a later) vs rejecting all the valid stamps

src/ocamlorg_package/lib/info.ml
```diff
@@ -137,9 +138,22 @@ let mk_revdeps pkg pkgs rdepends =
     []
   |> Lwt.map List.rev
 
+module String_map = Map.Make (String)
+
+let get_publication package =
+  let name = OpamPackage.name_to_string package in
+  let name_with_version = OpamPackage.to_string package in
+  let path = Printf.sprintf "packages/%s/%s/opam" name name_with_version in
```
dra27: `name_with_version` is itself a key - it'd be simpler if map were set-up so that `name.version` was already the key rather than having to put `packages/name/name.version/opam` back in

dra27 (commented):
A couple of drive-by comments, but very much looking forward to retiring this part of opam.ocaml.org :)

tmattio:
Thanks a lot for the review @dra27! I've updated the PR with your comments and this allows to move the population of timestamps at startup time instead of having a CI that generates a file: much more elegant!

src/ocamlorg_package/lib/opam_repository.mli
```diff
@@ -30,3 +30,6 @@ val new_files_since : a:string -> b:string -> (Fpath.t * string) list Lwt.t
 (** Files created during a range of commit. Associate the date at which each
     file as been added, in Git's relative format. Returns more recent files
     first. *)
+
+val get_package_to_timestamp : unit -> float OpamPackage.Map.t Lwt.t
+(** Get the timestamp of the first commit of the given file path. *)
```
koonwen:
```suggestion
(** Creates a map of package to timestamp, where the timestamp is the first commit of the package's opam file. *)
```
documentation

src/ocamlorg_package/lib/opam_repository.mli
```diff
@@ -30,3 +30,6 @@ val new_files_since : a:string -> b:string -> (Fpath.t * string) list Lwt.t
 (** Files created during a range of commit. Associate the date at which each
     file as been added, in Git's relative format. Returns more recent files
     first. *)
+
+val get_package_to_timestamp : unit -> float OpamPackage.Map.t Lwt.t
```
koonwen:
```suggestion
val create_package_to_timestamp : unit -> float OpamPackage.Map.t Lwt.t
```
Nitpicking here. I think this naming is less ambiguous

src/ocamlorg_package/lib/opam_repository.ml
```diff
@@ -133,3 +133,46 @@ let new_files_since ~a ~b =
...
+        let found =
+          match OpamPackage.of_string_opt pkg with
+          | Some pkg -> OpamPackage.Map.add pkg time found
+          | None -> found
+        in
+        read_file found time rest
+  in
+  read_time OpamPackage.Map.empty
+
+let get_package_to_timestamp () =
```
koonwen:
```suggestion
let create_package_to_timestamp () =
```
naming

src/ocamlorg_package/lib/info.ml
```diff
@@ -196,14 +199,16 @@ let of_opamfiles
   let dependencies = get_dependency_set packages opams in
   Logs.info (fun f -> f "Reverse dependencies...");
   let* rev_deps = rev_depends dependencies in
+  Logs.info (fun f -> f "Publication dates...");
+  let* timestamps = Opam_repository.get_package_to_timestamp () in
```
koonwen:
```suggestion
let* timestamps = Opam_repository.create_package_to_timestamp () in
```
naming

koonwen (commented):
Looks good to me! Just a nitpick on one of the function name.

Regarding the CI failure, I ran into this problem too and it has to do with the latest `timedesc` package release having API changes. If you constrain `(timedesc
   (= 0.6.0))` like in https://github.com/ocaml/ocaml.org/commit/a6571ad5ae66df682972b3e297db673dbb8dd530, the build should pass.

dra27:
> Regarding the CI failure, I ran into this problem too and it has to do with the latest `timedesc` package release having API changes. If you constrain `(timedesc (= 0.6.0))` like in [a6571ad](https://github.com/ocaml/ocaml.org/commit/a6571ad5ae66df682972b3e297db673dbb8dd530), the build should pass.

I haven’t looked, but what prevents updating the code to cope with the new API and using the superior `(timedesc (>= 0.8.0))`?

koonwen:
> > Regarding the CI failure, I ran into this problem too and it has to do with the latest `timedesc` package release having API changes. If you constrain `(timedesc (= 0.6.0))` like in [a6571ad](https://github.com/ocaml/ocaml.org/commit/a6571ad5ae66df682972b3e297db673dbb8dd530), the build should pass.
> 
> I haven’t looked, but what prevents updating the code to cope with the new API and using the superior `(timedesc (>= 0.8.0))`?

The return value of one of the `Timedesc.to_iso8601` is just a string in (0.8.0) instead of a string option in (0.6.0)

Easily enough, in `src/dream_dashboard/template/overview_template.eml`, altering
```ocaml
let human_date_of_timestamp ts = 
  let span = Timedesc.Span.make ~s:(Int64.of_float ts) () in
  let td = Option.get (Timedesc.of_timestamp ~tz_of_date_time:Timedesc.Time_zone.utc span) in
  Option.get (Timedesc.to_iso8601 td)
```
to
```ocaml
let human_date_of_timestamp ts = 
  let span = Timedesc.Span.make ~s:(Int64.of_float ts) () in
  let td = Option.get (Timedesc.of_timestamp ~tz_of_date_time:Timedesc.Time_zone.utc span) in
  Timedesc.to_iso8601 td
```
seems to work as well, so this might be a better option

dra27:
That seems better, yes - it doesn't look like the option-returning version is coming back (https://github.com/daypack-dev/timere/pull/53)

tmattio:
I applied the suggestions - thanks a lot for the review @koonwen and @dra27! 😁

PR #536, opened 2022-07-29
Add merlin-js in playground

voodoos:
This enable the use of `merlin-js` in the playground.

This requires a secondary worker: `merlin_worker`. This worker weights 1mb when gzipped.

Currently supported features of Merlin:
- Type on hover
- Errors highlighting
- Completion, including Stdlib modules.

https://user-images.githubusercontent.com/5031221/181747484-f9b1513c-6e5c-499d-a58a-084e51c83807.mov

@patricoferris: since Codemirror released the first stable V6 I thought it was a good idea to upgrade. I switched the pinned repository to my fork but this can be switched back when the upgrade will be upstreamed :-)

### DIFF
--- a/dune
+++ b/dune
@@ -34,10 +34,10 @@
   (deps
    (:js %{workspace_root}/src/ocamlorg_playground/worker.js))
   (action
-   (run jsoo_minify %{js} -o worker.js)))
+   (run jsoo_minify %{js} -o %{targets})))
  (rule
-  (targets merlin_worker.bc.js)
+  (targets merlin.js)
   (deps
    (:js %{workspace_root}/src/ocamlorg_playground/merlin_worker.bc.js))
   (action
-   (run jsoo_minify %{js} -o merlin_worker.bc.js))))
+   (run jsoo_minify %{js} -o %{targets}))))
--- a/dune
+++ b/dune
@@ -34,4 +34,10 @@
   (deps
    (:js %{workspace_root}/src/ocamlorg_playground/worker.js))
   (action
-   (run jsoo_minify %{js} -o worker.js))))
+   (run jsoo_minify %{js} -o worker.js)))
+ (rule
+  (targets merlin_worker.bc.js)
+  (deps
+   (:js %{workspace_root}/src/ocamlorg_playground/merlin_worker.bc.js))
+  (action
+   (run jsoo_minify %{js} -o merlin_worker.bc.js))))
--- a/dune-project
+++ b/dune-project
@@ -60,6 +60,7 @@
   js_top_worker
   js_top_worker-client
   code-mirror
+  merlin-js
   (js_of_ocaml
    (>= 4.0))
   tailwindcss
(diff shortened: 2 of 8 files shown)


voodoos:
@tmattio could you enable the CI ?

tmattio:
Thanks a lot @voodoos, this is a huge improvement on the playground! 🎉 

I tested it and it works really well! 🙂

One suggestion I have would be to have a "Problems" panel alongside the "Output" one to list all the errors detected by Merlin. That can come in another PR though.

> This requires a secondary worker: merlin_worker. This worker weights 1mb when gzipped.

I'm slightly concerned about the asset size. I don't think it will be a blocker since we're already sending a 3mb file... but still it would be nice to do some damage control.

- Did you have any idea on how we could decrease the bundle size? (do you know any jsoo incantation?)
- Is there a way to reuse the same worker? I'm assuming the two workers share a lot of code

> Currently supported features of Merlin

What are other Code Mirror features that Merlin could support that are not implemented in this PR? Could we document them somewhere in case someone would like to improve the playground after this PR is merged?

voodoos:
> One suggestion I have would be to have a "Problems" panel alongside the "Output" one to list all the errors detected by Merlin. That can come in another PR though.

Agreed that it would be very nice! However I probably won't be able to work on it anytime soon...

> I'm slightly concerned about the asset size. I don't think it will be a blocker since we're already sending a 3mb file... but still it would be nice to do some damage control.
>     * Did you have any idea on how we could decrease the bundle size? (do you know any jsoo incantation?)
>     * Is there a way to reuse the same worker? I'm assuming the two workers share a lot of code

I think it is a fairly small bundle regarding what's in: the complete collection of the Stdlib's CMTs.
I don't see how we could easily decrease it further. We could eventually load the CMTs asynchronously (it is 90% of the size of the bundle). But I don't think it is really worth it. Improvement could be made to the UI by showing a message saying "the toplevel is loading".

Workers load asynchronously so it doesn't slows the "first paint" and Merlin will be ready before the toplevel wich is a good thing IMO.

We could merge the workers but this would require some work, would delay Merlin's readiness and probably won't save a lot of weight: the only common part are the part of the runtime used by both workers.

However it looks like the toplevel's worker loading is blocking the editor from becoming responsive in the current version of the playground which is not satisfying since the javascript for the editor itself is only 120k.

> Currently supported features of Merlin
> What are other Code Mirror features that Merlin could support that are not implemented in this PR? Could we document them somewhere in case someone would like to improve the playground after this PR is merged?

The missing features that could be added:
- Jump to definition (inside the buffer, and maybe to the Stdlib sources ?)
- Fetch documentation comments from the Stdlib
- Load arbitrary CMT files from network
- Occurrences
- Destruct / Construct

I will probably work on some of these during the next hacking days.

tmattio:
I agree on keeping the workers separate. I've moved the toplevel worker initialisation so we don't wait for it to paint the editor. It would still be nice to investigate wether we can reduce the bundle sizes further, but since the workers are loaded async, let's not block the PR for this.

This is ready to go I think, thanks again for all your work on this :grin:

PR #538, opened 2022-08-08
Add OCaml Scientific Computing book

Leonidas-from-XIV:
As mentioned in the [OCaml Discuss](https://discuss.ocaml.org/t/ocaml-website-owl-book-not-listed/10274).


tmattio:
Thanks a lot @Leonidas-from-XIV!

A couple of thoughts:

- Let's keep the list of featured books to 3 so as to keep the Learn page to a reasonable size. I'd suggest removing More OCaml from the list and add OCaml Scientific Computing instead. More OCaml and OCaml From the Very Beginning are from the same author and I think featuring a books for beginners makes a bit more sense that an advanced one.
- Let's make the book's description a bit more concise. I think we could keep the first and third paragraph only, readers can read the table of content on the book's purchase page if the description caught their interest.

Leonidas-from-XIV:
Very good points, I've added these. Especially the long wall of text was actually more hurting than helping.

What I am somewhat unclear is how to find the books that are not featured. And who decides how on the amount of stars awarded to each book.

tmattio:
Thanks! I've pushed a fix for the layout on the Learn page and updated the link's name. This is ready to go 🙂 

> What I am somewhat unclear is how to find the books that are not featured.

They can be found at https://ocaml.org/books, which is linked in the footer, the Learn OCaml page, and the different resources sections (homepage, industrial users, etc.)

> And who decides how on the amount of stars awarded to each book.

They are taken from the books' Amazon page IIRC.

PR #540, opened 2022-08-09
Render all book links

Leonidas-from-XIV:
Some of the books have multiple links, it would be nice to display them all. The rendering looks ok, so no need for additional markup fortunately.


tmattio:
Looks great, thank you!

PR #541, opened 2022-08-09
Make the "read online" link consistent

Leonidas-from-XIV:
The other links already use "Read Online" phrasing, so it is nice to be consistent and also the text is slightly shorter.


tmattio:
Thank you!

PR #542, opened 2022-08-09
Remove "OCaml: Programação Funcional na Prática"

Leonidas-from-XIV:
The book website is gone and the shop that used to sell it doesn't seem to have it on its website any more either.


tmattio:
Thank you!

PR #551, opened 2022-08-11
Redirect to the latest documented version of a package

### LINKED ISSUE #174: A redirect to the latest documentation of a package
Currently we only have a redirect to the latest package's "home" page i.e. at the time of writing https://v3.ocaml.org/p/irmin goes to https://v3.ocaml.org/p/irmin/2.8.0. 

We might also want a redirect that takes people directly to the latest documentation of a package so perhaps https://v3.ocaml.org/p/irmin/docs goes to https://v3.ocaml.org/p/irmin/2.8.0/doc/index.html -- this would be useful for people to add to their READMEs/Websites pointing to the latest API of a package.

We should also perhaps advertise these redirects on the packages landing page for people to use (I already do that when looking up APIs for example)

cc: @craigfe (who was looking for something like this)

basseches:
Resolves #174

This PR implements a redirect from /p/package/docs to the latest documented version of a package. If the latest version is undocumented, /p/package/docs redirects to the last version that was documented (i.e. currently, Irmin's latest version—3.3.2—is undocumented, so my code redirects https://v3.ocaml.org/p/irmin/docs to https://v3.ocaml.org/p/irmin/3.3.1/doc/index.html).

(p.s. I wasn't positive about referencing a previous version if the latest version is undocumented, so the first commit on this PR implements a redirect to the latest version's documentation—even if it doesn't exist.)

### DIFF
--- a/src/ocamlorg_frontend/url.ml
+++ b/src/ocamlorg_frontend/url.ml
@@ -2,6 +2,7 @@ let index = "/"
 let packages = "/packages"
 let packages_search = "/packages/search"
 let package v = "/p/" ^ v
+let package_docs v = "/p/" ^ v ^ "/docs"
 let package_with_univ hash v = "/u/" ^ hash ^ "/" ^ v
 let package_with_version v version = "/p/" ^ v ^ "/" ^ version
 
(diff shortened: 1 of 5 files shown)


tmattio:
Thanks @basseches!

I changed the URL to `/doc` to be consistent with the other documentation URL, and I moved the code to `ocamlorg_package`. Merging now!

PR #556, opened 2022-08-20
Compilation: order of source files matter.

0xRyN:
Made a small change to help fix the `Unbound module` error. 

It should be mentioned that the order of files given to the compiler is important.

```ocamlopt -o hello bmodule.ml amodule.ml```

instead of

```ocamlopt -o hello amodule.ml bmodule.ml```

will result in an `Unbound module` error.

### DIFF
--- a/data/tutorials/lg_03_modules.md
+++ b/data/tutorials/lg_03_modules.md
@@ -41,7 +41,7 @@ We can compile the files in one command:
 $ ocamlopt -o hello amodule.ml bmodule.ml
 ```
 
-Note: The source files need to be ordered with the dependencies coming before
+Note: It's necessary to place the source files in the correct order. The dependencies must come before
 the dependent. In the example above, putting `bmodule.ml` before `amodule.ml`
 will result in an `Unbound module` error.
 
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: data/tutorials/lg_03_modules.md [hunk 1]
```diff
@@ -41,6 +41,10 @@ We can compile the files in one command:
 $ ocamlopt -o hello amodule.ml bmodule.ml
 ```
 
+Note: The source files need to be ordered with the dependencies coming before
+the dependent. In the example above, putting `bmodule.ml` before `amodule.ml`
+will result in an `Unbound module` error.
```
christinerose:
```suggestion
Note: It's necessary to place the source files in the correct order. The dependencies must come before
the dependent. In the example above, putting `bmodule.ml` before `amodule.ml`
will result in an `Unbound module` error.
```

christinerose (approved):

tmattio:
Thank you both!

data/tutorials/lg_03_modules.md
christinerose: The original sentence works, if you prefer to keep it. My suggestion breaks it up into two sentences to emphasize the importance of ordering in general and then the necessary order of dependencies/dependent.

tmattio: I like your suggestion of breaking the sentence up in two 🙂 I applied the changes and merged!

PR #557, opened 2022-08-20
docs: add note on ocamlopt argument order in modules tutorial

glennsl:
Based on an issue reported on here: https://stackoverflow.com/questions/73429850/unbound-module-personal-files


glennsl:
Seems I'm too late, this is a duplicate of #556. Sorry!

PR #554, opened 2022-08-15
Diskuv OCaml docs when OCaml 5 is released

jonahbeckford:
This is how I think Diskuv OCaml and the rest of the Windows options should look after the platform installer + Opam 2.2 + OCaml 5 are released. This accompanies and replicates much of #539 (but it explicitly de-elevates Diskuv OCaml).

PR #562, opened 2022-09-02
Update Getting Started instructions with modern shell invocation and OPAM 1.2.2 note

dra27:
cf. https://github.com/ocaml/ocaml/issues/11532.

- Switch to using `eval $(opam env)` - the website's styling for code (which adds backticks) made the text _extremely_ confusing!
- Add a pre-checking note for any users of an out-of-date distro who pick up OPAM 1.2.2 (AFAIK Debian 9 was the last distro still shipping OPAM 1.2.2 has been EOL since July this year). I think it's better to put the note there than pollute the instructions later on to handle an error which virtually no user should see.

### DIFF
--- a/data/tutorials/gs_00_up_and_running.md
+++ b/data/tutorials/gs_00_up_and_running.md
@@ -65,8 +65,7 @@ $ port install opam
 
 It's easy to install opam is with your system's package manager on
 Linux (e.g., `apt-get install opam` or similar). [Details of all installation
-methods.](https://opam.ocaml.org/doc/Install.html)
-
+methods](https://opam.ocaml.org/doc/Install.html). All supported Linux distributions package at least version 2.0.0 (you can check by running `opam --version`). If you are using an unsupported Linux distribution, please either download a precompiled binary or build opam from sources.
 
 ```shell
 # Ubuntu


data/tutorials/gs_00_up_and_running.md
```diff
@@ -36,27 +36,27 @@ port install opam
 
 **For Linux** the preferred way is to use your system's package manager on
 Linux (e.g `apt-get install opam` or similar). [Details of all installation
-methods.](https://opam.ocaml.org/doc/Install.html)
+methods.](https://opam.ocaml.org/doc/Install.html). All supported Linux distributions package at least version 2.0.0 (you can check by running `opam --version`). If you are using an unsupported Linux distribution, please either download a precompiled binary or build opam from sources.
```
gasche: Should this be in the "For Linux" category? My gut feeling is that the main source of severely outdated opam would be a virtual image of an old system, can't MacOS + brew have the same issue?

(Reading this section again, my impression is that the "For macOS / For Linux" split could be part of its own subsection "Install opam" before "Install OCaml", and that this subsection would be the right place to have this warning.)

gasche (commented):
Thanks! I was thinking of opening an issue but you bested me with a full PR.

gasche:
Note: @dra27 while you are at it, it might be that the Windows part of this page would benefit from a refresh as well: it links to the fdopen repository, which itself includes a large banner that it is discontinued, this could be worrying for users. Maybe for now removing the "if you need Windows binaries" paragraph would be the safer choice. (I guess this is a separate issue / separate PR, but you are a great person to talk to about it.)

gasche:
(Regarding Windows, the [Diskuv installer](https://github.com/diskuv/dkml-installer-ocaml) may deserve a mention, but it may be a bit experimental still -- as everything on Windows I guess.)

dra27:
This isn't _precisely_ the time/place to sort out the Windows part! I'm presently doing quite a good "cat typing at the keyboard" act, and the Windows OCaml packages are part of that (being tied in with both opam 2.2 and relocatable compiler). See also https://github.com/ocaml/ocaml.org/pull/554

data/tutorials/gs_00_up_and_running.md
dra27: I don't think it applies to macOS. On an out-of-date Debian, assuming `apt-get update` works at all out of the box of the box (eventually the repos get moved to archives), you will get OPAM 1.2.2 because it uses an old repository to go with the old distro. On macOS you get the same repository whichever version of macOS you're running and, even if for some reason opam 2.x wouldn't install, brew and macports only carry one version of a package, so you'd either get the latest version of opam regardless or an error trying to install it.

gasche: Okay, sounds reasonable.

gasche (approved):

kit-ty-kate (approved):

tmattio:
Thanks @dra27!

We had rewritten a large part of the Up and Running tutorial to make it more accessible to beginners, and it seems that your first commit had been addressed while doing so.

I've just merged the new version in https://github.com/ocaml/ocaml.org/pull/565 and rebased your PR on top of it. Let me know if the new diff looks good to you and I can merge the PR 🙂

dra27:
Even better, yes!

PR #563, opened 2022-09-02
Remove metaprogramming frameworks from Platform

tmattio:
Extracted from `staging`.

As discussed with @pitag-ha and @panglesd, the recommended metaprogramming method in OCaml is better documented as a guide, which is now available at https://ocaml.org/docs/metaprogramming.

We can now remove Ppxlib (and other metaprogramming frameworks) from the Platform, which now only contains developer tools.

### DIFF
--- a/data/tools.yml
+++ b/data/tools.yml
@@ -101,23 +101,6 @@ tools:
       Merlin is an assistant for editing OCaml code. It aims to provide the features available in modern IDEs: error reporting, auto completion, source browsing and much more.
     lifecycle: active
 
-  - name: Ppxlib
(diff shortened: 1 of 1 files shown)


panglesd:
Looks good to me!

Note that in https://ocaml.org/docs/metaprogramming, there is no mention of `camlp4` and `omp`, which are also removed from the platform.

tmattio:
> there is no mention of camlp4 and omp, which are also removed from the platform.

Good catch, perhaps we should add a note in that document mentioning that they were the previously recommended methods?

panglesd:
I will try to add a mention on this without interrupting the flow of the guide!

PR #575, opened 2022-09-20
only show the learn layout TOC sidebar on screens large enough to accommodate it

sabine:
The TOC sidebar now only appears on xl screens (instead of already on lg screens) and I allowed the TOC to take up a bit more space on the right hand side. (so that the whitespace to the right side is equal to the whitespace to the left side of the screen).

I tried to make the sidebars narrower on lg screens to cram it all onto the same page, but the experience was not great. Code is being cut off, which means that people need to scroll sideways. For comparison: Tailwind's documentation page also shows the TOC sidebar only on xl screens.

before lg             |  after lg
Screenshot 2022-09-20 at 10-40-20 Modules · OCaml Tutorials  |  Screenshot 2022-09-20 at 10-40-25 Modules · OCaml Tutorials

before xl             |  after xl
Screenshot 2022-09-20 at 10-40-59 Modules · OCaml Tutorials | Screenshot 2022-09-20 at 10-40-53 Modules · OCaml Tutorials

### DIFF
--- a/src/ocamlorg_frontend/pages/tutorial.eml
+++ b/src/ocamlorg_frontend/pages/tutorial.eml
@@ -43,9 +43,9 @@ Learn_layout.render ~current_tutorial:tutorial.slug ~title:(Printf.sprintf "%s Â
     </div>
   </div>
 </div>
-<div class="hidden lg:flex top-0 sticky h-screen">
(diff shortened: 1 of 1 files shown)


tmattio:
Thanks @sabine, this is much better!

Issue #148, opened 2021-10-07
HEAD requests yield 404s


TheLortex:
This is probably a Dream issue, but it's been encountered here: https://discuss.ocaml.org/t/ann-release-of-odoc-2-0-0/8582/5

Basically, `curl -sI https://v3.ocaml.org` returns 404, but `curl -sL https://v3.ocaml.org` works correctly. 
The difference ? `curl -sI` sends a `HEAD` http request, which is supposed to return the headers if a `GET` request would have been done.

TheLortex:
After experimenting a bit, I don't know if it's `dream`'s problem. Surely it's possible to derive `HEAD`'s output from the `GET` handler, but this is not efficient when the output is large. 
For each route we should provide a `Dream.head` handler that computes the output headers either by deriving it from the `GET` handler or by doing something smarter (for static files for example we can already know the `Content-Length` in advance without loading the file)

tmattio:
Should be relevant: https://github.com/rgrinberg/opium/blob/master/opium/src/middlewares/middleware_head.ml

PR #576, opened 2022-09-20
Handle HEAD requests

cuihtlauac:
Fix for https://github.com/ocaml/ocaml.org/issues/148

For each GET route added, add a twin HEAD route. By default dream
does not respond to HEAD request, unless if a ad-hoc route exists.
We probably don't want to maintain a separate set of routes for
GET and HEAD. Therefore, instead of calling Dream.get we call local
function get_and_head, which creates two routes, one for GET requests
and one for HEAD requests.

Dream documentation (dream.mli) says: "Middlewares are run only if
a route matches." Therefore, a turning HEAD requets into GET requests
using a middleware will not work as HEAD requests are discarded before
being sent through any middleware.

### DIFF
--- a/src/ocamlorg_web/lib/middleware.ml
+++ b/src/ocamlorg_web/lib/middleware.ml
@@ -13,7 +13,7 @@ let head handler request =
       let open Lwt.Syntax in
       let* response = handler request in
       let transfer_encoding = Dream.header response "Transfer-Encoding" in
-      let* _ =
+      let* () =
         if
           transfer_encoding = Some "chunked"
           || Dream.has_header response "Content-Length"
--- a/src/ocamlorg_web/lib/middleware.ml
+++ b/src/ocamlorg_web/lib/middleware.ml
@@ -7,17 +7,21 @@ let no_trailing_slash next_handler request =
         Dream.redirect request (String.sub target 0 (String.length target - 1))
       else next_handler request
 
-let head handler request = match Dream.method_ request with
-| `HEAD ->
-    let open Lwt.Syntax in
-    let* response = handler request in
-    let transfer_encoding = Dream.header response "Transfer-Encoding" in
-    let* _ =
-      if transfer_encoding = Some "chunked" || Dream.has_header response "Content-Length" then
-        Lwt.return_unit
-      else
-        let+ body = Dream.body response in
-        body |> String.length |> string_of_int |> Dream.add_header response "Content-Length"
+let head handler request =
+  match Dream.method_ request with
+  | `HEAD ->
+      let open Lwt.Syntax in
+      let* response = handler request in
+      let transfer_encoding = Dream.header response "Transfer-Encoding" in
+      let* _ =
+        if
+          transfer_encoding = Some "chunked"
+          || Dream.has_header response "Content-Length"
+        then Lwt.return_unit
+        else
+          let+ body = Dream.body response in
+          body |> String.length |> string_of_int
+          |> Dream.add_header response "Content-Length"
       in
-    Dream.empty ~headers:(Dream.all_headers response) (Dream.status response)
-| _ -> handler request
+      Dream.empty ~headers:(Dream.all_headers response) (Dream.status response)
+  | _ -> handler request
--- a/src/ocamlorg_web/lib/middleware.ml
+++ b/src/ocamlorg_web/lib/middleware.ml
@@ -11,6 +11,7 @@ let head handler request =
   match Dream.method_ request with
   | `HEAD ->
       let open Lwt.Syntax in
+      Dream.set_method_ request `GET;
       let* response = handler request in
       let transfer_encoding = Dream.header response "Transfer-Encoding" in
       let* () =
--- a/src/ocamlorg_web/lib/middleware.ml
+++ b/src/ocamlorg_web/lib/middleware.ml
@@ -6,3 +6,18 @@ let no_trailing_slash next_handler request =
       if String.ends_with ~suffix:"/" target then
         Dream.redirect request (String.sub target 0 (String.length target - 1))
       else next_handler request
+
+let head handler request = match Dream.method_ request with
+| `HEAD ->
+    let open Lwt.Syntax in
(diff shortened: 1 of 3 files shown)


tmattio:
Thanks @cuihtlauac!
This would need a middleware in addition to the route changes, right?
Currently, we'd reply to the HEAD requests with the full body

cuihtlauac:
My impression is HEAD requests currently all have 404 and zero length answers. This, to me, is consistent with the code: no calls to Dream.head and no implicit support for HEAD in Dream through Dream.get. If Dream.head is correctly implemented (which I haven't checked) no middleware is needed; if it is bugged, a middleware may work around it.

And now I'm doubtful, am I completely wrong? :-)

cuihtlauac:
Neither Dream nor ocaml.org properly handles HEAD requests. When adding ad-hoc routes for HEAD requests, the content is carried along with the 200 response, as in a GET answer, which is a violation of the HTTP standard (RFC9110/section 9.3.2: The HEAD method is identical to GET except that the server MUST NOT send content in the response).
```sh
echo -n "HEAD / HTTP/1.1\r\nUser-Agent: dummy\r\n\r\n" | nc -N localhost 8080
```

cuihtlauac:
Dream creates Content-Length: header automatically to match the provided body length in a way which does not allow building valid HTTP responses to HEAD requests. Neither handler nor middleware are able have a non zero Content-Length with a empty body (as required by the RFC). Manually adding a Content-Length header (using `Dream.set_header`) to an empty body response will result in having multiple Content-Length headers and dropping the body (using `Dream.set_body`) from a regular response will set the Content-Length header to zero.

It should be an issue in Dream.

cuihtlauac:
In the latest push, two semi alternative implementations are provided.
1. eedb7cc use a higher-order function to create handlers for HEAD requests from handlers for GET requests
2. 5a1206f use middleware to turn responses for HEAD requests into valid ones.

Interestingly, both rely on the same function: `head_handler`

src/ocamlorg_web/lib/router.ml
```diff
@@ -23,66 +23,81 @@ let media_loader =
       let* store in
       Media.last_modified store (Mirage_kv.Key.v path))
 
+let head_handler handler request =
+  let open Lwt.Syntax in
+  let* response = handler request in
+  let* body = Dream.body response in
+  let length = body |> String.length |> string_of_int in
+  Dream.drop_header response "Content-Length";
```
tmattio: Why drop the Content-Length if it's already here? In some cases, we probably want to keep if, for instance if we know the size of a file and return a stream (we won't be able to find the length of it without evaluating the stream in that case)

src/ocamlorg_web/lib/router.ml
```diff
@@ -23,66 +23,81 @@ let media_loader =
       let* store in
       Media.last_modified store (Mirage_kv.Key.v path))
 
+let head_handler handler request =
+  let open Lwt.Syntax in
+  let* response = handler request in
+  let* body = Dream.body response in
```
tmattio: I think this will evaluate the body if it's a stream, which we don't want to do. To avoid doing this, we can check if we find a `Transfer-Encoding: chunked` header and only remove the body without calculating the content length if we do.

src/ocamlorg_web/lib/router.ml
```diff
@@ -23,66 +23,81 @@ let media_loader =
       let* store in
       Media.last_modified store (Mirage_kv.Key.v path))
 
+let head_handler handler request =
```
tmattio: This is actually a middleware (a function that takes a handler and a request).

I have a preference for using the middleware approach if both the HOF approach and the middleware one work equally. So I would suggest inlining this inside the `head_middleware` function and moving the later to the `Middleware` module.

tmattio (commented):
Thanks for the detective work @cuihtlauac!

I think we could go with the middleware approach if that works, I left a few suggestions to improve it further to avoid problems with streamed requests

src/ocamlorg_web/lib/router.ml
cuihtlauac: Actually, I haven't observed a Content-Length here. I fail to understand why yet, intuitively, it should be there.

tmattio: If I understand correctly, content-length is added by Dream **if it's not already present**. So you won't observe a Content-Length at the middleware level unless the handler explicitly sets it. I'm not sure we do this at the moment, but I can see a few use cases where it would be useful, for instance when streaming a static file.

src/ocamlorg_web/lib/middleware.ml
```diff
@@ -7,17 +7,21 @@ let no_trailing_slash next_handler request =
...
-      else
-        let+ body = Dream.body response in
-        body |> String.length |> string_of_int |> Dream.add_header response "Content-Length"
+let head handler request =
+  match Dream.method_ request with
+  | `HEAD ->
+      let open Lwt.Syntax in
+      let* response = handler request in
+      let transfer_encoding = Dream.header response "Transfer-Encoding" in
+      let* _ =
```
tmattio: Just to enforce that the else branch returns a unit and not a new body
```suggestion
      let* () =
```

src/ocamlorg_web/lib/router.ml
```diff
@@ -23,71 +23,71 @@ let media_loader =
       let* store in
       Media.last_modified store (Mirage_kv.Key.v path))
 
-let get_and_head pattern handler = [
-  Dream.get pattern handler;
-  Dream.head pattern handler
-]
+let get_and_head pattern handler =
```
tmattio: We shouldn't need this if the middleware is applied before the router and has the following behaviour:

- If the request is HEAD, change it to GET and pass it to the handler (plus all the logic in the head middleware)
- Else, pass it to the handler

tmattio (approved):
@cuihtlauac I tried applying the middleware globally and it seems to be working.

The PR looks good to me, so if this is working on your side too, feel free to merge 🙂

PR #579, opened 2022-09-22
anchor target highlighting and clickability of anchors in package docs

sabine:
Use case: Person A kindly answers person B's OCaml question (e.g. on the OCaml discord, on StackOverflow, Slack, or wherever) and includes a link to a particular position in the documentation of a package.

before              |  after 
Screenshot from 2022-09-26 08-18-01 the hashtag symbol on Anchors was in an unsuitable position on h1,h2,h3,h4,h5,h6 Elements. The only way to get a link to a particular heading on the package docs was to click the little Hashbang. See https://ocaml.org/p/streaming/0.8.0/doc/index.html#streaming   |  Screenshot 2022-09-22 at 17-44-15 Documentation · streaming 0 8 0 · OCaml Packages The anchor of h1,h2,h3,h4,h5,h6 "fully covers" the heading, so the user can click on the heading to get a link to that particular section. The targeted section is now highlighted (like a text marker) to make it easier for the recipient of a targeted link to see which section they are being referred to.
Screenshot 2022-09-22 at 17-54-17 Streaming Flow · streaming 0 8 0 · OCaml Packages It is hard to see which definition is highlighted. See https://ocaml.org/p/streaming/0.8.0/doc/Streaming/Flow/index.html#val-identity | Screenshot 2022-09-22 at 17-52-34 Streaming Flow · streaming 0 8 0 · OCaml Packages

Further improvements to anchor clickability in package docs (on the type and function definitions generated by `odoc`) require upstream changes. The same applies to adding clickable anchors to section headings on the tutorials (which requires an upstream change to `omd`).

### DIFF
--- a/asset/css/doc.css
+++ b/asset/css/doc.css
@@ -11,7 +11,7 @@ div.odoc-include .spec {
   border-color: rgba(251, 146, 60, 1);
   border-radius: 0.25rem /* 4px */;
   background-color: rgba(243, 244, 246, 1);
-  padding-left: 1rem /* 16px */;
+  padding-left: 1.7rem /* needs to provide enough space for the hashtag anchor */;
   padding-right: 1rem /* 16px */;
   padding-top: 0.5rem /* 16px */;
   padding-bottom: 0.5rem /* 16px */;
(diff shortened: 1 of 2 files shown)


tmattio:
Thanks @sabine! The experience on streaming is much better, but it seems to be breaking the anchor on some pages, for instance: http://localhost:8080/p/base/v0.15.0/doc/Base/Applicative/Make2_using_map2/index.html (the anchors are all positioned at the top of the page and clicking on it scrolls to the latest signature of the page)

I think we could also add some more space between the hashtag and the element, what do you think?

sabine:
I couldn't reproduce the issue with anchors being stuck at the top of the page. It's necessary to open your network tab and reload the page with cache disabled (to make sure no old styles linger). If you still find a problem, please drop me a link that includes a broken hashtag.

Either way, looks like I had removed the on-hover hashtags from the headings. I brought them back in the force-push. Reason I removed them was that there wasn't enough space and they would be cut off. To accommodate them, I changed the left-padding on the content area.


Screenshot from 2022-09-23 14-45-21

Screenshot from 2022-09-23 14-49-04


ETA: A simpler/less hacky way to get easily-clickable anchors is if odoc would generate `<h1><a class="anchor" href="#[slug]">[title]</a></h1>` instead of `<h1><a class="anchor" href="#[slug]"></a>[title]</h1>`.

sabine:
I modified the styles some more to have a bigger hashtag on the spec values.


Screenshot from 2022-09-23 15-00-08

In the longer run, we want `odoc` to emit HTML where the anchor wraps around the name of the spec item (in this case: `all`), so that clicking this gives us a deep-link to the spec item.

tmattio (approved):
This is working great! Thanks a lot for doing this! 🎉

PR #580, opened 2022-09-23
package documentation status title attribute and inactive link

### LINKED ISSUE #494: Add a tooltip for package documentation status
I think it could help with a tooltip or some text to explain what the coloured dots mean.

https://github.com/ocaml/ocaml.org/blob/2f673ed2183c9f948b331f126dcf39bf52d2fa55/src/ocamlorg_frontend/layouts/package_layout.eml#L60-L70

sabine:
Resolves https://github.com/ocaml/ocaml.org/issues/494 and simplifies the markup by removing the unnecessary `button` tag inside `a`.

before              |  after 
Screenshot from 2022-09-23 09-06-41 "Documentation" tab is clickable and links to a 404 page, even when we know that documentation is not available. There is no indication of what the colored dot means. | Screenshot from 2022-09-23 09-05-48 "Documentation" tab is inactive and not clickable when there is no documentation available. A tooltip (title attribute) tells us why.

### DIFF
--- a/src/ocamlorg_frontend/layouts/package_layout.eml
+++ b/src/ocamlorg_frontend/layouts/package_layout.eml
@@ -12,7 +12,18 @@ let render
 ~(package : Package_intf.package)
 inner =
 let tab_class = "border-transparent border-b-2 hover:border-primary-600 p-4 font-medium relative text-body-400" in
+let unavailable_tab_class = "border-transparent border-b-2 p-4 font-medium relative text-gray-400" in
 let current_tab_class = "border-b-2 hover:border-primary-600 p-4 font-medium border-primary-600 relative text-primary-600" in
+let documentation_dot = match documentation_status with 
+  | `Failure -> "<span class=\"bg-red-500 ml-2 absolute top-2 p-1 rounded-full inline-block\"></span>"
+  | `Success -> "<span class=\"bg-green-500 ml-2 absolute top-2 p-1 rounded-full inline-block\"></span>"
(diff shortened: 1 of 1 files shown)


src/ocamlorg_frontend/layouts/package_layout.eml
```diff
@@ -48,29 +54,29 @@ Layout.render
...
+            | `Success -> %>
+            <a href="<%s Url.package_doc package.name package.version "index.html" %>"
+              class="<%s if tab = Documentation then current_tab_class else tab_class %>">
+                Documentation
+                <%s! documentation_dot %>
+            </a>
+          <% | _ -> %>
+            <span class="<%s unavailable_tab_class %>"
+              title="<%s match documentation_status with
+                | `Failure -> "documentation failed to build"
```
tmattio: We might still want to have a link to the documentation if it's a failure: the page then displays the documentation error, which can be useful for package authors (and the documentation pipeline maintainers) to debug the issue

sabine: Can you show me one such case? So far, all the ones I saw were 404, but that may be because I haven't come across the correct one yet.
The one you mean is in case we have a build failure? Then indeed, I'll change that!

sabine: Only the `Unknown` documentation status is now a deactivated tab. Indeed, I hadn't considered that in case the docs server provides the build errors, we still need to link to that.

tmattio (approved):
Looks great, thank you! I wasn't able to find an example of a documentation generation failure, but reading the code, it looks like the behaviour would be correct 🙂

Issue #583, opened 2022-09-25
docs generation: error with mld files in title


avsm:
See the monorepo docs: https://ocaml.org/p/opam-monorepo/0.3.2/doc/lock.html

The "module lock.html" is an incorrect title for the page.

<img width="1144" alt="image" src="https://user-images.githubusercontent.com/53164/192140659-349b14bd-1f8a-42db-91f2-564e89981091.png">

sabine:
Thanks for reporting, that's indeed a confusing sight.

sabine:
Interestingly, for this particular package, the docs also seem to be missing the first section:
https://github.com/tarides/opam-monorepo/blob/feeb325c9c8d560c6b92cbde62b6a9c5f20ed032/doc/lock.mld

And it's hard to find the other files in the docs folder as they are only linked inside the content of the index page. :thinking: But this is a topic for the sidebar rework.

Issue #203, opened 2022-01-05
Some more documentation regression


gadmm:
Stand-alone documentation pages are back, it's great. Here's another batch of comment regarding regressions in the package documentations:
* Missing title and first paragraph ([old](https://guillaume.munch.name/software/ocaml/memprof-limits/Memprof_limits/index.html) vs. [new](https://v3.ocaml.org/p/memprof-limits/0.2.0/doc/Memprof_limits/index.html); [old](https://guillaume.munch.name/software/ocaml/memprof-limits/recovering.html) vs. [new](https://v3.ocaml.org/p/memprof-limits/0.2.0/doc/recovering.html))
* You cannot go back to the toplevel page using the table of contents from other pages (nor using the page hierarchy on the top right): the only way is to click on the "Documentation" tab. (See e.g. https://v3.ocaml.org/p/memprof-limits/0.2.0/doc/Memprof_limits/index.html)
* Stand-alone page shows "Module" on the page hierarchy on the top right; it is not a module. (https://v3.ocaml.org/p/memprof-limits/0.2.0/doc/recovering.html)
* Paragraph title lacks distinct styling: [old](https://guillaume.munch.name/software/ocaml/memprof-limits/index.html#discussion:-precise-vs.-statistical-limits) vs. [new](https://v3.ocaml.org/p/memprof-limits/0.2.0/doc/index.html#discussion:-precise-vs.-statistical-limits). (Old one uses sans vs. serif; but just having it in bold is enough.)
* Missing syntax highlighting in the documentation

gadmm:
I just tested the now live version of ocaml.org. The first issue is still present but it is pretty critical.

Some of the other issues are still present. I noticed that paragraph titles have been styled, thanks.

jonludlam:
I belive I noticed this missing preamble in passing when upstreaming into odoc - so it should be fixed in part by https://github.com/ocaml/odoc/pull/842

gadmm:
Do you mean with the title too? Please let me know when it is online so that I can have a look.

gadmm:
Checking again, the critical issue of the title and the first paragraph missing is still there :) (compare [old](https://guillaume.munch.name/software/ocaml/memprof-limits/recovering.html) vs. [new](https://v3.ocaml.org/p/memprof-limits/0.2.0/doc/recovering.html))

In addition the new title is odd ("module recovering.html").

tmattio:
[...]

PR #585, opened 2022-09-26
omit breadcrumbs on package docs when there are no libraries

sabine:
Fixes (ETA: the problem of showing breadcrumbs for packages without a hierarchy seen in) https://github.com/ocaml/ocaml.org/issues/583 by not rendering breadcrumbs for packages without libraries.

When there are no libraries/modules in a package, we should omit the breadcrumbs because they don't make sense from a UI/UX perspective.

### DIFF
--- a/src/ocamlorg_frontend/pages/package_documentation.eml
+++ b/src/ocamlorg_frontend/pages/package_documentation.eml
@@ -41,7 +41,7 @@ Package_layout.render
       </div>
     </div>
     <div class="flex-1 overflow-hidden z-0 z- w-full lg:max-w-4xl mx-auto relative md:px-6 md:pl-12">
-      <%s! Breadcrumbs.render path %>
+      <%s! if maptoc != [] then Breadcrumbs.render path else "" %>
       <div class="odoc prose prose-orange">
         <%s! content %>
       </div>
--- a/src/ocamlorg_frontend/pages/package_documentation.eml
+++ b/src/ocamlorg_frontend/pages/package_documentation.eml
@@ -41,7 +41,7 @@ Package_layout.render
       </div>
     </div>
     <div class="flex-1 overflow-hidden z-0 z- w-full lg:max-w-4xl mx-auto relative md:px-6 md:pl-12">
-      <%s! if maptoc != [] then Breadcrumbs.render path else "" %>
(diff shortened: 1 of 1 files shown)


tmattio:
Thanks @sabine! I don't think the PR fixes #583 entirely though: even if a package has a library, all the mld files will be rendered incorrectly.

IIRC, the title comes from the package.json file, generated by the documentation pipeline. I'd look upstream to see if we could change the title for mld files. cc @TheLortex as you worked on this part of the pipeline, maybe you have pointers?

sabine:
Ah interesting, so there is a deeper underlying issue for these strange titles. :+1:

I'll stand by that it makes sense to omit breadcrumbs when there is no hierarchy, but indeed, the title not making sense in the first place is another thing.

sabine:
Ah but wait, there's no other way than breadcrumbs to see this mysterious title :facepalm:, so alright... what's strange is that our breadcrumbs show "module" (and the content of the title itself is another matter).

tmattio (approved):

tmattio:
> Ah but wait, there's no other way than breadcrumbs to see this mysterious title

Yeah, that's a bit hacky.. This was to work around https://github.com/ocaml/ocaml.org/issues/203#issuecomment-1242360713, but when we can display proper titles, the breadcrumb should become more "breadcrumb-y" 😅

sabine:
Ok good, then let's do this as a small fix and subsequently do a bigger cleanup that resolves the conflation between module pages and manual pages.

TheLortex:
The title is derived from the module path, which is itself derived from the request path. Any path section that isn't `index.html`, or doesn't start with `module-type` or `argument-<number>-` is considered to be a module path.
Here is the code that is at fault: https://github.com/ocaml/ocaml.org/blob/3cc3ed3bd6c91669beabec6fbcb3c517abb367e2/src/ocamlorg_package/lib/ocamlorg_package.ml#L227-L238

sabine:
@TheLortex sorry, I overlooked this. Yes, indeed, you are right, the deeper cause is that we cannot distinguish from the file name alone whether it came from a `.mld` file or from a module. I have a PR in the pipeline to fix this upstream in `odoc`.

PR #586, opened 2022-09-26
add Why OCaml? link on the Learn_layout's icon sidebar

sabine:
This PR adds a simple "Why OCaml?" link to the sidebar. Eventually, we may get a designer to make it pretty, but at the time being, I think the question mark will do.

I deliberated a bit on where to put it, and I think at the bottom is the correct position, as it's a "catch-all" for people who still need some motivation to learn.

@christinerose is this what you had in mind in https://github.com/ocaml/ocaml.org/pull/553?

before              |  after 
Screenshot 2022-09-26 at 12-17-23 Learn OCaml | Screenshot 2022-09-26 at 12-17-10 Learn OCaml there is a link "Why OCaml?" in the icon sidebar that leads to the corresponding page.

### DIFF
--- a/src/ocamlorg_frontend/layouts/learn_layout.eml
+++ b/src/ocamlorg_frontend/layouts/learn_layout.eml
@@ -42,6 +42,13 @@ let sidebar_icon_link_block
       </div>
       <div class="ml-3 font-semibold <%s if current != Exercises then "opacity-70" else "" %>">Exercises</div>
     </a>
+
+    <a href="<%s Url.about %>" class="flex justify-start items-center">
+      <div class="bg-primary-600 font-bold text-white rounded w-8 h-8 flex items-center justify-center">
+        ?
+      </div>
+      <div class="ml-3 font-semibold opacity-70">Why OCaml?</div>
+    </a>
   </div>
 
 let sidebar_link


tmattio:
To me, it'd make more sense to put it in the "Getting Started" section of the doc: the top-level links are really linking to different sections of the documentation, whereas "Why OCaml" is a page of the documentation (we could actually update it to use the documentation layout!)

FWIW, that's what the Rust documentation does: the general intro, which details why you should use Rust, is part of the documentation (https://doc.rust-lang.org/book/ch00-00-introduction.html)

sabine:
"Why OCaml?" is the core of our "pitch" and it's a beautiful page on its own. And we can expand it to make it even more compelling. In the long run this could grow into a page that deserves a lot of attention.

In this case, I think Rust isn't a great example, due to its own fragmented ecosystem and branding. They don't have a long-form pitch on https://www.rust-lang.org/ in the first place.

Either way, if we end up adding a "Why Ocaml?" link to "Learn", it makes sense to have the link stand out, because that's a core question: "why should I make the effort to learn all this?", or, when you are struggling doing the exercises: "why am I struggling with this again?". The reader doesn't need a textbook introduction here. They need a quick motivation boost.

Whether we should add "Why OCaml?" on the "Learn" page... let's ask some more people.

---

The "Why OCaml?" page as is has two audiences:
* individual programmers who are considering to learn OCaml, and
* decision makers in non-technical roles (so people who will not learn OCaml, even if their tech stack includes OCaml).

The current "Why OCaml?" page serves both these audiences reasonably well. For programmers, it could be improved by making the "Learn" section easier to reach from the "Why OCaml?" page. And for business decision makers, we should add a call to action so they know where to go, now that they've read what the value proposition of OCaml is.

We could also split "Why OCaml?" into one "Why OCaml?" page for programmers and another "Why OCaml?" page for business people and specialize both of them further. (However, the simpler-to-maintain thing to do is to add both calls to action on the bottom of the current "Why OCaml?" page.)

christinerose:
> 
> @christinerose is this what you had in mind in https://github.com/ocaml/ocaml.org/pull/553?
> 

YES!

christinerose:
> "Why OCaml?" is the core of our "pitch" and it's a beautiful page on its own. And we can expand it to make it even more compelling. In the long run this could grow into a page that deserves a lot of attention.
> 
> 
> Either way, if we end up adding a "Why Ocaml?" link to "Learn", it makes sense to have the link stand out, because that's a core question: "why should I make the effort to learn all this?", or, when you are struggling doing the exercises: "why am I struggling with this again?". The reader doesn't need a textbook introduction here. They need a quick motivation boost.
> 
> 
> 
> Whether we should add "Why OCaml?" on the "Learn" page... let's ask some more people.
> 


I like it on the sidebar - don't have to go searching for it, and it is the core of adoption. This is what we have to push more than anything -- why anyone should try it or learn it. Why when there are so many other tools more readily used. The "Why OCaml" is the most important page for adoption.

christinerose:
> 
> The current "Why OCaml?" page serves both these audiences reasonably well. For programmers, it could be improved by making the "Learn" section easier to reach from the "Why OCaml?" page. And for business decision makers, we should add a call to action so they know where to go, now that they've read what the value proposition of OCaml is.
> 

Yes on the CTA for each group, and I think on the Why OCaml page is the perfect place for that. The Why OCaml page should be easily accessible from anywhere on the site.

tmattio (approved):
Not 100% convinced this should be a top-level link vs a page of the documentation, but I'm happy to go with it if you both think it's better 🙂

PR #587, opened 2022-09-27
Generate canonical links

cuihtlauac:
Pages accessible using different urls shall include a canonical
link, indicating which url is the best (search Google for “consolidate
duplicates urls”). This is useful for web crawlers. Ahrefs reports
“Duplicate pages without canonical” as an issue. This patch adds
canonical links to several pages.

Main tag generation takes place in layout.eml.

Often, it is sufficient to pass the right url to Layout.render. This is
the case for:

* about.eml
* academic_users.eml
* best_pratices.eml
* blog.eml
* books.eml
* community.eml
* industrial_users.eml
* jobs.eml
* learn.eml
* news.eml
* packages.eml
* papers.eml
* platform.eml
* problems.eml
* releases.eml
* tutorial.eml

A few others are almost the same, except some obvious path part needs
to be appended. This is the case for:

* news_post.eml
* release.eml
* success_story.eml
* workshop.eml

Other changes includes parameter forwarding:

* learn_layout.eml
* ocamlorg_frontend.ml
* page.eml
* tutorial.eml

In package_search.eml, the same canonical is used as in packages.eml.

In home.eml, the empty string is used to have "https://ocaml.org" be
the best url (without trailing slash)

In package_version.eml, package version is not used in order to
have the top level pages being referenced.

In handler.ml a combination of parameter forwarding and canonical
url setting takes place.

In playground.eml, the string "https://ocaml.org" is hardwired a
second time because it uses an ad-hoc layout.

No canonical link is generated in not_found.eml, on purpose.

### DIFF
--- a/src/ocamlorg_frontend/layouts/layout.eml
+++ b/src/ocamlorg_frontend/layouts/layout.eml
@@ -1,4 +1,4 @@
-let render ?(use_swiper=false) ?(wide=false) ?description ?styles ~title inner =
+let render ?(use_swiper=false) ?(wide=false) ?description ?styles ~title ?canonical inner =
 <!DOCTYPE html>
 <html lang="en">
   <head>
@@ -19,6 +19,9 @@ let render ?(use_swiper=false) ?(wide=false) ?description ?styles ~title inner =
     <% | None -> ()); %>
     <meta name="theme-color" content="#fff" />
     <meta name="color-scheme" content="white" />
+    <% (match canonical with | Some canonical -> %>
+    <link rel="canonical" href="https://ocaml.org<%s canonical %>" />
+    <% | None -> ()); %>
     <link rel="icon" type="image/x-icon" href="/favicon.ico" />
     <link rel="manifest" href="/manifest.json" />
     <% (match styles with | Some styles -> styles |> List.iter (fun style -> %>
--- a/src/ocamlorg_frontend/layouts/learn_layout.eml
+++ b/src/ocamlorg_frontend/layouts/learn_layout.eml
@@ -81,6 +81,7 @@ let render
 ?styles
 ~title
 ~description
+~canonical
 sidebar_html
 inner_html
 =
@@ -89,7 +90,8 @@ inner_html
   ?styles
   ~wide:true
   ~title
-  ~description @@
+  ~description
+  ~canonical @@
   <div x-data="{ open: false, sidebar: window.innerWidth > 1024 && true, showOnMobile: false}" @resize.window="sidebar = window.innerWidth > 1024">
     <button class="bg-primary-600  p-3 z-30 rounded-r-xl text-white shadow-md top-2/4 fixed lg:hidden left-0"
       :class="sidebar ? 'pl-1 pr-2': ''" x-on:click="sidebar = ! sidebar">
--- a/src/ocamlorg_frontend/layouts/package_layout.eml
+++ b/src/ocamlorg_frontend/layouts/package_layout.eml
@@ -28,7 +28,8 @@ Layout.render
 ?styles
 ~wide:true
 ~title
-~description @@
+~description
+~canonical:(Url.package package.name) @@
 <div class="bg-white">
   <div class="py-5 lg:py-12">
     <div class="container-fluid wide">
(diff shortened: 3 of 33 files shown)


tmattio:
Thanks @cuihtlauac!

Going through the site, I notice a few pages that have the wrong canonical URL:

- `http://localhost:8080/packages/search?q=base` has `https://ocaml.org/packages`
- `http://localhost:8080/p/base/v0.15.1~5.0preview` has `https://ocaml.org/p/base`
- `http://localhost:8080/p/base/v0.15.0/doc/index.html` has `https://ocaml.org/p/base`
- `http://localhost:8080/blog?p=2` has `http://localhost:8080/blog` (not sure if the parameters should be included in the canonical, but these are two different pages)

tmattio:
> In package_search.eml, the same canonical is used as in packages.eml.

Maybe I misunderstand the goal of the canonical URLs, but why did you choose to have the same URL for these?

cuihtlauac:
I'm not a canonical links expert :-) My understanding may be wrong. I've based my code on this:

https://developers.google.com/search/docs/crawling-indexing/consolidate-duplicate-urls

As well as ahrefs docs and a couple of others publicly available resources.

> http://localhost:8080/packages/search?q=base has https://ocaml.org/packages

The `package/search` path is used to display search results, which probably does not make sense to appear in web crawling/search results. We want to go either to the search home page or to the `base` package page, the latter has its own canonical, therefore I pick the former. Alternatively, we could just skip canonical on this one.

> http://localhost:8080/p/base/v0.15.1~5.0preview has https://ocaml.org/p/base
> http://localhost:8080/p/base/v0.15.0/doc/index.html has https://ocaml.org/p/base

When searching in Google for "ocaml base package" the right answer path's should probably be `/p/base` as it is immutable, this is why it's preferred.

> http://localhost:8080/blog?p=2 has https://ocaml.org/blog (not sure if the parameters should be included in the canonical, but these are two different pages)

Again, the page number is not relevant in web search, better forget about it. However, actual blog posts should have their own canonicals (using a slug). 

> > In package_search.eml, the same canonical is used as in packages.eml.
>
> Maybe I misunderstand the goal of the canonical URLs, but why did you choose to have the same URL for these?

That's the issue you pointed in `packages/search?q=base`

rr0gi:
> Again, the page number is not relevant in web search, better forget about it. 

the point of canonicals is to pick among duplicate pages, /blog and /blog?p=2 are different pages with different content, they are not duplicate. OTOH /blog and /blog?p=1 are same content and can be linked with canonical. idk what is the downside of putting obviously wrong canonical but iirc when pages don't have duplicate content as seen by crawler then canonical tag may ignored altogether.

cuihtlauac:
Regarding search results, here is a quick comparison, not really conclusive. Looks like a matter of taste :-)

1. Android & Epic Games - Not duplicates with the same canonical, parameters removed

https://developer.android.com/s/results?q=activity
https://developer.android.com/s/results?q=telephony 
https://developer.android.com/s/results?q=compose

Same canonical:

https://developer.android.com/s/results

https://dev.epicgames.com/en-US/search?filter=Documentation&keyword=leaderboard
https://dev.epicgames.com/en-US/search?filter=Documentation&keyword=auth

Same canonical:

https://dev.epicgames.com/en-US/search

2. AWS - No canonicals

https://docs.aws.amazon.com/search/doc-search.html?searchPath=documentation-guide&searchQuery=UpdateClientCertificate

3. Apple & Facebook - Canonicals with parameters

Search:    https://developers.facebook.com/search/?q=scroll
Canonical: https://developers.facebook.com/search/?q=scroll

Search:    https://developer.apple.com/search/?q=compose
Canonical: https://developer.apple.com/search/?q=compose

cuihtlauac:
RFC6596 seems to indicate Facebook and Apple are the winners. 

https://www.ietf.org/rfc/rfc6596.txt

>  The target (canonical) IRI MUST identify content that is either duplicative or a superset of the content at the context (referring) IRI.

In the case of Anroid and Epic Games, the target is neither duplicate nor superset.

tmattio:
Thanks @cuihtlauac!

PR #598, opened 2022-10-08
Update homepage for OCaml 5 Beta release

### LINKED ISSUE #590: Information about OCaml 5
Currently ocaml.org does not talk about the next major release of the OCaml programming language, OCaml 5. Given that alpha releases are out, it may be useful to have it featured prominently so as to encourage early adopters to try it out. 

We may consider including a "Try OCaml 5.0" banner that links to the current, up to date installation instructions. 

This was brought up originally in https://discuss.ocaml.org/t/multicore-with-opam-instructions/10548/7

tmattio:
In preparation for the upcoming release of OCaml 5 Beta, this adds a banner to announce the release and redirect to the Up and Running page.

- The Up and Running instructions are updated to create switches with `5.0.0~beta0`
- The code snippet on the homepage uses the parallel Fibonacci example from https://github.com/ocaml-multicore/parallel-programming-in-multicore-ocaml/blob/master/code/task/fibonacci_multicore.ml

image

Closes #590 
cc @kayceesrk @avsm @Octachron 

Note that the Up and Running guide is maybe not the best place to redirect users with instructions to try OCaml 5, since it contains other documentation like how to download and init opam. As an alternative, we could (1) create an announcement post that contains setup instructions, or (2) upload the release (manual and changelog with setup instructions) when the beta is out and overwrite it when the stable release is out. @dra27 and I thought the latter could be a good option.

Other open questions:

- Should we put the banner only on the homepage?
- Should we make the cross button functional (this would require adding a cookie) or be removed
- Is a banner the best way to promote the release? We could also update the hero to add a CTA with the same link


kayceesrk:
The code in the screenshot is different from the linked one. The screenshot code uses domains directly whereas the linked code uses domains lib. 

It is also the case that the code spawns more than 8 domains? Both the recursive calls get passed with `d - 1`. Would showing the example with domainslib be too long? If so, what about a simple eio example?

avsm:
I don't think we should direct users to a beta release in the up-and-coming instructions. A separate callout would be better.

Incidentally, I tend to use

```
opam switch create 5-latest --packages ocaml.5.0.0
```

in order to avoid hardcoding the `~release` as that also gets rapidly out of date.

data/tutorials/gs_00_up_and_running.md
```diff
@@ -115,21 +115,21 @@ Now check the installation by running `opam --version`. You can compare it with
 
 #### Creating a New Switch on Unix
 
-If you want a specific version of OCaml, or a new independent environment, you can create a new switch with the `opam switch create` command. Specify which version as shown below (i.e., `opam switch create 4.14.0`). All possible compiler versions can be found with `opam switch list-available`. The most current version can be found at [opam.ocaml.org](https://opam.ocaml.org/packages/ocaml-base-compiler/).
+If you want a specific version of OCaml, or a new independent environment, you can create a new switch with the `opam switch create` command. Specify which version as shown below (i.e., `opam switch create 5.0.0~beta0`). All possible compiler versions can be found with `opam switch list-available`. The most current version can be found at [opam.ocaml.org](https://opam.ocaml.org/packages/ocaml-base-compiler/).
```
dra27: Agree wholeheartedly with @avsm's suggestion (in general, `opam switch create <version>` is a hangover-hack from OPAM 1.x days; it's better in docs and examples always to specify `ocaml.<version>` since `opam switch create 5.0.0` will fail if you've installed OCaml 5.0.0 via your OS package manager.
```suggestion
If you want a specific version of OCaml, or a new independent environment, you can create a new switch with the `opam switch create` command. Specify which version as shown below (i.e., `opam switch create ocaml-5.0 ocaml.5.0.0`). All possible compiler versions can be found with `opam switch list-available`. The most current version can be found at [opam.ocaml.org](https://opam.ocaml.org/packages/ocaml-base-compiler/).
```
(same edits then on all the hunks below, obviously)

src/ocamlorg_frontend/layouts/layout.eml
```diff
@@ -39,6 +39,32 @@ let render ?(use_swiper=false) ?(wide=false) ?description ?styles ~title ?canoni
   </head>
 
   <body class="light">
+    <div class="relative bg-primary-600">
+      <div class="mx-auto max-w-7xl py-3 px-3 sm:px-6 lg:px-8">
+        <div class="pr-16 sm:px-16 sm:text-center">
+          <p class="font-medium text-white">
+            <span class="md:hidden">OCaml 5 Beta is out!</span>
+            <span class="hidden md:inline">OCaml 5 Beta is out! OCaml 5 is the next major release of OCaml, which comes with support for Multicore!</span>
```
dra27: Bike-shedding the text slightly (text taken from the multicore PR):
```suggestion
            <span class="md:hidden">OCaml 5.0 is in beta!</span>
            <span class="hidden md:inline">OCaml 5.0 is beta! OCaml 5.0 is the next major release of OCaml, which comes with support for shared-memory parallelism through domains and direct-style concurrency through algebraic effects!</span>
```

Octachron:
I would propose to have an Domainslib parallel loop as example rather than yet another Fibonacci's function with a double exponential slowdown. A parallel for (a vector addition maybe) seems to me both more recognizable and less subject to caveat.

src/ocamlorg_frontend/pages/home.eml
```diff
@@ -36,17 +36,20 @@ Layout.render
...
+</span><span class="ocaml-source">
+</span><span class="ocaml-keyword-other">#</span><span class="ocaml-source"> </span><span class="ocaml-constant-language-capital-identifier">Printf</span><span class="ocaml-keyword-other-ocaml punctuation-other-period punctuation-separator">.</span><span class="ocaml-source">printf</span><span class="ocaml-source"> </span><span class="ocaml-string-quoted-double">"</span><span class="ocaml-string-quoted-double">fib(</span><span class="ocaml-constant-character-printf">%d</span><span class="ocaml-string-quoted-double">) = </span><span class="ocaml-constant-character-printf">%d</span><span class="ocaml-constant-character-escape">\n</span><span class="ocaml-string-quoted-double">"</span><span class="ocaml-source"> </span><span class="ocaml-constant-numeric-decimal-integer">40</span><span class="ocaml-source"> </span><span class="ocaml-source">(</span><span class="ocaml-source">fib_par</span><span class="ocaml-source"> </span><span class="ocaml-constant-numeric-decimal-integer">40</span><span class="ocaml-source"> </span><span class="ocaml-constant-numeric-decimal-integer">8</span><span class="ocaml-source">)</span><span class="ocaml-source">
```
dra27: We should avoid hard-coding numbers of domains:
```suggestion
</span><span class="ocaml-keyword-other">#</span><span class="ocaml-source"> </span><span class="ocaml-constant-language-capital-identifier">Printf</span><span class="ocaml-keyword-other-ocaml punctuation-other-period punctuation-separator">.</span><span class="ocaml-source">printf</span><span class="ocaml-source"> </span><span class="ocaml-string-quoted-double">"</span><span class="ocaml-string-quoted-double">fib(</span><span class="ocaml-constant-character-printf">%d</span><span class="ocaml-string-quoted-double">) = </span><span class="ocaml-constant-character-printf">%d</span><span class="ocaml-constant-character-escape">\n</span><span class="ocaml-string-quoted-double">"</span><span class="ocaml-source"> </span><span class="ocaml-constant-numeric-decimal-integer">40</span><span class="ocaml-source"> </span><span class="ocaml-source">(</span><span class="ocaml-source">fib_par</span><span class="ocaml-source"> </span><span class="ocaml-constant-numeric-decimal-integer">40</span><span class="ocaml-source"> </span><span class="ocaml-constant-numeric-decimal-integer">(Domain.recommended_domain_count ())</span><span class="ocaml-source">)</span><span class="ocaml-source">
```

Octachron:
Also the first beta will be labelled `beta1`: the `alpha0` was supposed to be an exception and it leads to all kind of ordinal troubles: is `beta0` the first beta release? the zeroth? What about `beta1`, would that be the first or second `beta` release?

tmattio:
I'm not entirely sure about using Domainslib in the code snippet.

On the one hand, I understand that we want to recommend users to use it instead of using the Domain API directly. So the code would be closer to what users would to see "in the wild".

On the other hand, the code snippet would then require installing `domainslib` to be run. Users who visit the site, copy the code and try it on a project or a REPL would get an error. And if they are new users, they will probably not know what to do with the cryptic `Unbound module Domainslib` error. That's not a great first-time experience.

Octachron:
I think it is probably better to add an instruction for installing `domainslib` in the OCaml 5 beta guide (and mention the possible error in such guide).

tmattio:
Do you mean that it would be acceptable to use Domainslib in the code snippet if we added instructions on how to install the opam package in the OCaml 5 beta announcement?

Octachron:
Yes, taking in account that installing the beta is not something that I expect new users to do without following some installation instructions, I think it would be fine to have some pointer for installing domainslib in those instructions.

dra27:
The playground presumably can have `domainslib`?

tmattio:
> Yes, taking in account that installing the beta is not something that I expect new users to do without following some installation instructions, I think it would be fine to have some pointer for installing domainslib in those instructions.

True, although I could imagine someone landing on the homepage, going to Get Started and installing OCaml 4.14, then trying to run the snippet.

Reading the announcement, installing OCaml 5 beta and trying to run the snippet is one path the user could take, but there are many more they could take which would ultimately lead to a confusing error if we use something else than the standard library in the code snippet.

Saying that I realize I'm arguing against the code I put in the PR: users would get an error because of the Domain module if they try to run the code after following the Get Started guide (which installs OCaml 4.14).

I guess if we really want to show code that uses Eio/Domainslib I could try and change the code snippet to tell users they need to follow a certain guide to be able to run the code, but I have a preference for a code snippet they would be able to run no matter the environment.

tmattio:
> The playground presumably can have domainslib?

Wouldn't that depend on effects? (which the playground won't support since it's compiled with Jsoo)

Octachron:
I think that there is some unavoidable tension here between having a showcase snippet that works anywhere and a representative snippet that looks like some code that might be used in OCaml 5.

But maybe this is a sign that we don't want to switch the focus of the welcome code snippet? Looking at the current snippet, neither the `square` function nor the `factorial` function would benefit from using `Domain` for instance.

dra27:
> > The playground presumably can have domainslib?
> 
> Wouldn't that depend on effects? (which the playground won't support since it's compiled with Jsoo)

Oops - I completely forgot about Domainslib's scheduler using effects 🤦

kayceesrk:
Looks like there are several good reasons not to include domainslib. If the playground is compiled with jsoo then even plain domain won't work, would it? JavaScript doesn't have parallelism.

tmattio:
I removed the parts that needed more discussion and only kept the banner for now, which links to the announcement post.

I'm happy to revisit the code snippet in a second time.

src/ocamlorg_frontend/layouts/layout.eml
```diff
@@ -39,6 +39,25 @@ let render ?(use_swiper=false) ?(wide=false) ?description ?styles ~title ?canoni
   </head>
 
   <body class="light">
+    <% if banner then ( %>
+    <div class="relative bg-primary-600">
+      <div class="mx-auto max-w-7xl py-3 px-3 sm:px-6 lg:px-8">
+        <div class="pr-16 sm:px-16 sm:text-center">
+          <p class="font-medium text-white">
+            <span class="md:hidden">OCaml 5.0 is in beta!</span>
+            <span class="hidden md:inline">OCaml 5.0 is beta! OCaml 5.0 is the next major release of OCaml, which comes with support for shared-memory parallelism through domains and direct-style concurrency through algebraic effects!</span>
```
kit-ty-kate, line 48:
```suggestion
            <span class="hidden md:inline">OCaml 5.0 is in beta! It is the next major release of OCaml, which comes with support for shared-memory parallelism through domains and direct-style concurrency through algebraic effects!</span>
```

PR #603, opened 2022-10-13
Make links clickable and line-editing

tmattio:
@Octachron this makes the links clickable in [the rendered version of the announcement](https://ocaml.org/news/ocaml-5.0.beta1), let me know if that looks ok to you!

### DIFF
--- a/data/news/platform/ocaml-5.0.beta1.md
+++ b/data/news/platform/ocaml-5.0.beta1.md
@@ -5,15 +5,12 @@ date: "2022-10-12"
 tags: [ocaml, platform]
 ---
 
-
 The release of OCaml 5.0.0 is drawing near.
 The standard library has been stabilized and many opam packages already
 work with this release.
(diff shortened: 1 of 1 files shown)


data/news/platform/ocaml-5.0.beta1.md
```diff
@@ -92,9 +81,7 @@ In both cases, all available options can be listed with `opam search ocaml-optio
 ## Optional opam Alpha Repository
 
 During the beta release, if your dependencies are not yet compatible with OCaml 5.0.0, 
-you might want to check the alpha opam repository:
-
-https://github.com/kit-ty-kate/opam-alpha-repository
+you might want to check the [alpha opam repository](https://github.com/kit-ty-kate/opam-alpha-repository).
```
Octachron: It is not that useful to make this specific link clickable since I expect most people to only use this url through opam.

Octachron (approved):
The changes look good to me (except a minor remark that can be ignored).

tmattio:
Thanks! I've extracted the link to the alpha repository. It's still clickable, but now people can copy it as well to use it through opam.

PR #605, opened 2022-10-15
Fix editor slider order for Vim and Emacs

MQ37:
There is wrong order of editor previews for Vim and Emacs on the ocaml.org homepage. When you click Vim slider icon it shows Emacs preview and vice versa.

Fixed that by changing order of editor previews to match the order of icons (VSCode, Vim, Emacs).


cuihtlauac:
Thanks, that was not obvious to see.

PR #609, opened 2022-10-18
Rebase staging

cuihtlauac:
It's been a while since staging wasn't updated. Before pushing anything there, we probably need to be in sync with main. The only commit found in staging which wasn't in main was `Add governance section` from Thibaut Mattio, Asaad Mahmood and MirzaBabarBaig. It was bringed up.


tmattio (approved):

tmattio:
Thanks @cuihtlauac! I rebased manually and pushed to `staging`

PR #616, opened 2022-10-27
updating OCamlformat version in the intro

ac27182:


cuihtlauac:
Thanks for the suggestion. I agree leaving old versions' numbers is documentation should be avoided. However, here, I believe removing the number altogether makes more sense.

tmattio (approved):

PR #626, opened 2022-11-15
Fix executable path in helloworld project

DrTobe:
I have just tried the "Get Up and Running" tutorial part and I think the executable path is wrong


cuihtlauac:
Hi @DrTobe, thanks for your feedback.

The `dune exec ./bin/main.exe` command is a pretty counterintuitive, it should be said explicitly in the tutorial. Actually `./bin/main.exe` is *not* a file path! It means “run the executable defined in `main.ml`”

DrTobe:
Oh yes, that works, too. Sorry.

Issue #620, opened 2022-11-02
Turn headings into links


cuihtlauac:
Turning headings into links, as Rust documentation does:

https://doc.rust-lang.org/std/

allows easy sharing/bookmarking of sections.

Thanks to @d4h0 for initially reporting this on https://github.com/ocaml/ocaml.org/issues/618. This issue deals with the second half of it.

cuihtlauac:
PR: https://github.com/ocaml/ocaml.org/pull/628

PR #628, opened 2022-11-15
Add self-referencing links to headings

cuihtlauac:
Markddown header are turned into headers with links refering to
themselves. This addresses issue:
https://github.com/ocaml/ocaml.org/issues/620

### DIFF
--- a/src/ocamlorg_frontend/css/styles.css
+++ b/src/ocamlorg_frontend/css/styles.css
@@ -106,10 +106,33 @@ body {
   @apply text-primary-600 no-underline;
 }
 
+.prose [id] {
+  scroll-margin-top: 2rem;
+}
+
 .prose :is(h1, h2, h3, h4, h5, h6) > a,
-.prose :is(h1, h2, h3, h4, h5, h6) > a code {
+.prose :is(h1, h2, h3, h4, h5, h6) > a code,
+.prose :is(h1, h2, h3, h4, h5, h6) > a:hover,
+.prose :is(h1, h2, h3, h4, h5, h6) > a:hover code {
   color: unset;
-  font-weight: unset;
+  @apply no-underline;
+}
+
+.prose :is(h1:target, h2:target, h3:target, h4:target, h5:target, h6:target) {
+  position:relative;
+  background:rgb(255, 243, 173);
+  padding: 0.2rem 0rem;
+}
+
+.prose :is(h1, h2, h3, h4, h5, h6) > a:hover::before,
+.prose :is(h1, h2, h3, h4, h5, h6) > a:hover code::before {
+  content: "#";
+  position: absolute;
+  color: rgb(90, 98, 111);
+  margin-left: -1.2em;
+  padding: 0.39em;
+  margin-top: -.37em;
+  font-weight: 500;
 }
 
 .prose a:hover {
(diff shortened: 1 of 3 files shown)


tmattio:
This is changing the style of the documentation titles

Before
image

After
image

I agree having anchors on the titles is useful, but the current style looks better

cuihtlauac:
@Zineb-Ada has a solution to avoid inline styling inside tutorials.ml

tmattio:
I'm a bit concerned that we're adding a bit too many hacks just to get links for titles in the doc. Before we continue on this, are there other solutions that would fix the issue? E.g. can we have anchors "only" as seen on the tailwind doc for instance https://tailwindcss.com/docs/using-with-preprocessors.

Zineb-Ada:
good idea, i can try to do it and see if it will be more effective.

cuihtlauac:
Thanks to @Zineb-Ada we managed to restrict the changes to only 2 files:
* `src/ocamlorg_frontend/css/styles.css`
* `tool/ood-gen/lib/tutorial.ml`

I believe the hack factor remains acceptable :-)

From my understanding, having an anchor inside the headings seems to be the standard approach to allow users to access a section URL when reading the text. For instance, it's the way it's done in Tailwind CSS, Rust and Mozilla Developer Network:
* https://tailwindcss.com/docs/customizing-colors#arbitrary-values
* https://developer.mozilla.org/docs/Learn/Tools_and_testing#modules
* https://doc.rust-lang.org/std/boxed/index.html#memory-layout

However, w.r.t. styling, I've observed different approaches. I picked the Mozilla/Rust one. HTML changes are streamlined to the bare minimum, no class attribute is inserted (tailwindcss.com does that). It minimizes the diff size on `tutorial.ml`. We've failed to do styling in `tailwind.config.js`. We had to tweak `styles.css`. However, although the rules look scary to me, they say: “style anchor as heading, when occurring inside them”. Headings are styled the same as before, except they are underlined when hovering them.

sabine:
The other option is to do what odoc does: emit an anchor next to the heading text (see https://ocaml.org/p/irmin/3.4.2/doc/Irmin/index.html#preliminaries). I have styled this anchor with position:absolute to cover the whole heading element here: https://github.com/ocaml/ocaml.org/blob/main/asset/css/doc.css#L25-L49

However, I did it this way because I had no option to change the markup. It's just as hacky as all the other solutions.

tmattio:
Thanks @Zineb-Ada and @cuihtlauac! The implementation looks ok to me 🙂

I would try to make the behaviours of the doc and the package doc the same, I noticed a couple of differences:
- When hovering over the titles, it underlines it in the doc, not in the package doc. We can unset the underlying behaviour I think.
- When clicking on the title, it highlights it in yellow in the package doc, not in the doc. We can highlight it in the doc as well.

We could also reuse the trick @sabine used in https://github.com/ocaml/ocaml.org/pull/639: `scroll-margin-top: 2rem;`.
And we could look into adding anchors, similar to the package doc.

A bunch of minor changes, but I think it would be good to have the same UX on the documentation and package docs.

sabine:
Note: the OCaml language manual does the same as odoc, producing heading elements that contain an anchor: `<h1 id="anchor"><a href="#anchor"></a>...</h1>`.

So if we can generate the same markup, we can make make CSS rules that cover all these cases consistently.

PR #633, opened 2022-11-22
Add text covering the option type

### LINKED ISSUE #507: data-types not covered with Optional type
https://ocaml.org/docs/data-types

For real world OCaml code, we often using `option`

like example

```ocaml
utop # type netrcItem = {host: string option;
username: string option;
password: string option;
};;
type netrcItem = {
  host : string option;
  username : string option;
  password : string option;
}
utop # let a : netrcItem = {host = Some "localhost"; username = None; password = None;};;
val a : netrcItem = {host = Some "localhost"; username = None; password = None}
```

Shall we please improve the document of the  data types, to show the power of OCaml `option` type?

OCaml has much more power, but document never told us. sigh..

cuihtlauac:
* Show definition of the type
* Provide an example: Sys.getenv_opt
* Explain map and join
* Explain fold and unfold

Fix: https://github.com/ocaml/ocaml.org/issues/507

### DIFF
--- a/data/tutorials/lg_00_data_types_and_matching.md
+++ b/data/tutorials/lg_00_data_types_and_matching.md
@@ -301,8 +301,8 @@ a list of pairs to or from a tree dictionary and so on.
 
 ## Example: options
 
-The OCaml language has been supporting the [option
-type](https://en.wikipedia.org/wiki/Option_type) since its inception.
+A simple, yet very common, use of polymorphic type with constructors is the
+`option` type. Like `list`, it's predefined. Here is how:
 
 <!-- $MDX non-deterministic=command -->
 ```ocaml
@@ -310,28 +310,49 @@ type](https://en.wikipedia.org/wiki/Option_type) since its inception.
 type 'a option = None | Some of 'a
 ```
 
-A value of type `t option` for some type `t` represent either a value `v` of
-type `t`, wraped as `Some v`, or no such value, in such case `o` has the value
-`None`. The option type is very useful when lack of data is better handled as a
-special value (_i.e._ `None`) rather than an exception. It is the type-safe
-version of returning error values such as in C, for instance.
+`option` used to wrap some data, if available, or absence of data otherwise.
+Here is 42, stored inside an `option` using the data carrying constructor
+`Some`:
 
-The function `Sys.getenv_opt : string -> string opt` from the standard library
-allows to query the value of an environment variable from the running process.
-Here is what may happen on a machine where a variable called `EDITOR` is
-defined:
+```ocaml
+# Some 42;;
+- : int option = Some 42
+```
+
+The `None` constructor means no data is availble.
+
+In other words a value of type `t option` for some type `t` represents:
+
+* either a value `v` of type `t`, wrapped as `Some v`,
+* no such value, then `o` has the value `None`.
+
+The option type is very useful when lack of data is better handled as a special
+value (_i.e._ `None`) rather than an exception. It is the type-safe version of
+returning error values such as in C, for instance. Since no data has any special
+meaning, confusion between regular values and absence of value is impossible. In
+computer science, this type is called the [option
+type](https://en.wikipedia.org/wiki/Option_type). OCaml has supported `option`
+since its inception.
+
+The function `Sys.getenv : string -> string` from the standard library allows to
+query the value of an environment variable, however the it throws an exception
+if the variable is not defined. On the other hand, the function `Sys.getenv_opt
+: string -> string opt` does the same except it returns `None` is the variable
+is not defined. Here is what may happen on a machine where a variable called
+`EDITOR` is defined:
 
 <!-- $MDX non-deterministic=output -->
 ```ocaml
-# Sys.getenv_opt "EDITOR";;
-- : string option = Some "vi"
+# Sys.getenv_opt "DUMMY";;
+Exception: Not_found.
+# Sys.getenv_opt "DUMMY";;
+- : string option = None
 ```
 
-However if no such variable is defined the returned value will be `None`.
-
 Using pattern-matching, it is possible to define functions allowing to easily
-work with option values. Here is `map` which allows to apply a function to the
-value wrapped inside an option, if present:
+work with option values. Here is `map` of type `('a -> 'b) -> 'a option -> 'b
+option`. It allows to apply a function to the value wrapped inside an `option`,
+if present:
 
 ```ocaml
 # let map f = function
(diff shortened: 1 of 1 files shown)


data/tutorials/lg_00_data_types_and_matching.md
```diff
@@ -299,6 +299,76 @@ val insert : 'a * 'b -> ('a * 'b) tree -> ('a * 'b) tree = <fun>
 Similar functions can be written to look up values in a dictionary, to convert
 a list of pairs to or from a tree dictionary and so on.
 
+## Example: options
+
+The OCaml language has been supporting the [option
+type](https://en.wikipedia.org/wiki/Option_type) since its inception.
```
tmattio: The Wikipedia article reads:

> In [programming languages](https://en.wikipedia.org/wiki/Programming_language) (especially [functional programming](https://en.wikipedia.org/wiki/Functional_programming) languages) and [type theory](https://en.wikipedia.org/wiki/Type_theory), an option type or maybe type is a [polymorphic type](https://en.wikipedia.org/wiki/Parametric_polymorphism) that represents encapsulation of an optional value; e.g., it is used as the return type of functions which may or may not return a meaningful value when they are applied. It consists of a constructor which either is empty (often named None or Nothing), or which encapsulates the original data type A (often written Just A or Some A).

I was expecting something much more complicated, but still, I'm not sure it's the most beginner-friendly explanation we could give. E.g. this assumes that the reader knows what encapsulation means (and coming from OOP, what one understands by "encapsulation" can be something completely different!). Same for polymorphic types. I'd recommend writing a friendlier introduction of the option type in the doc.

data/tutorials/lg_00_data_types_and_matching.md
```diff
@@ -299,6 +299,76 @@ val insert : 'a * 'b -> ('a * 'b) tree -> ('a * 'b) tree = <fun>
...
+value wrapped inside an option, if present:
+
+```ocaml
+# let map f = function
+  | None -> None
+  | Some v -> Some (f v);;
+val map : ('a -> 'b) -> 'a option -> 'b option = <fun>
+```
+
+Here is `join`, which peels off one layer from a doubly wrapped option:
```
tmattio: I found that using the types when explaining functions can aid the reader/learner's understanding. For instance, the above can seem a bit abstract, but reading `'a option option -> 'a option` at the same time might help.

data/tutorials/lg_00_data_types_and_matching.md
```diff
@@ -299,6 +299,76 @@ val insert : 'a * 'b -> ('a * 'b) tree -> ('a * 'b) tree = <fun>
...
+    else
+      None;;
+val fold : ('a -> 'b) -> 'b -> 'a option -> 'b = <fun>
+val unfold : ('a -> bool) -> ('a -> 'b) -> 'a -> 'b option = <fun>
+```
+
+Most of those functions as well as other useful ones are provided by the
+Standard Library in the [`Stdlib.Option`](https://ocaml.org/api/Option.html)
+supporting module.
+
```
tmattio, line 418: Could be a nice addition after these, to add something like "oh, by the way, a module that defines the functions above is called a monad, but we'll get back to it".

data/tutorials/lg_00_data_types_and_matching.md
```diff
@@ -299,6 +299,76 @@ val insert : 'a * 'b -> ('a * 'b) tree -> ('a * 'b) tree = <fun>
 Similar functions can be written to look up values in a dictionary, to convert
 a list of pairs to or from a tree dictionary and so on.
 
+## Example: options
```
christinerose:
```suggestion
## Example: Options
```

data/tutorials/lg_00_data_types_and_matching.md
```diff
@@ -299,6 +299,76 @@ val insert : 'a * 'b -> ('a * 'b) tree -> ('a * 'b) tree = <fun>
 Similar functions can be written to look up values in a dictionary, to convert
 a list of pairs to or from a tree dictionary and so on.
```
christinerose:
```suggestion
Similar functions can be written to look up values in a dictionary, to convert
a list of pairs to or from a tree dictionary, and so on.
```

data/tutorials/lg_00_data_types_and_matching.md
```diff
@@ -299,6 +299,76 @@ val insert : 'a * 'b -> ('a * 'b) tree -> ('a * 'b) tree = <fun>
...
+type](https://en.wikipedia.org/wiki/Option_type) since its inception.
+
+<!-- $MDX non-deterministic=command -->
+```ocaml
+# #show option;;
+type 'a option = None | Some of 'a
+```
+
+A value of type `t option` for some type `t` represent either a value `v` of
+type `t`, wraped as `Some v`, or no such value, in such case `o` has the value
```
christinerose:
```suggestion
type `t`, wrapped as `Some v`, or no such value, in such case `o` has the value
```

data/tutorials/lg_00_data_types_and_matching.md
```diff
@@ -299,6 +299,76 @@ val insert : 'a * 'b -> ('a * 'b) tree -> ('a * 'b) tree = <fun>
...
+`None`. The option type is very useful when lack of data is better handled as a
+special value (_i.e._ `None`) rather than an exception. It is the type-safe
```
christinerose:
```suggestion
special value (_i.e.,_ `None`) rather than an exception. It is the type-safe
```

data/tutorials/lg_00_data_types_and_matching.md
```diff
@@ -299,6 +299,76 @@ val insert : 'a * 'b -> ('a * 'b) tree -> ('a * 'b) tree = <fun>
...
+version of returning error values such as in C, for instance.
+
+The function `Sys.getenv_opt : string -> string opt` from the standard library
+allows to query the value of an environment variable from the running process.
+Here is what may happen on a machine where a variable called `EDITOR` is
+defined:
```
christinerose:
```suggestion
The function `Sys.getenv_opt : string -> string opt` from the standard library
allows users to query the value of an environment variable from the running process.
Here is what may happen on a machine where a variable called `EDITOR` is
defined:
```

data/tutorials/lg_00_data_types_and_matching.md
```diff
@@ -299,6 +299,76 @@ val insert : 'a * 'b -> ('a * 'b) tree -> ('a * 'b) tree = <fun>
...
+
+<!-- $MDX non-deterministic=output -->
+```ocaml
+# Sys.getenv_opt "EDITOR";;
+- : string option = Some "vi"
+```
+
+However if no such variable is defined the returned value will be `None`.
```
christinerose:
```suggestion
However if no such variable is defined, the returned value will be `None`.
```

data/tutorials/lg_00_data_types_and_matching.md
```diff
@@ -299,6 +299,76 @@ val insert : 'a * 'b -> ('a * 'b) tree -> ('a * 'b) tree = <fun>
...
+
+Using pattern-matching, it is possible to define functions allowing to easily
```
christinerose:
```suggestion
Using pattern-matching, it is possible to define functions, allowing users to easily
```

data/tutorials/lg_00_data_types_and_matching.md
```diff
@@ -299,6 +299,76 @@ val insert : 'a * 'b -> ('a * 'b) tree -> ('a * 'b) tree = <fun>
...
+work with option values. Here is `map` which allows to apply a function to the
```
christinerose:
```suggestion
work with option values. Here is `map` which allows the application of a function to the
```

data/tutorials/lg_00_data_types_and_matching.md
```diff
@@ -299,6 +299,76 @@ val insert : 'a * 'b -> ('a * 'b) tree -> ('a * 'b) tree = <fun>
...
+Here is `join`, which peels off one layer from a doubly wrapped option:
+
+```ocaml
+# let join = function
+  | Some o -> o
+  | None -> None;;
+val join : 'a option option -> 'a option = <fun>
+```
+
+Here are generic functions allowing to consume and produce option values. `fold`
```
christinerose:
```suggestion
Here are generic functions allowing users to consume and produce option values. `fold`
```

data/tutorials/lg_00_data_types_and_matching.md
```diff
@@ -299,6 +299,76 @@ val insert : 'a * 'b -> ('a * 'b) tree -> ('a * 'b) tree = <fun>
...
+    if p x then
+      Some (f x)
+    else
+      None;;
+val fold : ('a -> 'b) -> 'b -> 'a option -> 'b = <fun>
+val unfold : ('a -> bool) -> ('a -> 'b) -> 'a -> 'b option = <fun>
+```
+
+Most of those functions as well as other useful ones are provided by the
+Standard Library in the [`Stdlib.Option`](https://ocaml.org/api/Option.html)
```
christinerose: You have "Standard Library" capitalised here but not earlier in the doc. Which is correct?

christinerose (changes requested):

data/tutorials/lg_00_data_types_and_matching.md
```diff
@@ -299,6 +299,128 @@ val insert : 'a * 'b -> ('a * 'b) tree -> ('a * 'b) tree = <fun>
...
+if the variable is not defined. On the other hand, the function `Sys.getenv_opt
+: string -> string opt` does the same except it returns `None` is the variable
+is not defined. Here is what may happen on a machine where a variable called
+`EDITOR` is defined:
+
+```ocaml
+# Sys.getenv "DUMMY";;
+Exception: Not_found.
+# Sys.getenv_opt "DUMMY";;
+- : string option = None
```
tmattio:
```suggestion
# Sys.getenv "UNDEFINED_ENVIRONMENT_VARIABLE";;
Exception: Not_found.
# Sys.getenv_opt "UNDEFINED_ENVIRONMENT_VARIABLE";;
- : string option = None
```

data/tutorials/lg_00_data_types_and_matching.md
```diff
@@ -299,6 +299,128 @@ val insert : 'a * 'b -> ('a * 'b) tree -> ('a * 'b) tree = <fun>
...
+The option type is very useful when lack of data is better handled as a special
+value (_i.e._ `None`) rather than an exception. It is the type-safe version of
+returning error values such as in C, for instance. Since no data has any special
+meaning, confusion between regular values and absence of value is impossible. In
+computer science, this type is called the [option
+type](https://en.wikipedia.org/wiki/Option_type). OCaml has supported `option`
+since its inception.
+
+The function `Sys.getenv : string -> string` from the standard library allows to
+query the value of an environment variable, however the it throws an exception
```
tmattio:
```suggestion
query the value of an environment variable, however, it throws an exception
```

data/tutorials/lg_00_data_types_and_matching.md
```diff
@@ -299,6 +299,128 @@ val insert : 'a * 'b -> ('a * 'b) tree -> ('a * 'b) tree = <fun>
...
+if the variable is not defined. On the other hand, the function `Sys.getenv_opt
+: string -> string opt` does the same except it returns `None` is the variable
+is not defined. Here is what may happen on a machine where a variable called
+`EDITOR` is defined:
```
tmattio:
```suggestion
is not defined. Here is what may happen if we try to access an undefined environment variable:
```

data/tutorials/lg_00_data_types_and_matching.md
cuihtlauac: Nope :-) It should be all the same way. I'll fix that.

tmattio (approved):

PR #637, opened 2022-11-25
Fix broken anchor links in tutorials

cuihtlauac:
Used git grep for anchor links. Used some basic text processing to
generate url and test them. Had to rename tt_00


tmattio (approved):
Thanks! I tested the changes and didn't notice any issue.

PR #640, opened 2022-11-28
collapse table of contents to sidebar on smaller-than-xl screens

sabine:
On lg-screen, collapse the table of contents right sidebar and place table of contents in the left sidebar. This ensures that the content area is wide enough on all screen widths.

* includes minor style adjustments to ToC (spacing, font weight, hover color)
* left sidebar and content area now have the same padding-top as the right sidebar which means the alignment is no longer off

| screen| before | after |
|---|---|---|
| xl | Screenshot 2022-11-28 at 16-14-45 Irmin · irmin 3 4 3 · OCaml Packages | Screenshot 2022-11-28 at 16-14-39 Irmin · irmin 3 4 3 · OCaml Packages | 
|lg | Screenshot 2022-11-28 at 16-15-01 Irmin · irmin 3 4 3 · OCaml Packages | Screenshot 2022-11-28 at 16-15-06 Irmin · irmin 3 4 3 · OCaml Packages |

### DIFF
--- a/asset/css/doc.css
+++ b/asset/css/doc.css
@@ -1,6 +1,4 @@
 div.odoc {
-  margin-left: auto;
-  margin-right: auto;
   max-width: 56rem /* 896px */;
   position: relative;
 }
--- a/src/ocamlorg_frontend/components/toc.eml
+++ b/src/ocamlorg_frontend/components/toc.eml
@@ -9,28 +9,27 @@ type t = toc list
 let render (t : t) =
   <div>
     <% (match t with [] ->  %>
-    <span class="block py-2 text-gray-900">No table of content</span>
+    <span class="block py-2 text-gray-900">No table of contents</span>
     <% | _ -> %>
-    <ul class="text-gray-700 text-sm leading-6">
+    <ul class="leading-6 space-y-3 text-sm">
       <% t |> List.iter begin fun item -> %>
         <li>
-          <a href="<%s item.href %>" class="block py-1 font-medium text-gray-900 transition-colors duration-300">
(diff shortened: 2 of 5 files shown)


tmattio:
Do we want to keep the "ON THIS PAGE" section title when the menu is on the left?

sabine:
I would argue no, because on the left side, vertical space is very expensive. On the right hand side, we can keep it because we have plenty of vertical space available.

tmattio:
As a counterpoint, I found the list of links without context to be a bit confusing. On staging, I initially thought these were the list of "top-level pages", until I clicked and was redirected on a section of the same page. So now I know, but without context, I didn't know what these links were.

sabine:
It's a tradeoff. The majority of users will quickly be familiar by muscle-memory and benefit from seeing more items on the sidebar for packages with a lot of content. The mitigating factor here is the good discoverability: The scroll animation makes it obvious what the links do.

It's not a hill I must die on. :shrug:  My hypothesis here is: on a page that I estimate to be used by 95%+ repeat users, using this space to display content is more useful than spending this space on (maybe) making the TOC easier to recognize for first time users.

We could keep "ON THIS PAGE" and compact the layout further when we later introduce a switch between "manual" and "api" to the sidebar.
Or, alternatively, place the whole TOC including "ON THIS PAGE" on md-size and lg-size screens between odoc's preamble and content block (Wikipedia-style), when we bring back the missing preambles.

tmattio (approved):
Let's merge and revisit after we've received user feedback. In any case the current PR is an improvement over what we have.

PR #641, opened 2022-11-28
Package Explorer in Sidebar and Breadcrumbs

sabine:
This patch applies some visual and behavioral changes to the "package explorer" in the left sidebar of the package documentation area.

* more compact layout of the package explorer, inspired by file tree views of code editors
* visual highlighting of current entry in the package explorer
* colorful icons on the package explorer facilitate quicker recognition of page type (module, module type, parameter, class, class type)
* breadcrumbs updated to have better visual hierarchy (include package and library) using the color symbolism of the package explorer

| screen | before | after |
|-|-|-|
| sm + md | Screenshot 2022-11-29 at 12-53-52 Irmin Backend Conf · irmin 3 4 2 · OCaml Packages | Screenshot 2022-11-29 at 12-53-46 Irmin Backend Conf · irmin 3 4 2 · OCaml Packages |
| lg | Screenshot 2022-11-29 at 12-53-33 Irmin Backend Conf · irmin 3 4 2 · OCaml Packages | Screenshot 2022-11-29 at 13-04-55 Irmin Backend Conf · irmin 3 4 2 · OCaml Packages |
| xl | Screenshot 2022-11-29 at 12-53-24 Irmin Backend Conf · irmin 3 4 2 · OCaml Packages | Screenshot 2022-11-29 at 12-53-20 Irmin Backend Conf · irmin 3 4 2 · OCaml Packages |

### DIFF
--- a/asset/css/doc.css
+++ b/asset/css/doc.css
@@ -187,3 +187,128 @@ div.odoc ul.at-tags p {
 div.odoc span.at-tag {
   font-weight: bold;
 }
+
+/* package explorer navmap and breadcrumb tag */
+
+.navmap:hover ul {
+  border-color: gainsboro;
+}
+
+.no-expand, .icon-expand  {
+  white-space:nowrap;
+  display:flex;
+  flex-wrap: nowrap;
+}
+
+.icon-expand {
+  cursor: pointer;
+}
+
+span.icon-expand:not(.open) > .navmap-tag , span.no-expand > .navmap-tag {
+  background:white;
+}
+
+.navmap-tag {
+  display:flex;
+  align-items: center;
+  justify-content: center;
+  text-align: center;
+  width:2.1em;
+  min-width:2.1em;
+  margin:3px 0;
+  padding:2px 1px;
+  line-height:1;
+  margin-right:6px;
+  font-family: Inter, sans-serif;
+  border-width: 2px;
+  border-radius: 0.25em;
+  font-weight:bold;
+  font-size: .875rem;
+  box-sizing: border-box;
+}
+
+.navmap-tag.library-tag::after {
+  content: "lib";
+}
+.library-tag {
+  color: rgb(91, 102, 114);
+  background-color: rgb(91, 102, 114);
+  border-color: rgb(91, 102, 114);
+}
+.navmap-tag.module-tag::after {
+  content: "M";
+}
+.module-tag {
+  color: #cc4e0c;
+  background-color: #cc4e0c;
+  border-color: #cc4e0c;
+}
+.navmap-tag.module-type-tag::after {
+  content: "Mt";
+}
+.module-type-tag{
+  color: #027491;
+  background-color: #027491;
+  border-color: #027491;
+}
+.navmap-tag.parameter-tag::after {
+  content: "P";
+}
+.parameter-tag{
+  color: green;
+  background-color: green;
+  border-color: green;
+}
+.navmap-tag.class-tag::after {
+  content: "C";
+}
+.class-tag{
+  color: rgb(163, 34, 34);
+  background-color: rgb(163, 34, 34);
+  border-color: rgb(163, 34, 34);
+}
+.navmap-tag.class-type-tag::after {
+  content: "Ct";
+}
+.class-type-tag{
+  color: rgb(32, 68, 165);
+  background-color: rgb(32, 68, 165);
+  border-color: rgb(32, 68, 165);
+}
+
+span.no-expand > .navmap-tag {
+  margin-left: 12px;
+}
+
+span.icon-expand.open > .navmap-tag {
+  color:white;
+}
+
+span.icon-expand::before {
+  font-weight: bold;
+  content: " \25B8";
+  text-align: center;
+  box-sizing: border-box;
+  display:flex;
+  align-items: center;
+  padding-right:4px;
+}
+
+span.icon-expand.open::before {
+  content: " \25BE";
+}
+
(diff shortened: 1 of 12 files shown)


tmattio:
Thanks @sabine!

Just have a couple of questions/remarks on the UI:

- The "breadcrumb" above the title is a bit awkward. It only links to the package (which we're already on) and the real breadcrumb is the title. I don't really see a use case for this, can we remove it?
- Why replace "IN THIS PACKAGE" with "package <package name>"? I find the latter to be more confusing.
- Where did the TOC go on md screens?

Apart from these, it looks very good to me, it's a huge improvement over the current package explorer! 🙂

sabine:
> * The "breadcrumb" above the title is a bit awkward. It only links to the package (which we're already on) and the real breadcrumb is the title. I don't really see a use case for this, can we remove it?

When we remove the tabs, there is no way to get back to the package documentation index page anymore.

In UX testing, people remarked that they didn't see an obvious way to get back to the package documentation index page, even when we had the "Documentation" tab.

The breadcrumbs are one of the obvious places to look for a way back to the package documentation index page.

> * Why replace "IN THIS PACKAGE" with "package "? I find the latter to be more confusing.

"package X" acts as a link to go back to the index page of the package documentation. Having a link back to the package documentation index page in the sidebar is useful for people who arrive on the documentation page by a targeted link (and, thus, do not see the breadcrumbs unless they scroll up).

I figured, we can combine "IN THIS PACKAGE" and "package X" to take up less vertical space, so that we have more space available for the package explorer.

> * Where did the TOC go on md screens?

This particular page has no TOC. I chose it to show the highlighting on the package explorer tree.

Here is one that has a TOC on md screen:

| before | after|
|-|-|
| Screenshot 2022-11-29 at 14-14-15 Irmin · irmin 3 4 2 · OCaml Packages | Screenshot 2022-11-29 at 14-13-33 Irmin · irmin 3 4 2 · OCaml Packages |

tmattio:
> The breadcrumbs are one of the obvious places to look for a way back to the package documentation index page.

It wasn't obvious to me that this was a way to get back to the overview. Do you think having a link "<- back to the overview" somewhere would make sense?

--

Also one random question, have you thought about inlining the package explorer? That is, listing the modules one by one, like the Phoenix docs for instance: https://hexdocs.pm/phoenix/Phoenix.html

One obvious downside would be the size of the navbar on certain packages (Irmin, Base), but I suspect the vast majority of packages don't have so many modules. And I'm not sure having a long list of modules is much worse than a very deep package tree view where you have to click 4-5 times to get to some module. Curious to know what you think?

sabine:
> Do you think having a link "<- back to the overview" somewhere would make sense?

For the package documentation `index.html`, I think I can safely put that in the sidebar, above the TOC.

For the package overview... it's trickier. I could add a left-arrow in the package header (only when we're on the Documentation), to make it more obvious that clicking on the name of the package leads back to the overview page. :thinking:

What about this:

Screenshot 2022-11-30 at 15-04-54 GraphicsX11 · ocaml-base-compiler 4 07 1 · OCaml Packages
?

> have you thought about inlining the package explorer?

It makes sense to explore that in upcoming iterations. And indeed, there are packages that have a very long list of items, when we inline everything (but many of those already are very long when opening the nodes in question). This patch tries to stay close to the previous way the sidebar worked, while reducing the extent of existing problems.

My proposal for this: When this patch is merged, we can proceed to explore inlining on staging.

tmattio:
> For the package documentation index.html, I think I can safely put that in the sidebar, above the TOC.
> 
> For the package overview... it's trickier. I could add a left-arrow in the package header (only when we're on the Documentation), to make it more obvious that clicking on the name of the package leads back to the overview page. 🤔

Do we really need both? To me a link on the sidebar (above the TOC or package explorer) is sufficiently visible?

> My proposal for this: When this patch is merged, we can proceed to explore inlining on staging.

Sounds great, this patch is clearly a huge improvement, let's bring it to users as soon as possible and take the time to improve it further once it's merged!

sabine:
> Do we really need both?

They lead to different pages, one leads to Overview, the other leads to Documentation. Both are necessary.

tmattio:
I'm still not really convinced we need the arrow on the package name. Clicking on it redirects to the overview, which is enough I think.

For the "back to documentation root", this is the same link as the package one, so do we actually need it as well?

This is adding a lot of visual information and links that contribute to making the mental model of how to navigate in the package doc more complex (when the initial intent is to make it simpler!). To me, "click on the package name to go to overview" and "click on the top-level link of the package explorer to go back to the root of the package doc," is a simple model, that seems intuitive, and provides the same navigation capabilities.

I'm usually trying to be less pushy on UI changes, but in this specific case, I feel this is actually hurting the usability of the doc. Or is there something I didn't understand?

tmattio (approved):

sabine:
Thanks for looking at this with a critical eye, the feedback has been very helpful to simplify this.

I removed the spurious instances of "back to documentation root". Now, this link only appears when there is a TOC and we are on lg-size or smaller screens. Since the root of the package explorer is likely obvious enough when it is at the top of the sidebar, that seems a reasonable compromise. Let's treat it as an interim solution until we overhaul the package navigation/overview.

| lg | xl |
|-|-|
| Screenshot 2022-12-05 at 10-00-05 Graphics · ocaml-base-compiler 4 07 1 · OCaml Packages | Screenshot 2022-12-05 at 09-59-53 Graphics · ocaml-base-compiler 4 07 1 · OCaml Packages |

PR #642, opened 2022-11-28
more compact package layout header, removes package layout tabs

sabine:
Reduces the height of the package layout's header by applying a different layout for the package name, version and description.

Update: Tabs are removed in favor of a "Documentation" button below the package repository URL (or error in case the package documentation failed to build or its state is unknown). See large screen for the button, and small- and md-screen for the error box. Error box is also a link to the "Documentation" URL of the package, because, sometimes, there are error messages available under that URL.
| screen | before | after|
|-|-|-|
| small |Screenshot 2022-11-28 at 21-24-35 irmin 0 9 0 · OCaml Package | Screenshot 2022-11-28 at 21-24-39 irmin 0 9 0 · OCaml Package |
| md | Screenshot 2022-11-28 at 21-24-20 irmin 0 9 0 · OCaml Package | Screenshot 2022-11-28 at 21-24-24 irmin 0 9 0 · OCaml Package |
| lg | Screenshot 2022-11-28 at 21-26-07 irmin 3 4 2 · OCaml Package | Screenshot 2022-11-28 at 21-26-11 irmin 3 4 2 · OCaml Package |

### DIFF
--- a/src/ocamlorg_frontend/pages/package_overview.eml
+++ b/src/ocamlorg_frontend/pages/package_overview.eml
@@ -10,10 +10,8 @@ let render
 ~license_filename
 (package : Package_intf.package) =
 Package_layout.render
-~documentation_status
 ~title:(Printf.sprintf "%s %s · OCaml Package" package.name package.version)
 ~description:(Printf.sprintf "%s %s: %s" package.name package.version package.description)
-~tab:Overview
 ~canonical:(Url.package_with_version package.name package.version)
 ~package @@
 <div class="flex lg:space-x-4 xl:space-x-12 flex-col xl:flex-row justify-between">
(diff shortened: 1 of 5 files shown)


tmattio:
Thanks @sabine!

I have a preference for keeping the synopsis on a new line: it's a bit awkward to have the title - version select - synopsis on the same line.

However, I think we can remove the tabs. We can add a link to the documentation in the package metadata instead, and that should save *a lot* of vertical space.

What do you think?

tmattio:
FWIW, the reason I used tabs here was that we were going to work on package toplevels and I thought we could add them as another tab. But now it's not clear when we'll introduce package toplevels, or whether they will even be on a standalone page when we introduce them. So the tabs don't seem so useful anymore if it's only to switch from overview to documentation.

sabine:
Sounds like a good plan to me. Will ping after I update the PR to remove the tabs, keeping the package description on its own line.

sabine:
Updated to remove the tabs in favor of a documentation button.

src/ocamlorg_frontend/pages/package_overview.eml
```diff
@@ -66,6 +64,15 @@ Package_layout.render
...
+            Documentation status is unknown. Sorry about that.</a><% )
+            | `Failure -> ( %><a href="<%s Url.package_doc package.name package.version ~page:"index.html" %>" class="p-4 mt-4 flex gap-x-2 bg-error-600 text-white font-semibold text-base">
+            <svg xmlns="http://www.w3.org/2000/svg" height="24px" viewBox="0 0 24 24" width="24px" fill="white"><path d="M0 0h24v24H0V0z" fill="none"/><path d="M15.32 3H8.68c-.26 0-.52.11-.7.29L3.29 7.98c-.18.18-.29.44-.29.7v6.63c0 .27.11.52.29.71l4.68 4.68c.19.19.45.3.71.3h6.63c.27 0 .52-.11.71-.29l4.68-4.68c.19-.19.29-.44.29-.71V8.68c0-.27-.11-.52-.29-.71l-4.68-4.68c-.18-.18-.44-.29-.7-.29zM12 17.3c-.72 0-1.3-.58-1.3-1.3s.58-1.3 1.3-1.3 1.3.58 1.3 1.3-.58 1.3-1.3 1.3zm0-4.3c-.55 0-1-.45-1-1V8c0-.55.45-1 1-1s1 .45 1 1v4c0 .55-.45 1-1 1z"/></svg>
+            Documentation failed to build! Sorry about that.</a><% ));%>
```
cuihtlauac: I'd rather go for something like this:
```html
   <div class="mt-4">
      <% (match documentation_status with
         | `Success -> %>
         <a class="btn btn-secondary" href="<%s Url.package_doc package.name package.version ~page:"index.html" %>">Documentation</a>
      <% | `Unknown ->  %>
         <div class="btn btn-sm-disabled" href="<%s Url.package_doc package.name package.version ~page:"index.html" %>">Documentation status is unknown. Sorry about that.</div>
      <% | `Failure -> %>
         <div class="btn btn-sm-disabled" href="<%s Url.package_doc package.name package.version ~page:"index.html" %>">Documentation failed to build! Sorry about that.</div>
      <%  ); %>
    </div>
```
* Slightly more compact
* No broken links
* More consistent appearance (rounded corners)

It relies on some CSS tweaking
```diff
diff --git a/src/ocamlorg_frontend/css/partials/buttons.css b/src/ocamlorg_frontend/css/partials/buttons.css
index 707ba22..bc6b887 100644
--- a/src/ocamlorg_frontend/css/partials/buttons.css
+++ b/src/ocamlorg_frontend/css/partials/buttons.css
@@ -18,6 +18,14 @@
   @apply px-4 h-10 text-sm;
 }
 
+.btn-sm-disabled {
+  @apply bg-primary-600/50 px-4 h-10 text-sm;
+}
+
+.btn-sm-disabled:hover {
+  @apply bg-primary-600/50 px-4 h-10 text-sm;
+}
+
 .btn-secondary {
   @apply bg-transparent border-primary-600 text-primary-600 border;
 }
````

cuihtlauac: Did I tell anyone I'm a bit fed up with _orange_? I mean as a color of course :-)

tmattio:
Thanks, this looks very good. One minor comment: I find the alert to be *very* aggressive. Can we disable the button with a tooltip saying the doc isn't available instead?

cuihtlauac:
> Thanks, this looks very good. One minor comment: I find the alert to be _very_ aggressive. Can we disable the button with a tooltip saying the doc isn't available instead?

I share the feeling. I've proposed something in the comments in this direction, yet it doesn't have a tooltip

sabine:
I was going for something that motivates us to look into fixing the issues with docs not building, but certainly, we can do something more subtle. :grin: 

We can't remove the links when there are sometimes error messages to look at. This is an altogether separate issue to fix: Having a button that leads to the errors page, and disabling the button altogether when it would be a 404.

I changed the color of the info-box to be much more subtle, using the background color of the navbar:

Screenshot 2022-11-29 at 12-26-13 irmin 0 9 4 · OCaml Package

It is necessary to find out when exactly do we have an errors page to look at, in order to remove the 404 links. So that we can render a link, when we have errors to look at, and render just an error message about missing docs otherwise.

tmattio (approved):
LGTM!

src/ocamlorg_frontend/pages/package_overview.eml
sabine: Button color is about visual hierarchy. It's about guiding the eye of the user to the most important parts of the page. So making one of the most important links on the page a flashy Call-to-Action is reasonable. (We do want to make sure the user finds the documentation immediately, as this is most likely the reason they're browsing in the first place.)

I agree that the shade of orange we currently have is problematic because it does not meet the WCAG accessibility standards wrt to contrast.

sabine:
I realized that removing the tabs means we don't have a way back from package documentation to package overview. So I made the name of the package a link back to the overview page.

PR #646, opened 2022-11-30
adjust whitespace on package docs to get the #-symbol on anchor tags to show properly

sabine:
There used to be the issue of the #-symbol of the anchors in headings being either cut off or not showing (due to `overflow-hidden` and padding-left on the content area not being large enough to avoid the cutoff).

This patch removes `overflow-hidden` on the content area (so that we do not need to rely on padding to show the #-symbol) and adjusts the gap between left sidebar, content, and right sidebar. Right sidebar's padding is reduced to only padding-top and width is reduced.

| before | after |
|-|-|
| Screenshot 2022-11-30 at 12-04-04 Documentation · streaming 0 8 0 · OCaml Packages | Screenshot 2022-11-30 at 12-03-56 Documentation · streaming 0 8 0 · OCaml Packages |
| Screenshot from 2022-11-30 12-05-09 hovering over the heading, the #-symbol does not show because of `overflow-hidden`| Screenshot from 2022-11-30 12-04-59 #-symbol is visible when hovering and there is enough gap between left sidebar and content area to guarantee no overlap between sidebar and #-symbol|

### DIFF
--- a/src/ocamlorg_frontend/pages/package_documentation.eml
+++ b/src/ocamlorg_frontend/pages/package_documentation.eml
@@ -30,7 +30,7 @@ Package_layout.render
     </div>
   </button>
 
-  <div class="flex flex-col lg:flex-row md:gap-6">
+  <div class="flex flex-col lg:flex-row md:gap-12">
     <div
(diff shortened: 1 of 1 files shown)


tmattio (approved):
Looks great!

PR #647, opened 2022-11-30
More multicore and effects academic papers

Octachron:
On a suggestion of @fpottier, this PR proposes to update the list of academic papers about multicore OCaml and effects.

The PR adds the two PLDI articles discussing the OCaml memory model and the implementation of effect handlers, two articles on proving multicore program correct, and a last article on proving programs with effect correct.

I have also added the featured tags to the two PLDI articles. To keep the same number of featured articles, I have removed the featured tag from the Eff article and the previous version (the ICFP one) version of the memory article.

### DIFF
--- a/data/papers.yml
+++ b/data/papers.yml
@@ -1,4 +1,156 @@
 papers:
+  - title: Bounding data races in space and time
+    publication: Programming Language Design and Implementation (PLDI)
+    abstract: >
+      We propose a new semantics for shared-memory parallel
+      programs that gives strong guarantees even in the presence
+      of data races. Our local data race freedom property guar-
+      antees that all data-race-free portions of programs exhibit
+      sequential semantics. We provide a straightforward oper-
+      ational semantics and an equivalent axiomatic model, and
+      evaluate an implementation for the OCaml programming
+      language. Our evaluation demonstrates that it is possible to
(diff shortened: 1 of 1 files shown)


cuihtlauac:
It looks good to me.

Off-topic remarks:
* Should we include paper DOI?
* List of paper should be sorted in reverse chronological order in order to see fresh papers first

fpottier:
Hi!

* Most recent papers first sounds wise.
* Regarding DOIs, I don't think the DOI number itself matters. What matters is to provide a URL to a publicly-accessible version of the paper (i.e., not behind a paywall), if possible.

Octachron:
Idle thought: it could make sense to have the papers data in the bibtex format (or an extension), there are few OCaml libraries that can parse this format, and it is probably the easiest format for academics to submit update to the list of articles.

cuihtlauac:
Makes a lot of sense. Which library would you recommend?

Octachron:
There might be less libraries that handle bibtex that I thought. I had my own library in mind when I wrote the message, and there is `bibtex2html` which can handle transforming bibtex to html, but that would probably require to use that generated html as a data file. Both options could work but I am not sure which one would be lighter for ocaml.org .

Issue #649, opened 2022-11-30
Sort science papers in reverse chronological order


cuihtlauac:
Show fresh papers first.

See: https://github.com/ocaml/ocaml.org/pull/647#issuecomment-1332220532

cuihtlauac:
Fix: https://github.com/ocaml/ocaml.org/pull/651

cuihtlauac:
Fix merged

PR #651, opened 2022-12-01
Sort papers in reverse publication date order

cuihtlauac:
Generated a sorted list of papers from the Yaml. Most recent papers
are coming first.


cuihtlauac:
Issue: https://github.com/ocaml/ocaml.org/issues/649

PR #658, opened 2022-12-02
Add job publication date, sort fresh first

cuihtlauac:
- Remove obviously closed positions
- Remove TrustInSoft jobs
- Update SimCorp Jobs
- Sort jobs by publication date
- Add dates to functori, Marigold and CEA positions
- Add dummy dates to SimCorp jobs
- Make publication date optional
- Add Nobias Pharma job publication date
- Fix typos in dates
- Show pub date in board, if available

### DIFF
--- a/tool/ood-gen/lib/job.ml
+++ b/tool/ood-gen/lib/job.ml
@@ -2,6 +2,7 @@ type metadata = {
   title : string;
   link : string;
   location : string;
+  publication_date: string;
   company : string;
   company_logo : string;
 }
@@ -11,6 +12,7 @@ type t = {
   title : string;
   link : string;
   location : string;
+  publication_date: string;
   company : string;
   company_logo : string;
 }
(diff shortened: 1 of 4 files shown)


tool/ood-gen/lib/job.ml
```diff
@@ -65,10 +70,13 @@ type t =
...
   ; link : string
   ; location : string
+  ; publication_date : string option
   ; company : string
   ; company_logo : string
   }
   
-let all = %a
+let all =
+  let job_date j = Option.value ~default:"1970-01-01" j.publication_date in
```
tmattio, line 79: I'd suggest letting it fail with an exception here if there's no publication date, instead of giving it a wrong publication date. Actually, the check can be pushed to the file parsing by making the `publication_date` a `string` instead of a `string option`

cuihtlauac: The bogus date is only used for sorting purpose. I agree surfacing such a date would be bizarre. On the other hand job offering rarely come with a publication date, we probably have to cope with it.

tmattio:
I think the HTML markup is missing a closing tag or something 😁 

image

Issue #474, opened 2022-06-02
Playground should provide a feedback when it runs


Khady:
I was giving a try to https://ocaml.org/play today. And ended up hitting the `run` button many times with the hope of something happening. My code didn't have side effect, so I couldn't see if anything was really happening. I later realized that even on bad code there's no feedback. For example if I change the example to add a typo and press the `run` button, nothing happens

```ocaml
type ('a, 'b) res = Ok of 'a | Err of 'b

let pp_res = function
  | Ok _ -> print_endlin "OK!" (* should yell a me for bad print_endlin *)
  | Err _ -> print_endline "Error!!"

let () = pp_res (Ok 10)
```

patricoferris:
With the integration of Merlin and the changes in #670 I think this should be much better now.

![Image](https://user-images.githubusercontent.com/20166594/205490317-d3c89391-c366-4b25-a0a3-99ec8c7eee40.png)

cuihtlauac:
Fixed by https://github.com/ocaml/ocaml.org/pull/670

PR #670, opened 2022-12-04
Preserve newlines and exceptions in playground

### LINKED ISSUE #650: playground doesn't correctly handles newlines in the output
When printing newlines for example with `print_endline` the text in the output box doesn't contain newlines. Instead the multiple lines are concatenated with spaces and printed as a single line. It is a surprising behavior.

```ocaml
let () = print_endline "ocaml"
let () = print_endline "love"
let () = print_endline "story"
```

gives

```
ocaml love story
```

image

patricoferris:
Currently the output section of the playground doesn't show exceptions or render newlines. This PR tries to fix both of these issues. 

(Fix https://github.com/ocaml/ocaml.org/issues/650)

**Before**

Screenshot 2022-12-04 at 12 17 13

**After**

Screenshot 2022-12-04 at 12 16 03

### DIFF
--- a/src/ocamlorg_playground/main.ml
+++ b/src/ocamlorg_playground/main.ml
@@ -52,12 +52,28 @@ let get_el_by_id s =
       Console.warn [ Jstr.v "Failed to get elemented by id" ];
       invalid_arg s
 
-let cyan el = El.set_inline_style (Jstr.v "color") (Jstr.v "cyan") el
+let cyan = "cyan"
+let red = "red"
+
+let render_output color = function
+  | None -> None
+  | Some output ->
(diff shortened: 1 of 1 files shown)


cuihtlauac (approved):

cuihtlauac:
This looks good to me. 
Fix [474](https://github.com/ocaml/ocaml.org/issues/474) and [650](https://github.com/ocaml/ocaml.org/issues/650).

tmattio:
Thanks a lot @patricoferris!!

I just tested and it looks like running the code now prints the interface of the module as an error:
image

Was it intentional?

PR #692, opened 2022-12-08
Set ocamlformat to 0.24.1

### LINKED ISSUE #689: Ocamlformat version discrepancy
PR https://github.com/ocaml/ocaml.org/pull/660 has allowed upgrading many opam packages of the ocaml.org switch. However, this has upgraded ocamlformat (to 0.24.1 at the time of writing) while `.ocamlformat` requires 0.21.0. The discrepancy has to be resolved, either my stalling ocamlformat in the dependencies or updating `.ocamlformat`. Maybe we should do both.

cuihtlauac:
Fix: https://github.com/ocaml/ocaml.org/issues/689

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -33,8 +33,6 @@
   (ocaml
    (>= 4.14.0))
   dune
-  (ocamlformat
-    (= 0.24.1))
   cohttp
   cohttp-lwt-unix
   bos
--- a/dune-project
+++ b/dune-project
@@ -33,6 +33,8 @@
   (ocaml
    (>= 4.14.0))
   dune
+  (ocamlformat
+    (= 0.24.1))
   cohttp
   cohttp-lwt-unix
   bos
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,7 +46,7 @@ jobs:
         run: opam install --deps-only --with-test .
 
       - name: Install ocamlformat
-        run: opam install ocamlformat.0.21.0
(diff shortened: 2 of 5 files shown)


dune-project
```diff
@@ -33,6 +33,8 @@
   (ocaml
    (>= 4.14.0))
   dune
+  (ocamlformat
```
tmattio: ocamlformat isn't a dependency of the project so we shouldn't add it here

cuihtlauac: Actually, I put it here to trigger this discussion :-)

If it's not installed, automatic PR checks are likely to fail as they are using it and the dev is not. Although not technically a dependency of the project, it is nevertheless on the critical path for seamless maintenance. 

An alternative could be to have `make fmt` in a git pre-push hook. But one could argue that would make ocamlformat a dependency :-)

tmattio: Having it installed alongside the other dependencies won't help to enforce that contributors format their code. A git hook is an option, but I think it requires manual steps to setup? So we'd need to document it and it's not a given that contributors will do it. So it doesn't completely solve the problem either.

IMO asking contributors to format their code when we see that the CI check failed is low friction and I don't see a solution that removes the need for a CI check (or reduces the odds that the check will fail)

tmattio (approved):
Just pushed a few changes to update the version in the CI and Makefile, and format the code. Feel free to merge when you want :)

cuihtlauac:
Thanks, I had overlooked the CI & Makefile, I should have used git grep

PR #694, opened 2022-12-09
Fix problems with exercices ToC and order

### LINKED ISSUE #693: Problems with problems
* Exercises 9 & 10 are missing from the list of exercises
* Some pairs of exercises are reversed: (61, 61A), (62, 62B) and (70, 70C)

cuihtlauac:
- Fix empty string tag
- Fix problem numbers

Fix: https://github.com/ocaml/ocaml.org/issues/693


tmattio (approved):
Good catch! LGTM

PR #696, opened 2022-12-12
Ood: Inline book metadata

cuihtlauac:
Replace field Book.t.meta by fieds from Book.metadata


sabine (approved):
LGTM. This improves consistency with the other files.

PR #698, opened 2022-12-12
Odd: Use type aliases for identical types

cuihtlauac:


sabine (approved):
`meetup.ml` and `paper.ml` seem to do this the opposite way around, i.e. `type metadata = t`. Could go for full consistency by adjusting `meetup.ml` and `paper.ml` as well. But this is also fine.

PR #699, opened 2022-12-12
Ood: Code uniformization, paper

cuihtlauac:
* Let metadata be the @@deriving yaml type
* Let Paper.t be an alias of Paper.metadata
* Misc required changes


sabine (approved):
Ah ok, you already saw and fixed that. LGTM.

PR #713, opened 2022-12-16
Repair build broken by mirage-kv 6.0.0

cuihtlauac:
In mirage-kv 5.0.0 the function `last_modified` returned a result of
type:
```ocaml
  (int * int64, error) result Lwt.t
```

This was changed in 6.0.0, where it returns a result of type:
```ocaml
  (Ptime.t, error) result Lwt.t
```

See: https://github.com/mirage/mirage-kv/commit/9161b54b13411d740151d9e7bf7f403dbcc61583 from PR https://github.com/mirage/mirage-kv/pull/34

This caused a typing error. Removing the call to function
```ocaml
  Ptime.v : int * int64 -> Ptime.t
```

solved the issue.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -37,8 +37,7 @@
   cohttp-lwt-unix
   bos
   crunch
-  (mirage-kv-mem
-    (<= 3.1.0))
+  mirage-kv-mem
   (dream
    (>= 1.0.0~alpha3))
   dream-accept
--- a/ocamlorg.opam
+++ b/ocamlorg.opam
@@ -24,7 +24,7 @@ depends: [
   "cohttp-lwt-unix"
   "bos"
   "crunch"
-  "mirage-kv-mem" {<= "3.1.0"}
+  "mirage-kv-mem"
   "dream" {>= "1.0.0~alpha3"}
   "dream-accept"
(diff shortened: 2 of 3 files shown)


dune-project
```diff
@@ -37,8 +37,7 @@
   cohttp-lwt-unix
   bos
   crunch
-  (mirage-kv-mem
-    (<= 3.1.0))
+  mirage-kv-mem
```
patricoferris, line 40: Would be good to add a lower bound of `6.0.0` on this I think, just bumped into this because I had an old version and an out of date opam :))

cuihtlauac: I have mixed feelings about this. I added this right before the release of OCaml 5.0.0 in [#710](https://github.com/ocaml/ocaml.org/pull/710) as a sufficient mean to make `opam` install `mirage-kv.5.0.0`, transitively. Having a broken build for ocaml.org on the day of that release would have been painful. I didn't add `mirage-kv < 6.0.0`, although it was what was really needed because `mirage-kv` isn't a direct dependency. In this PR, changes in `static.ml` are intended to support `mirage-kv.6.0.0`.

Issue #733, opened 2022-12-26
Playground ignores last statement unless it ends with a new line


Timmmm:
If you put `1+2;;` into the playground and hit Run, nothing happens. If you add a new line after the `;;` it works as expected.

Took me ages to figure this out. Very confusing for a beginner, trust me!

I guess an easy fix is to add an extra newline somewhere maybe on this line? https://github.com/ocaml/ocaml.org/blob/main/src/ocamlorg_playground/main.ml#L137

To be honest I couldn't figure out why that line adds `;;` so maybe it's not the right place.

cuihtlauac:
Thanks @Timmmm. Nice observation, I hadn't figured out that. You're right this needs to be fixed. There are a number of issues with the playground: https://github.com/ocaml/ocaml.org/issues/717

tmattio:
Looks like @cuihtlauac and I can't reproduce anymore, so closing for now, but feel free to re-open if you see the bug happen again

Issue #717, opened 2022-12-17
OCaml playground broken


bloa:
I saw the recent announcement for OCaml 5 and ended up on the online playground.
The default example does not run, instead leads to the following error:

    Line 7, characters 0-0:
    Error: Syntax error

As seen on this screenshot:
![chromium](https://i.imgur.com/ex33Mbu.png)

Steps to replicate:
1. go to https://ocaml.org/play
2. click on "Run"

The error seems to appear as soon as the program includes a newline.  
Single-line programs don't trigger a syntax error but seem to silently fail instead.

Happy holidays :)

cuihtlauac:
Thanks for this detailed report. Happy holidays too

tmattio:
@cuihtlauac and I can't reproduce anymore, so closing for now, but feel free to re-open if you see the bug happen again

PR #759, opened 2023-01-05
Upgrade playground to OCaml 5

tmattio:
This update the playground to be compatible with OCaml.
Upstream dependencies are updated: https://github.com/voodoos/merlin-js/commit/74edff465942cc4f0b3b51b3f8264db53a460aa2 and https://github.com/tmattio/js_top_worker/commit/768b88b272668d10b876a71f4317223a784d1723.

Since the whole website can't be upgraded to OCaml 5 easily (blocked by ocamlnet in River, Dream, lambdasoup, etc.), we separate the playground into its own directory and commit the generated JavaScript files to the repository. To update the playground, one can setup an OCaml 5 switch and work in the `playground/` subdirectory.

I also wasn't able to reproduce https://github.com/ocaml/ocaml.org/issues/717 and https://github.com/ocaml/ocaml.org/issues/733, either with this patch or at https://ocaml.org/play. Pretty sure I saw the syntax error before as well, but it seems to be fixed. Can you confirm you can't reproduce these with this patch and live @cuihtlauac?


cuihtlauac:
I can't reproduce issues https://github.com/ocaml/ocaml.org/issues/717 and https://github.com/ocaml/ocaml.org/issues/733 any longer on the live site. Upstream update must have make it go away.

Screenshot_2023-01-05_13-50-26

cuihtlauac:
Right now, building this PR fails on my setup (after opam update && opam upgrade). However, it's probably my set up as CI build is pass.

```
ocaml.org:ocaml5-playground>make start
opam exec -- dune build --root .
Browserslist: caniuse-lite is outdated. Please run:
  npx browserslist@latest --update-db
  Why you should do it regularly: https://github.com/browserslist/browserslist#browsers-data-updating

Done in 392ms.
Browserslist: caniuse-lite is outdated. Please run:
  npx browserslist@latest --update-db
  Why you should do it regularly: https://github.com/browserslist/browserslist#browsers-data-updating

Done in 1453ms.
File "src/ocamlorg_web/lib/.ocamlorg_web.objs/native/_unknown_", line 1, characters 0-0:
Fatal error: exception Stack overflow
make: *** [Makefile:5: all] Error 1 
```

cuihtlauac:
Something weird is taking place. Using a fresh setup:
```shell
git clone https://github.com/ocaml/ocaml.org.git
cd ocaml.org
make switch 
make start
```

I get the same issue as above and as in the MacOS CI build failure: https://github.com/ocaml/ocaml.org/actions/runs/3846834471/jobs/6552585655

Looks like a system dependency issue. My version of Ubuntu (22.10) probably has the same as the MacOS CI one, which breaks. Your set up and the Ubuntu CI probably has a version of the same dependency which doesn't have the issue.  

Update: Ubuntu CI is 22.04.1

tmattio:
Same thing is happening with the Docker build in the CI.
I don't have a clear idea of where this comes from. The only thing I can think of is that we probably increased the size of the generated JS assets by passing `--enable=effects` to js_of_ocaml, and that this in terms increases the size of the generated module that are crunched into the binary, causing the stack overflow.

If that's the case, we probably reached the limit of what we can do with the current model (crunching all the assets in the binary), and we need to explore other options. One such option could be to host the playground files on GitHub and setup redirections.

tmattio:
@cuihtlauac Can you try with the last commit? I'm still checking the Docker build, but if the issue was the size of the JavaScript assets, you should be able to run the server now.

cuihtlauac:
Compilation is now pass. However, I get a empty & unresponsive play page :-(

Screenshot_2023-01-05_15-49-52

tmattio:
Sorry, it was working locally because of the browser's cache it should work better now if you want to try again

cuihtlauac:
It works. Why is the output of Printf taking place _before_ the typing output?

Screenshot_2023-01-05_16-19-47

tmattio:
> Why is the output of Printf taking place before the typing output?

It's reusing the top-level, so the output order is the same as Utop (and OCaml's top-level).
That's also the current behaviour, not something introduced in this PR.

cuihtlauac:
Comment lines need to be wrapped tighter (say 70 chars) in order to look nicer on small screens.

tmattio:
For clarity, the current changes introduced in this PR are:

- Separate the playground into its own opam package and move it under the `playground/` subdirectory
- Update the pin of `js_top_worker` and `merlin-js` to versions that are compatible with OCaml 5
- Promote the generated JavaScript files for the top-level in the source tree and commit them to the repo under `playground/asset`
- Add a router rule that serves (from the filesystem) the files in `playground/asset/` as `/play/**`
- Hardcode a suffix (`-v1`) in the name of the playground assets so we can easily invalidate the cache when we update them
- Update the welcome message and code examples of the playground (the example is a parallel version of the Fibonacci sequence)
- Add documentation `HACKING.md` on how to re-generate the playground assets
- Update the docker file to copy the files from `playground/asset`

Does that list of changes look sensible to you @cuihtlauac?

One benefit of serving the playground from the filesystem is that the creation of the server's binary should be faster.
The setup with the `-v1` suffix might prove to be cumbersome, but we don't change the playground that often and if we see that it's a pain point, we can reconsider the approach later.

cuihtlauac:
BTW. Isn't `playground/asset/playground-v1.js` automatically updated when `playground/src/example.ml` is changed?

tmattio:
> BTW. Isn't playground/asset/playground-v1.js automatically updated when playground/src/example.ml is changed?

Not anymore, to re-generate the playground, you now need to manually build the project in `playground/`

cuihtlauac:
Repository structure at the end of `HACKING.md` no longer in sync.

cuihtlauac:
I've got a file named `#playground-v1.js#` in `playground/asset` is it intentional?

tmattio:
> I've got a file named #playground-v1.js# in playground/asset is it intentional?

You mean with the sharp signs? It should be just `playground-v1.js`, and looking at the committed file, there's no `#playground-v1.js#`, so not sure what created it.

tmattio:
> Repository structure at the end of HACKING.md no longer in sync.

Done, thanks!

cuihtlauac:
It may be more robust to allow building the playground in a local switch. In order to do that, I believe `js_top_worker` is missing some dependencies. I've created a PR in those lines: https://github.com/tmattio/js_top_worker/pull/1

tmattio:
Thanks, I've updated the pin to your commit

cuihtlauac:
I've successfully built and run the playground using a local switch inside `/playground` 

I think you can merge.

PR #761, opened 2023-01-05
add js library 'htmx' to preserve sidebar scroll position via AJAX-fetch

### LINKED ISSUE #623: Usability problem in documentation: Side menu position is reset when selecting a module.
For example, when using [documentation for ocaml-compiler 4.07.1](https://ocaml.org/p/ocaml-base-compiler/4.07.1/) :
You select Library ocamlcommon, then module Env (on left side/"Module" to develop not on right sode/"Env" that would jump to module Env's signature), then you select its module PathMap.

Module Env.PathMap signature is displayed as expected, but the vertical position of the side menu is immediately reset (IN THIS PACKAGE ... library bigarray ...). This behaviour is really not usable and that makes navigation exhausting.

This is a pure GUI issue. The selected position of the side menu MUST be managed/stored.
New comers will be sensitive to this boring usability problem...

I'm not aware about how to fix that for the moment because I didn't develop this website and didn't look at its code.
So let's discuss and see if you can fix that simply or if I can help fixing that.

sabine:
This patch preserves the sidebar scroll state by performing an AJAX-fetch via the library htmx (https://github.com/bigskysoftware/htmx) and replacing the content area and sidebars with the fetched page's content. The new URL is pushed to the window history to reflect the new location.

This is a noticeable UX improvement on the learn page as people can navigate via the sidebar while keeping their position in the sidebar.

It's also an improvement on the package docs pages, however, there we have the issue that the sidebar content (default open modules) varies depending on where you are in the tree, so navigating is still a bit jumpy (however, much less than before).

Should resolve #623


tmattio:
I agree this is a great UX improvement, we should definitely try to add something like this in the site. However, the current implementation breaks the documentation layout:
image

(the left side bar is not below the content, and there's a huge space after the doc content)

sabine:
Doh, sorry about that. Indeed, that won't work as is on the learn_layout. I'll refactor the learn_layout a bit to make it work!

sabine:
Rebased on main after #763.

Also removes hiding of other libraries on the package documentation when drilling down into a package. Since the scroll state is less janky via the AJAX-fetch, there's no need to hide libraries.

It's not perfect, but better than the existing situation and we can move away from htmx when we find a better way to do this.

sabine:
Rebase after conflicts.

cuihtlauac:
It is indeed a great improvement. I haven't found major issues.

If I understand correctly, htmx is intended as an interim solution. Does it mean an issue needs to be created right after merging this?

As a side question, do we agree on the longer term goal to have a left navigation bar which does not change at all unless it's clicked or moved with its own bar?

cuihtlauac (approved):

sabine:
`htmx` is as much interim as `alpinejs`. It's adding the ability to declaratively describe when to do an AJAX-fetch to swap in HTML content. For some pages, it may even be the best solution in the long term. So I don't think we need an issue for moving away from it.

The sidebar on the package documentation is its own beast, and may need a more sophisticated approach (client-side rendering) to give us full control over how it behaves wrt scroll state. (So that may warrant an enhancement issue.)

sabine:
Rebase on main. Do we pull the trigger? :thinking:

PR #764, opened 2023-01-06
Add a link to the opam file on the package overview page

### LINKED ISSUE #755: Edit link to opam file from package page
As requested by @avsm in https://github.com/ocaml/infrastructure/issues/26 :

> the ocaml.org package page isn't at parity in terms of the information it displays vs opam.ocaml.org. The ~~two~~ missing field~~s are package conflicts, and an~~ (is) edit link to the opam file.

This is required to merge opam.ocaml.org and ocaml.org/p

tmattio:
Closes #755

### DIFF
--- a/src/ocamlorg_frontend/pages/package_overview.eml
+++ b/src/ocamlorg_frontend/pages/package_overview.eml
@@ -52,29 +52,6 @@ Package_layout.render
                 </button>
             </div>
         </div>
-        <div class="mt-4">
-            <% homepages |> List.iter (fun homepage -> %>
-            <a href="<%s homepage %>" class="text-primary-600 hover:underline">
-                <%s homepage %>
-                <svg
-                    xmlns="http://www.w3.org/2000/svg" class="h-5 w-5 ml-2 inline-block -mt-1" fill="none"
-                    viewBox="0 0 24 24" stroke="currentColor">
-                    <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2"
-                        d="M10 6H6a2 2 0 00-2 2v10a2 2 0 002 2h10a2 2 0 002-2v-4M14 4h6m0 0v6m0-6L10 14" />
-                </svg>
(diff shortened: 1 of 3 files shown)


sabine (commented):
Might need better contrast, icon size / stroke width of icon is different to the icon of the repository link.

LGTM otherwise.

tmattio:
Thanks, I updated the stroke width. The contrast is the same as the other links, so I think we can keep it like this?

sabine:
Screenshot 2023-01-06 at 12-46-05 solo5 0 7 5 · OCaml Package

Icon size is inconsistent with the icon above which is used in a similar role.

Other than that, I think it's okay.

cuihtlauac:
BTW, do we need the arrow-top-right-on-square? I don't get the point

tmattio:
Ok, the links in the overview are getting messy and inconsistent, so I did some work to make it more consistent.
Here's the result:
image
And when the documentation fails:
image

cuihtlauac:
Much nicer!

cuihtlauac:
Me being cheeky: Shouldn't the source tarball link receive the same treatment?

tmattio:
>  Shouldn't the source tarball link receive the same treatment?

The sources are a bit different: we need to display the hash, which is long, so we need a way to scroll it horizontally if we want it on the side panel. It's also common to have multiple sources, so a separate section seems sensible.

cuihtlauac (approved):

sabine:
It is an great improvement (specifically the icons being aligned so neatly, and using better icons goes such a long way).

My only peeve is the removal of the emphasized nature of the documentation link and the repository link. It would be better if these highly important items would still be more emphasized so even new users can find them quickly. I find the documentation link much harder to click now because it is so much smaller.

I suggest returning the documentation to a CTA-button and making the repository link highlighted in orange by default (as previously, except that now it nicely sits together with the other, less important links).

PR #773, opened 2023-01-09
Do not allow opam updates beyond freeze

### LINKED ISSUE #772: Dockerfile incorrect management of opam dependencies
Project `Dockerfile` intention is to “freeze time” at some point in `opam-repository` history. 

This used to be done by having:
```dockerfile
RUN cd ~/opam-repository && git pull origin master && git reset --hard 6496b2727e && opam update
```

This was changed into:
```dockerfile
RUN cd ~/opam-repository && git pull origin 6496b2727e51cc19c9ebaeb73008ded920eea501 && opam update
```

by PR https://github.com/ocaml/ocaml.org/pull/671, which I authored. My bad, this was faulty.

I wrongly assumed the first line of the `Dockerfile`, which contains this:
```
FROM ocaml/opam:alpine-3.17-ocaml-4.14 as build
```

Always resolves into the same image. This assumption is false; this image is built and renewed periodically.

Building on my faulty assumption, I thought: “It's useless to download commits from 6496b2727e to the current HEAD and discard them right away. Let's just move up to where it's needed”. The tricky thing is, at the time you write something as I did, it works. But some time after, HEAD overtakes the hardcoded commit, pull does nothing, and we're floating with a recent HEAD and corresponding untested set of dependencies. 

The even trickier thing is, for a reason I fail to understand yet, I wasn't immediately able to reproduce the issue on my side because referring to `ocaml/opam:alpine-3.17-ocaml-4.14` did not invalidate my local docker cache!

Thanks: @mtelvers for helping me understand this

cuihtlauac:
Fix: https://github.com/ocaml/ocaml.org/issues/772

Should we pin the docker image used for build? i.e. replace:
```dockerfile
FROM ocaml/opam:alpine-3.17-ocaml-4.14 as build
```

by:
```dockerfile
FROM ocaml/opam@sha256:a921b8cd3a1015adced68ee3a9bdf4d8488e705e4cb4d1e6b799793cce5daaa3 as build
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -3,8 +3,8 @@ FROM ocaml/opam:alpine-3.17-ocaml-4.14 as build
 # Install system dependencies
 RUN sudo apk update && sudo apk add --update libev-dev openssl-dev gmp-dev oniguruma-dev
 
-# Set opam repo at: f72694d Merge pull request #22599 from dbuenzli/b0-publish-ptime.1.1.0
-RUN cd ~/opam-repository && git pull origin f72694dcd7c7cd6c6438f3f988c4742f0758737a && opam update
+# freeze branch was opam-repo HEAD at the time of commit
+RUN cd ~/opam-repository && git checkout -b freeze b457e9f3d6 && opam update
 
 WORKDIR /home/opam
 


sabine:
As I understand now, this restores the previous semantics of checking out the specific commit id on opam-repository. That means `opam-repository` is pinned (as it was before) to that commit.

Regarding pinning the docker image: No, I think we should not pin that, because this is infrastructure that should be able to auto-update in a mostly backwards-compatible way. When that breaks, it's reasonable to be fixed ASAP because the infrastructure must get security updates.

ETA: After looking more closely, it seems that the docker image used to build is different from the docker image that is used to run ocaml.org, so it would be an option to pin the docker image used for the build. :thinking:

sabine (approved):

tmattio (approved):

PR #780, opened 2023-01-10
Rename `.gitattribute` to `.gitattributes`

tmattio:
I always make the same mistake. Pro tip: if you use VSCode, .gitattributes gets a custom icon, so you can easily see if you make a type or not.


cuihtlauac:
I was wondering this morning why it wasn't taken into account.

Shall add a rule for `.eml` file to make them count as OCaml ? (or HTML ?)

tmattio:
> Shall add a rule for .eml file to make them count as OCaml ? (or HTML ?)

I'd be inclined to treat them as HTML. We try to keep the OCaml code in these to a minimum so they can be worked on just like HTML files by frontend developers.

Issue #673, opened 2022-12-05
Package documentation headings formating including <code>


cuihtlauac:
PR https://github.com/ocaml/ocaml.org/pull/639 has a small bug on section titles when then include &lt;code&gt; see, for instance:

https://ocaml.org/p/opam-monorepo/0.3.2/doc/index.html#opam-monorepo-docs

![Screenshot_2022-12-05_11-48-08](https://user-images.githubusercontent.com/1917592/205619138-1a83046c-53d7-4549-ba85-96cd46175be0.png)

sabine:
What exactly is the bug in relation to #639? It looks normal to me:

![Screenshot 2022-12-05 at 12-19-59 Documentation · opam-monorepo 0 3 2 · OCaml Packages](https://user-images.githubusercontent.com/6594573/205625012-83ed5f99-34c3-4f6c-a2ca-0a35c708d6e0.png)

Can you try to reload the page (disable cache via the network tab of your browser)?

cuihtlauac:
I find the difference of font bitween “opam monorepo” and “Docs” a bit weird, don't you?

sabine:
Oh, that. Apparently, that's how the package author wrote the heading. There's no connection to #639 as the patch did nothing with respect to fonts. 

I agree, this looks jarring.

Two options: 
1. set the font family for `h1 code, h2 code`, etc. to the heading font family
2. Play with styles to make it look at least a bit more intentional/similar (but I think we would need another, stronger, font weight on the monospace font for that) to honor the markup the package authors chose.

sabine:
Ok, so after looking at this a bit more... My impression is that the main dissonance here is that the monospace font has a different font size. And this actually applies to all `code` tags on the site.

The culprit is https://github.com/tailwindlabs/tailwindcss-typography/blob/4c3b76ccc8120f307514d5721598763fd09761c0/src/styles.js#L76-L84.

I can see that GitHub in their own styles also makes the `code` elements smaller, but it tries to offset this by giving them a background color. (I think it looks bad on GitHub too, but it could easily be remedied by increasing the top padding so that the background aligns with the line-height of the surrounding text.)

However, on ocaml.org where we have no background color for the `code` elements, the misalignment just looks way off. Either the `code` elements need to be the same font-size as the surrounding text, or we need to add a background as a visual element to restore the perception of correct alignment.

[...]

PR #785, opened 2023-01-11
make code tags render in the same font size as surrounding body text

sabine:
It has been identified in #673 that the `<code>` tags (particularly within headers, but also throughout the site) are misaligned with the rest of the text (due to different font size).

This patch proposes changes the font size of `<code>` tags to be the same as the surrounding text.

| before | after |
|-|-|
| Screenshot 2023-01-11 at 13-58-20 Documentation · opam-monorepo 0 3 5 · OCaml Packages | Screenshot 2023-01-11 at 13-58-23 Documentation · opam-monorepo 0 3 5 · OCaml Packages |

### DIFF
--- a/tailwind.config.js
+++ b/tailwind.config.js
@@ -7,15 +7,28 @@ module.exports = {
     extend: {
       typography: {
         DEFAULT: {
-          css: {
+          css: [{
             'code::before': {
               content: '""',
             },
             'code::after': {
               content: '""',
             },
-          }
-        }
+          },{
+            h1: {
+              fontWeight: 700,
+            },
+            code: {
+              fontSize: "1em",
+            },
+            'h2 code': {
+              fontSize: "1em",
+            },
+            'h3 code': {
+              fontSize: "1em",
+            },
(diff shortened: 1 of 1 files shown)


cuihtlauac:
It is indeed better. Don't you think the difference in weight between the fonts remains a bit awkward?

sabine:
The weight difference is mildly irritating, but might serve as emphasis that the author wrote a `code` block there. :shrug:

I think there's only two ways to reduce the weight difference:
1. bring in and serve a dedicated web font that provides such a strong weight, for the headings, or
2. reduce the weight of the headings all throughout the site (from 800 to 700).

Here's what it would look like with 700 font weight on the `h1` heading:

Screenshot 2023-01-11 at 14-24-31 Documentation · opam-monorepo 0 3 5 · OCaml Packages

sabine:
Included the change to 700 font weight on `h1` elements (other heading elements are of a lighter weight anyways).

cuihtlauac (approved):

PR #788, opened 2023-01-11
add a background-blurring overlay and make the sidebar not cover the full screen

sabine:
On the package documentation page, there was a stray `w-full` in the sidebar styles which made the sidebar (on md- and smaller screens) cover the full screen when opened.

This patch also adds an overlay to blur and darken the background when the site is open.

|before|after|
|-|-|
|Screenshot from 2023-01-12 10-11-26|Screenshot from 2023-01-12 10-11-15|

### DIFF
--- a/src/ocamlorg_frontend/layouts/learn_layout.eml
+++ b/src/ocamlorg_frontend/layouts/learn_layout.eml
@@ -102,10 +102,13 @@ inner_html
         </svg>
       </div>
     </div>
+    <div class="fixed z-10 inset-0 bg-background-dark-blue/20 backdrop-blur-sm lg:hidden"
+      :class='sidebar ? "" : "hidden"' aria-hidden="true" x-on:click="sidebar = ! sidebar"></div>
+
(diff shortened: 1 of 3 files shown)


cuihtlauac (approved):

tmattio:
If you do this, I think we should blur (or change the opacity) of the page when the sidebar is open, what do you think?

tmattio:
I feel bad always referring to the tailwindcss documentation to illustrate my comments, but they are full of great UI patterns: https://tailwindcss.com/docs/installation

image

sabine:
Yeah sure, that makes sense. I'll add that. Either way, the learn page had the smaller slide-in sidebar for a while.

Screenshot updated to show the new overlay.

ETA: no need to feel bad referring to the Tailwind docs. Ideally, one day we will be as good or better with ocaml.org. :grin: There is really nothing wrong with taking inspiration from sites that work pleasantly.

sabine:
Rebase after conflict, should be good to go now.

Issue #730, opened 2022-12-22
Feature request: support `latest` in the url that links to the latest version of package documentation


arbipher:
I searched the issues, and it's similar to #726 but this is for package documentation.

It would be useful when the documentation is bookmarked with the `latest` in the URL, then we don't need to update the links periodically.

e.g.
Current: `https://v3.ocaml.org/p/base/v0.15.1/doc/Base/index.html`
Expected: `https://v3.ocaml.org/p/base/latest/doc/Base/index.html`

This link `https://v3.ocaml.org/p/base/` can redirect to the latest version `https://v3.ocaml.org/p/base/v0.15.1/`. However, the entry page is not convenient to use.

cuihtlauac:
Thanks, @arbipher; you are right; the `latest` thing should also apply to the package documentation and is badly needed. Here is another issue which is relevant to this discussion:

https://github.com/ocaml/ocaml.org/issues/534

arbipher:
Thanks for pointing out an existing issue.

I think we are concerning the same problem. If the fixed URL for the search engine is solved, so does the bookmark.

PR #793, opened 2023-01-12
Adding latest url to package links

SaySayo:
fix for (#730). This PR adds, 

- for every package, a URL to refer to that package's latest version: `/p/{PACKAGE_NAME}/latest` . If `v0.15.1` is the latest version of package base, both `/p/base/v0.15.1` (versioned URL) and `/p/base/latest` (latest URL) exist.

- The package search results now link to the latest version of the respective packages. 

- The latest version of a package is also now the newest version that does not have the `avoid-version` flag set in its opam file.

- The version select dropdown now contains a "latest" option, in addition to the individual versions of the package.

- `/p/{PACKAGE_NAME}` redirects to `/p/{PACKAGE_NAME}/latest`.

- `/p/{PACKAGE_NAME}/docs` redirects to `/p/{PACKAGE_NAME}/latest/docs`. 

- The canonical link should be the versioned url, as the latest will most likely be periodically updated when new versions get released.


sabine:
Nice work! This patch makes it so that when we are on a URL `/p/PACKAGE_NAME/latest`, we can go to the package documentation and navigate within the package documentation while staying on the `latest`-URL

Here's a few things to improve:
* When I go back to the overview page of the package, it links to the specific version here:
Screenshot from 2023-01-12 12-18-56
* package search leads to the versioned URL of the package, not to the `latest`-URL. I think it would make sense to add this behaviour to the patch.

src/ocamlorg_web/lib/router.ml
```diff
@@ -70,10 +70,10 @@ let package_route t =
         (Url.package_with_version ~hash:":hash" ":name" ":version")
         ((Handler.package_versioned t) Handler.Universe);
       Dream.get
-        (Url.package_doc ":name" ":version" ~page:"**")
+        (Url.package_doc ":name" ":version" ~page:"**" ~is_latest_url:false)
         ((Handler.package_doc t) Handler.Package);
       Dream.get
-        (Url.package_doc ~hash:":hash" ~page:"**" ":name" ":version")
+        (Url.package_doc ~hash:":hash" ~page:"**" ":name" ":version" ~is_latest_url:false)
```
sabine: we could make `is_latest_url` an optional parameter (function parameter ?-syntax), so that there are no changes to `router.ml`.

src/ocamlorg_frontend/layouts/package_layout.eml
```diff
@@ -22,7 +22,7 @@ Layout.render
       <div class="flex justify-between flex-col md:flex-row border-b border-gray-200">
         <div class="flex flex-col items-baseline mb-6">
           <h4 class="font-bold md:whitespace-nowrap flex items-center">
-            <a class="flex items-center" href="<%s Url.package_with_version package.name package.version %>">
+            <a class="flex items-center" href="<%s Url.package package.name %>">
```
sabine: What happens when we are on a versioned URL of the package and want to go back to the versioned overview page? I think that changing the link to `Url.package package.name` is not the behaviour the user would expect here.

src/ocamlorg_web/lib/handler.ml
```diff
@@ -396,20 +387,29 @@ let package_docs t req =
...
+     Ocamlorg_package.get_package t name version
   in
-  let package = Ocamlorg_package.get_package t name version in
   match package with
   | None -> not_found req
   | Some package ->
       let open Lwt.Syntax in
+      let _version = Ocamlorg_package.version package in
+      let is_latest_url = if version_from_url = "latest" then 
+        true else false in
```
```suggestion
      let is_latest_url = version_from_url = "latest" in
```

src/ocamlorg_web/lib/handler.ml
```diff
@@ -452,18 +452,27 @@ let package_versioned t kind req =
...
+  let version_from_url = Dream.param req "version"
   in
-  let package = Ocamlorg_package.get_package t name version in
+  let package = if version_from_url = "latest" then 
+    Ocamlorg_package.get_package_latest t name
+  else 
+    let version = Ocamlorg_package.Version.of_string @@ version_from_url in
+     Ocamlorg_package.get_package t name version in
+  (* let is_latest_url = if version_from_url = "latest" then 
+      in *)
```
```suggestion
```

remove unnecessary comment

src/ocamlorg_web/lib/handler.ml
```diff
@@ -452,18 +452,27 @@ let package_versioned t kind req =
...
   match package with
   | None -> not_found req
   | Some package -> (
       let open Lwt.Syntax in
+      let is_latest_url = if version_from_url = "latest" then 
+          true else false 
+    in
```
```suggestion
      let is_latest_url = version_from_url = "latest" in
```

sabine:
Alright, so I think to get to the point where we can merge, we additionally need the ability to go to the "latest" URL via the version dropdown.

There is also the idea to remove "latest" from the URL entirely (so that we have a URL scheme where `/p/{PACKAGE_NAME}/` renders the latest version of the package), and I agree that this looks nice. However, if someone had a package with a version named, e.g. `"doc"`, then that would result in conflicting URLs `/p/{PACKAGE_NAME}/doc`.

src/ocamlorg_web/lib/handler.ml
```diff
@@ -371,20 +371,11 @@ let packages_search t req =
...
-  let version = find_default_version name in
-  match version with
-  | Some version ->
-      let version = Ocamlorg_package.Version.to_string version in
-      let target = Ocamlorg_frontend.Url.package_with_version package version in
-      Dream.redirect req target
-  | None -> not_found req
+  let version_string = "latest" in
+  let target = Ocamlorg_frontend.Url.package_with_version package version_string in
+    Dream.redirect req target
```
sabine: Can be simplified as

```suggestion
  let target = Ocamlorg_frontend.Url.package_with_version package "latest" in
    Dream.redirect req target
```

src/ocamlorg_web/lib/handler.ml
```diff
@@ -396,20 +387,29 @@ let package_docs t req =
...
+  else 
+    let version = Ocamlorg_package.Version.of_string @@ version_from_url in
+     Ocamlorg_package.get_package t name version
   in
-  let package = Ocamlorg_package.get_package t name version in
   match package with
   | None -> not_found req
   | Some package ->
       let open Lwt.Syntax in
+      let _version = Ocamlorg_package.version package in
```
sabine: underscore here means you can remove this line, since you never use the variable

src/ocamlorg_web/lib/handler.ml
```suggestion
      let is_latest_url = version_from_url = "latest" in
```

sabine (commented):
almost there :+1:

src/ocamlorg_web/lib/handler.ml
```diff
@@ -396,20 +386,27 @@ let package_docs t req =
   match version_opt with
   | None -> not_found req
   | Some version ->
-      let version = Ocamlorg_package.Version.to_string version in
-      let target = Ocamlorg_frontend.Url.package_doc package version in
+        let version_string = if Ocamlorg_package.is_latest_version t name version then 
+         "latest" else
+          Ocamlorg_package.Version.to_string version in
+      let target = Ocamlorg_frontend.Url.package_doc package version_string ~is_latest_url:false in
       Dream.redirect req target
```
sabine: Sorry I overlooked this last time. See the comment above the function `package_docs`. It says

```
(** Redirect any URL with suffix /p/PACKAGE/docs to the latest documentation for
    PACKAGE. *)
```

So I would expect that we can simplify this, similar to how `Url.package` was simplified.

Though, I think, this function and the corresponding route could probably be removed altogether. :thinking: Oh well, that would be the topic for another PR.

SaySayo: I think I am misunderstanding your comment. Do you mean to take out the `version_string` variable and just get the latest package with the Url.package function? Cause if I do that, then we dont use the version gotten from the match case

sabine: Ahh, sorry, I didn't mean `Url.package`. I meant the `package` handler function just above this function. It is much shorter and much simpler because we always redirect to the `latest`-URL.

cuihtlauac:
@SaySayo Could you resolve the conflicts which are listed?

SaySayo:
I have resolved the listed conflicts cc @cuihtlauac @sabine

src/ocamlorg_web/lib/handler.ml
```diff
@@ -451,19 +448,26 @@ let package_versioned t kind req =
       in
       Dream.html
         (Ocamlorg_frontend.package_overview ~documentation_status ~readme
+           ~readme_title ~dependencies ~rev_dependencies ~homepages ~source
+           ~changes_filename ~license_filename ~is_latest_url package_meta)
```
cuihtlauac, line 452:
```suggestion
           ~changes_filename ~license_filename ~is_latest_url package_meta
```

cuihtlauac (changes requested):

cuihtlauac:
@SaySayo, could you use `make fmt` to format the code, otherwise CI build fails

src/ocamlorg_web/lib/handler.ml
```diff
@@ -373,7 +373,7 @@ let packages_search t req =
 
 let package _t req =
   let package = Dream.param req "name" in
-  let target = Ocamlorg_frontend.Url.package_with_version package "latest" in
+  let target = Ocamlorg_frontend.Url.package package in
     Dream.redirect req target
```
sabine: This would give an infinite redirect. The handler is redirecting to its own URL.

src/ocamlorg_frontend/pages/package_overview.eml
```diff
@@ -93,6 +119,22 @@ Package_layout.render
...
+        <div class="flex flex-col mt-9 space-y-4 text-body-400">
+            <% (match changes_filename with Some changes_filename -> %>
+            <a href="<%s Url.package_doc package.name package.version ~is_latest_url ~page:(changes_filename ^ ".html") %>" class="flex items-center hover:text-primary-600">
+                <svg xmlns="http://www.w3.org/2000/svg" class="h-4 w-4 mr-2 inline-block" fill="none" viewBox="0 0 24 24"
+                    stroke="currentColor">
+                    <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2"
+                        d="M9 12h6m-6 4h6m2 5H7a2 2 0 01-2-2V5a2 2 0 012-2h5.586a1 1 0 01.707.293l5.414 5.414a1 1 0 01.293.707V19a2 2 0 01-2 2z" />
+                </svg>
+                </a>
+
```
cuihtlauac:
```suggestion
        <% | None -> ()); %>
```

src/ocamlorg_web/lib/handler.ml
```diff
@@ -451,19 +452,25 @@ let package_versioned t kind req =
       in
       Dream.html
         (Ocamlorg_frontend.package_overview ~documentation_status ~readme
-           ~readme_title ~dependencies ~rev_dependencies ~conflicts ~homepages
-           ~source ~changes_filename ~license_filename package_meta)
+           ~readme_title ~dependencies ~rev_dependencies ~homepages ~source
+           ~changes_filename ~license_filename ~is_latest_url package_meta)
```
cuihtlauac:
```suggestion
           ~readme_title ~dependencies ~rev_dependencies ~conflics ~homepages
           ~source ~changes_filename ~license_filename ~is_latest_url
           package_meta)
```

cuihtlauac (changes requested):

cuihtlauac:
I have filed two change requests to fix compilation

src/ocamlorg_web/lib/handler.ml
```diff
@@ -365,20 +365,10 @@ let packages_search t req =
...
-  let name = Ocamlorg_package.Name.of_string package in
-  let version = find_default_version name in
-  match version with
-  | Some version ->
-      let version = Ocamlorg_package.Version.to_string version in
-      let target = Ocamlorg_frontend.Url.package_with_version package version in
-      Dream.redirect req target
-  | None -> not_found req
+  let target = Ocamlorg_frontend.Url.package package in
+  Dream.redirect req target
```
sabine, line 371: this is an infinite redirect. It should redirect to the "latest"-URL of the package

src/ocamlorg_web/lib/handler.ml
```diff
@@ -390,22 +380,33 @@ let package_docs t req =
...
-      let target = Ocamlorg_frontend.Url.package_doc package version in
+      let version_string =
+        if Ocamlorg_package.is_latest_version t name version then "latest"
+        else Ocamlorg_package.Version.to_string version
+      in
+      let target =
+        Ocamlorg_frontend.Url.package_doc package version_string
+          ~is_latest_url:false
+      in
       Dream.redirect req target
```
sabine: this can be simplified, since all this function needs to do is to redirect to the latest-URL of the package documentation.

src/ocamlorg_web/lib/handler.ml
sabine: this is an infinite redirect. It should redirect to the latest-URL of the package

src/ocamlorg_web/lib/handler.ml
SaySayo: It doesn't need to just redirect to latest, its supposed to redirect to latest if the user wants to go to latest OR versioned if thats what the user wants, thats why there's the if then else statement

cuihtlauac:
In `handler.ml`, the function `Url.package_doc` is passed the parameter `"latest"`. In other call sites, it is passed the parameter `is_latest_url`. This is confusing. 

To me, the version parameter should be optional. If provided, it should be a string representing a version; if absent, it should mean "latest".

src/ocamlorg_frontend/layouts/package_layout.eml
```diff
@@ -24,19 +25,22 @@ Layout.render
...
+            <select id="version" name="version" onchange="location = this.value;" class="appearance-none cursor-pointer py-0 pt-1 border rounded-md border-transparent pr-8"> 
+              <option value="/p/<%s package.name %>/<%s "latest" %>">
+              <%s "latest" %>
+              </option>
               <% package.versions |> List.iter begin fun item -> %>
-              <option value="/p/<%s package.name %>/<%s item %>" <%s if item = package.version then "selected" else "" %>>
+              <option value="/p/<%s package.name %>/<%s item %>"<%s if not is_latest_url && item = package.version then "selected" else "" %>>
                 <%s item %>
-              </option>
+              </option>
```
cuihtlauac:
```suggestion
              </option>
```

src/ocamlorg_frontend/layouts/package_layout.eml
```diff
@@ -50,4 +54,4 @@ Layout.render
       </div>
     </div>
   </div>
-</div>
+</div>
```
cuihtlauac:
```suggestion
</div>

```

src/ocamlorg_frontend/pages/package_overview.eml
```diff
@@ -296,4 +298,4 @@ Package_layout.render
     });
 
 })));
-</script>
+</script>
```
cuihtlauac:
```suggestion
</script>

```

src/ocamlorg_frontend/url.ml
```diff
@@ -8,8 +8,10 @@ let package_docs v = "/p/" ^ v ^ "/doc"
 let package_with_version ?hash v version =
   with_hash hash ^ "/" ^ v ^ "/" ^ version
 
-let package_doc ?hash ?(page = "index.html") v version =
-  with_hash hash ^ "/" ^ v ^ "/" ^ version ^ "/doc/" ^ page
+let package_doc ?hash ?(page = "index.html") ?(is_latest_url = false) v version
+    =
+  if is_latest_url then with_hash hash ^ "/" ^ v ^ "/latest/doc/" ^ page
+  else with_hash hash ^ "/" ^ v ^ "/" ^ version ^ "/doc/" ^ page
```
cuihtlauac:
```suggestion
  let version = if is_latest_url then "latest" else version in
  with_hash hash ^ "/" ^ v ^ "/" ^ version ^ "/doc/" ^ page
```

src/ocamlorg_package/lib/ocamlorg_package.ml
```diff
@@ -386,6 +386,11 @@ let latest_documented_version t name =
   | None -> Lwt.return None
   | Some vlist -> aux (List.rev vlist)
 
+let is_latest_version t name version =
+  match get_package_versions t name with
+  | None -> false
+  | Some vlist -> List.hd vlist == version
```
cuihtlauac:
```suggestion
  | Some (lastest :: _) -> latest = version
  | _ -> assert false
```

src/ocamlorg_package/lib/ocamlorg_package.mli
```diff
@@ -156,6 +156,8 @@ val get_package : state -> Name.t -> Version.t -> t option
 val latest_documented_version : state -> Name.t -> Version.t option Lwt.t
 (** Find the latest documented version of a package. **)
 
+val is_latest_version : state -> Name.t -> Version.t -> bool
```
cuihtlauac, line 169: Add documentation

cuihtlauac (changes requested):
I like the overall design approach to be described

SaySayo:
I have made the suggested changes. cc @cuihtlauac

cuihtlauac (changes requested):
Consider the diff with respect to the branch https://github.com/ocaml/ocaml.org/tree/latest-with-version as suggestions

src/ocamlorg_web/lib/handler.ml
```diff
@@ -365,47 +365,35 @@ let packages_search t req =
...
-  match version_opt with
-  | None -> not_found req
-  | Some version ->
-      let version = Ocamlorg_package.Version.to_string version in
-      let target = Ocamlorg_frontend.Url.package_doc package version in
-      Dream.redirect req target
+  let target = Ocamlorg_frontend.Url.package_doc package in
+  Dream.redirect req target
 
 let installer req = Dream.redirect req Ocamlorg_frontend.Url.github_installer
```
cuihtlauac, line 376:
```suggestion
let to_package t name version =
  if version = "latest" then
    Ocamlorg_package.get_package_latest t name
  else
    let version = Ocamlorg_package.Version.of_string version in
    Ocamlorg_package.get_package t name version
```

src/ocamlorg_web/lib/handler.ml
```diff
@@ -365,47 +365,35 @@ let packages_search t req =
...
-  let version =
-    Ocamlorg_package.Version.of_string @@ Dream.param req "version"
+  let version_from_url = Dream.param req "version" in
+  let package =
+    if version_from_url = "latest" then
+      Ocamlorg_package.get_package_latest t name
+    else
+      let version = Ocamlorg_package.Version.of_string @@ version_from_url in
+      Ocamlorg_package.get_package t name version
   in
```
cuihtlauac:
```suggestion
  let package = to_package t name version_from_url in
```

src/ocamlorg_web/lib/handler.ml
```diff
@@ -452,33 +440,44 @@ let package_versioned t kind req =
...
```
cuihtlauac:
```suggestion
  let package = to_package t name version_from_url in
```

src/ocamlorg_web/lib/handler.ml
```diff
@@ -452,33 +440,44 @@ let package_versioned t kind req =
...
         | Universe -> `Universe (Dream.param req "hash")
       in
       let path = (Dream.path [@ocaml.warning "-3"]) req |> String.concat "/" in
       let root =
+        let version =
+          if is_latest_url then None
+          else Some (Ocamlorg_package.Version.to_string version)
+        in
         let make =
           match kind with
```
cuihtlauac:
```suggestion
        let hash = match kind with `Package -> None | `Universe u -> Some u in
        Ocamlorg_frontend.Url.package_doc ?hash ~page:"" ?version
          (Ocamlorg_package.Name.to_string name)
```

src/ocamlorg_web/lib/handler.ml
```diff
@@ -452,33 +440,44 @@ let package_versioned t kind req =
...
-          | `Package -> Ocamlorg_frontend.Url.package_doc ?hash:None ~page:""
-          | `Universe u -> Ocamlorg_frontend.Url.package_doc ~hash:u ~page:""
+          | `Package ->
+              Ocamlorg_frontend.Url.package_doc ?hash:None ~page:"" ?version
+          | `Universe u ->
+              Ocamlorg_frontend.Url.package_doc ~hash:u ~page:"" ?version
```
cuihtlauac:
```suggestion
```

src/ocamlorg_web/lib/handler.ml
```diff
@@ -452,33 +440,44 @@ let package_versioned t kind req =
...
         in
-        make
-          (Ocamlorg_package.Name.to_string name)
-          (Ocamlorg_package.Version.to_string version)
+        make (Ocamlorg_package.Name.to_string name)
       in
```
cuihtlauac:
```suggestion
```

cuihtlauac (changes requested):
Minor code cleanup
See: https://github.com/ocaml/ocaml.org/tree/latest-with-version

cuihtlauac (approved):
Looks good to me

src/ocamlorg_package/lib/ocamlorg_package.ml
```diff
@@ -386,6 +386,12 @@ let latest_documented_version t name =
   | None -> Lwt.return None
   | Some vlist -> aux (List.rev vlist)
 
+let is_latest_version t name version =
+  match get_package_versions t name with
```
kit-ty-kate: Unless I’ve missed another List.rev somewhere else I think this is missing. `Map.bindings` returns the list in increasing version number but you need the decreasing version number here
```suggestion
  match List.rev (get_package_versions t name) with
```

tmattio:
From offline conversations, here is a list of things to consider for this PR:

- Could change the dropdown latest option to include the version number, e.g. `Latest (v0.15.0)` (nice to have)
- Remove the version of the package in the title and description (change required)
- Latest shouldn’t be latest (we don’t want to include rc, alpha, previews, etc.). We want to be consistent with opam on this, there's probably a function to return the latest "preferred" version of a package. (change required)
- What do we do about the canonical? We don’t want all the versions to be index by google (when searching, only return the latest version) (out of scope)
- Need to fix an issue with the htmx logic that causes the canonical not to change (cc @sabine). Also should check other fields in the head section. (out of scope)

src/ocamlorg_package/lib/ocamlorg_package.ml
SaySayo: The expected behaviour is to get the latest version of the package which is assumed to be the first of the list. Since we are trying to use pre-released versions as latest but also want to make with the most recent stable release, will reversing the entire list fix that? as the entire list of version would then have the oldest releases at the top and the most recent one at the bottom

SaySayo:
Hi @tmattio, to be clear, am I removing the title and description for just the latest url? or for all the package version?

sabine:
Since we will never navigate by htmx on the version dropdown, update of metadata on htmx-navigation is irrelevant to this PR. Similarly, search engines do not navigate by htmx. However, I agree that it would be nice to keep the metadata in the HTML page head and the title in sync: https://github.com/ocaml/ocaml.org/issues/892

sabine:
> Latest shouldn’t be latest (we don’t want to include rc, alpha, previews, etc.). We want to be consistent with opam on this, there's probably a function to return the latest "preferred" version of a package. (change required)

Makes sense. Indeed, if opam gives us a notion of "latest", we should use that and extend `Ocamlorg_package` to work with it.

src/ocamlorg_package/lib/ocamlorg_package.ml
kit-ty-kate: I’m not sure to get that. Have you tested this function? I haven’t but just reading `get_package_versions` it seems to be the inverse of what you want (the first element of the list is the oldest version not the latest)

data/tutorials/gs_00_up_and_running.md
```diff
@@ -108,6 +108,8 @@ On Unix, it's essential to initialise opam, which will (if needed) install the O
 
 This step is done automatically by the Platform Installer, [see below](#up-and-running-with-the-platform-installer).
 
+This step is done automatically by the Platform Installer, [see below](#up-and-running-with-the-platform-installer).
```
christinerose: Is this paragraph repeated above? It looks like it's in there twice, unless I'm misreading it.

christinerose (approved):

cuihtlauac:
@SaySayo, I think there's been a small mistake. The file `data/tutorials/gs_00_up_and_running.md` should not be part of your changeset

src/ocamlorg_web/lib/handler.ml
```diff
@@ -490,7 +478,7 @@ let package_doc t kind req =
                  | `ClassType s -> s)
             |> String.concat "."
           in
-          let title =
+          let _title =
```
sabine: `_title` is never used.

Two possible ways to deal with that:
1. move the logic that decides the title of the page depending on is_latest_version over here
2. remove `_title` and keep the title rendering logic in the template

src/ocamlorg_frontend/pages/package_documentation.eml
```diff
@@ -41,60 +44,40 @@ let path_page = match str_path |> String.concat "/" with
...
-    window.location = evt.detail.requestConfig.path;
-  });
-  document.body.addEventListener('htmx:historyCacheMissError', (evt) => {
-    window.location = evt.detail.path;
-  });
-  document.body.addEventListener('htmx:onLoadError', (evt) => {
-    window.location = evt.detail.requestConfig.path;
-  });
-};
-</script>
```
sabine: Don't remove this. We need to keep this.

src/ocamlorg_frontend/pages/package_documentation.eml
```diff
@@ -41,60 +44,40 @@ let path_page = match str_path |> String.concat "/" with
...
     </div>
-    <div class="hidden xl:flex top-0 sticky h-screen flex-col w-60 overflow-auto right-sidebar" id="htmx-right-sidebar">
-      <%s! Toc.render toc %>
+    <div class="hidden xl:flex top-0 sticky h-screen">
+      <div class="flex-col w-60">
+        <div class="h-screen overflow-scroll right-sidebar pt-6">
+          <div class="font-semibold text-gray-500 text-sm mb-4">ON THIS PAGE</div>
+          <%s! Toc.render toc %>
+        </div>
+      </div>
```
sabine: Keep this the same as on the main branch. Your patch shouldn't make changes here.

src/ocamlorg_frontend/pages/package_documentation.eml
```diff
@@ -41,60 +44,40 @@ let path_page = match str_path |> String.concat "/" with
...
     <div
       class="p-10 z-20 bg-white flex-col fixed h-screen overflow-auto lg:pr-0 lg:flex left-0 top-0 lg:sticky w-72 lg:p-0 lg:pt-6"
       x-show="sidebar" x-transition:enter="transition duration-200 ease-out"
       x-transition:enter-start="-translate-x-full" x-transition:leave="transition duration-100 ease-in"
       x-transition:leave-end="-translate-x-full">
-      <%s! sidebar ~str_path ~toc ~maptoc package %>
+      <%s! sidebar ~str_path ~toc ~maptoc ~is_latest_url package %>
     </div>
-    <div class="flex-1 z-0 z- min-w-0 lg:max-w-3xl lg:pt-6" id="htmx-content">
+    <div class="flex-1 z-0 z- w-full lg:max-w-3xl relative lg:pt-6">
```
sabine: Needs to be the same as main branch.

src/ocamlorg_frontend/pages/package_documentation.eml
```diff
@@ -41,60 +44,40 @@ let path_page = match str_path |> String.concat "/" with
...
 ~styles:["/css/main.css"; "/css/doc.css"] @@
-<div x-data="{ open: false, sidebar: window.innerWidth > 1024 && true, showOnMobile: false}" @resize.window="sidebar = window.innerWidth > 1024" x-on:close-sidebar="sidebar=window.innerWidth > 1024 && true">
+<div x-data="{ open: false, sidebar: window.innerWidth > 1024 && true, showOnMobile: false}" @resize.window="sidebar = window.innerWidth > 1024">
   <button :title='(sidebar? "close" : "open")+" sidebar"' aria-pressed="false" :aria-pressed="sidebar" class="flex items-center bg-primary-600 p-3 z-30 rounded-full text-white shadow-md bottom-20 fixed lg:hidden right-10"
     x-on:click="sidebar = ! sidebar">
     <%s! Icons.sidebar_menu "h-8 w-8" %>
     <span class="hidden md:flex font-semibold px-2">side menu</span>
   </button>
-  <div class="fixed z-10 inset-0 bg-background-dark-blue/20 backdrop-blur-sm lg:hidden"
-    :class='sidebar ? "" : "hidden"' aria-hidden="true" x-on:click="sidebar = ! sidebar"></div>
```
sabine: Dont remove.

src/ocamlorg_frontend/pages/package_documentation.eml
```diff
@@ -41,60 +44,40 @@ let path_page = match str_path |> String.concat "/" with
...
-~description:(Printf.sprintf "%s %s: %s" package.name package.version package.description)
+~is_latest_url
+~title:(if is_latest_url then (Printf.sprintf "%s %s · OCaml Package" package.name "latest") else (Printf.sprintf "%s %s · OCaml Package" package.name package.version))
+~description:(if is_latest_url then (Printf.sprintf "%s %s: %s" package.name "latest" package.description) else (Printf.sprintf "%s %s: %s" package.name package.version package.description))
 ~package
 ~path
-~canonical:(Url.package_doc package.name package.version ?page:path_page)
+~canonical:(Url.package_doc package.name ?version ?page:path_page)
 ~styles:["/css/main.css"; "/css/doc.css"] @@
-<div x-data="{ open: false, sidebar: window.innerWidth > 1024 && true, showOnMobile: false}" @resize.window="sidebar = window.innerWidth > 1024" x-on:close-sidebar="sidebar=window.innerWidth > 1024 && true">
```
sabine, line 51: Keep the same as main branch.

src/ocamlorg_frontend/pages/package_documentation.eml
sabine: The background overlay when the sidebar is open is still being removed in the patch. The rest looks reasonable to me.

src/ocamlorg_frontend/pages/package_documentation.eml
```diff
@@ -54,17 +58,17 @@ Package_layout.render
...
       class="p-10 z-20 bg-white flex-col fixed h-screen overflow-auto lg:pr-0 lg:flex left-0 top-0 lg:sticky w-72 lg:p-0 lg:pt-6"
       x-show="sidebar" x-transition:enter="transition duration-200 ease-out"
       x-transition:enter-start="-translate-x-full" x-transition:leave="transition duration-100 ease-in"
       x-transition:leave-end="-translate-x-full">
-      <%s! sidebar ~str_path ~toc ~maptoc package %>
+      <%s! sidebar ~str_path ~toc ~maptoc ~is_latest_url package %>
     </div>
-    <div class="flex-1 z-0 z- min-w-0 lg:max-w-3xl lg:pt-6" id="htmx-content">
+   <div class="flex-1 z-0 z- min-w-0 lg:max-w-3xl lg:pt-6" id="htmx-content">
       <div class="odoc prose prose-orange">
```
sabine:
```suggestion
    <div class="flex-1 z-0 z- min-w-0 lg:max-w-3xl lg:pt-6" id="htmx-content">
      <div class="odoc prose prose-orange">
```

sabine:
I found the reason why the patch kept segfaulting for me:

[Info.t version needs to be bumped in order to invalidate the package info cache (click here for a diff that also recommends another code simplification)](https://github.com/ocaml/ocaml.org/commit/82c05c4b8d6af4b6c89f83758b8ec0cc33fab49e).

Segmentation fault happened because `Info.t` in my package info cache didn't have the `flags` field.

sabine (approved):
LGTM

cuihtlauac (approved):

Issue #824, opened 2023-01-17
Improvement to the Multicore Example in Playground


sabine:
There were concerns about the multicore Fibonacci code example misrepresenting the skills of OCaml programmers, as it shows an inefficient implementation.

Suggestion from [discuss.ocaml.org](https://discuss.ocaml.org/t/looking-for-participants-for-user-survey-on-ocaml-org-package-documentation-and-learn-area/11128/18):
> Rename the function to `naive_fib` and point to this URL in a comment just above the function name?
> 
> [See: OPCEB chapter 8 on memoization for a more performant implementation](https://cs3110.github.io/textbook/chapters/ds/memoization.html#fibonacci)

cuihtlauac:
Relevant discussion on twitter: https://twitter.com/wicko3/status/1482042995138600960

cuihtlauac:
Possible duplicate of: https://github.com/ocaml/ocaml.org/issues/233

PR #831, opened 2023-01-17
minor changes to the playground example

sabine:
As it was brought up on discuss.ocaml.org that

(1) beginners should be exposed to a more performant implementation of the Fibonacci sequence, and
(2) experienced users may want a link to a more realistic example of parallel computation,

here we add both of them in a comment. Minor edits to reduce verbosity included.

Addresses https://github.com/ocaml/ocaml.org/issues/824.

### DIFF
--- a/playground/src/example.ml
+++ b/playground/src/example.ml
@@ -7,9 +7,10 @@ let adts =
   You don't need to install anything - just write your code
   and see the results appear in the Output panel.
 
-  It's also compatible with OCaml 5: you can use the new
-  Domain module and effect handlers. Below is a naive example
-  that calculates the Fibonacci sequence in parallel.
+  This playground is powered by OCaml 5 which comes with
+  support for shared-memory parallelism through domains and effects.
+  Below is some naive example code that calculates
+  the Fibonacci sequence in parallel.
   
   Happy hacking!
 *)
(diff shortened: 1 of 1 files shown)


playground/src/example.ml
```diff
@@ -32,4 +31,22 @@ let rec fib_par n d =
...
+  the Fibonacci sequence is this:
+
+  let fib n =
+    let rec loop n =
+      if n < 1 then (0, 1)
+      else
+        let x, y = loop (n-1) in
+        y, (x+y)
+    in
+    snd (loop n);;
```
cuihtlauac: Isn't a tail-recursive version preferable? Here is how I like to write it:
```ocaml
let rec fib m n i = if i < 1 then m else fib n (n + m) (i - 1)
let fib = fib 0 1
```
But there are so many good ways to write that kind of stuff :-)

cuihtlauac (changes requested):

playground/src/example.ml
sabine: changed example to this variant

cuihtlauac (approved):

sabine:
There might be a build step I am missing - the updated example doesn't show up at https://ocaml.org/play. :thinking:  @tmattio

PR #837, opened 2023-01-18
Ood: Add missing slug field

cuihtlauac:


sabine:
Does this also need to update the router to make the corresponding page available under the slug? :thinking:

Or wait... are these data items that don't need a slug because they aren't exposed under a URL?

Or do we add slugs for consistency's sake, for the case that these eventually get their own URL on ocaml.org?

cuihtlauac:
My title is wrong. The slug was generated in the `pp` function. With this PR, generation just takes place earlier. This allows the “data” type to be the same as the one in the template function string.

This is a preliminary step towards more automatic processing in an upcoming PR :-)

sabine (approved):
LGTM

Issue #838, opened 2023-01-18
OCaml Playground's large bundle size


mooreryan:
It's been mentioned a couple of other times, eg

- https://github.com/ocaml/ocaml.org/pull/106
- https://github.com/ocaml/ocaml.org/pull/536

But I'm wondering about the size of the JS for the playground.

- playground-v1.js: 3.5mb
- merlin-v1.js: 4.8mb
- worker-v1.js: 22.4mb

I get that jsoo bundles can be large and the toplevel is linking a lot of stuff, but I'm wondering about the caching and compression.  It doesn't appear that these JS files are being cached by the browser, or compressed over the wire.  The playground-v1.js is transferred each time I reload, then that (seems to) make a call to fetch the other two.

Just wondering if there is something preventing the caching/zipping on ocaml.org or if it is something wrong on my end.

(I tried both firefox and chrome (I can give versions if needed).)

sabine:
Agreed, there's multiple issues here:
1. large bundle size
2. lack of compression (fix in #840)
3. lack of caching (I opened #839 for that)

Keeping this issue open because the bundle size itself would be nice to optimize as well.

mooreryan:
Thanks for looking at this...assets down to less than 7mb transferred.  Appreciate it! (I saw it mentioned in the pull request as well, but figured I would mention here as well.)

sabine:
@mooreryan Thanks for reporting :+1: , there's so many little things like this that have a huge negative effect on people. So even if you think we might know from older issues: me and @cuihtlauac became maintainers only fairly recently, so we might not have seen it.

tmattio:
There's unfortunately not much we can do about the bundle size. What we can do, however, is compress the files that are stored in the filesystem so we don't compress them several times at runtime.

This is an optimization that's not related to the bundle size, so we'll open a new issue for that. Closing this for now.

PR #840, opened 2023-01-19
enable compression for playground assets

sabine:
As reported in https://github.com/ocaml/ocaml.org/issues/838, the compression middleware was not applied to the playground assets. This patch enables it.


cuihtlauac:
Looks awesome to me:
- Before: 30.8 MiB transferred
- After: 6.4 MiB transferred

cuihtlauac (approved):

PR #856, opened 2023-01-23
Playground: update example

cuihtlauac:
Follow-up from: https://github.com/ocaml/ocaml.org/pull/831

Playground needs to be built and assets committed to be visible


sabine (approved):

sabine:
Thank you!

PR #858, opened 2023-01-23
Update HACKING.md

cuihtlauac:
Reflect updated playground asset need to be git committed into
ocaml/ocaml.org as a PR

Follow-up from: https://github.com/ocaml/ocaml.org/pull/831


sabine (approved):
Thank you, updating the readme as well is very useful!

PR #864, opened 2023-01-25
remove wide parameter from layout, header is always wide

sabine:
The top navbar jumping between narrow and wide was bothersome. The top navbar is now always "wide". The `wide` parameter on the header and layouts has been removed.

Does not change the widths of the content areas. Every page defines its own container (regular or wide) according to the content it hosts. Reason: some of the layouts did not look so great when "stretched out" to wide.

|before|after|
|-|-|
|Screenshot 2023-01-25 at 18-00-02 The OCaml Community|Screenshot 2023-01-25 at 18-00-06 The OCaml Community|

### DIFF
--- a/src/ocamlorg_frontend/components/header.eml
+++ b/src/ocamlorg_frontend/components/header.eml
@@ -17,7 +17,6 @@ let menu_link
 
 let render
 ?(show_get_started=true)
-?(wide=false)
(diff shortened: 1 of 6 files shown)


cuihtlauac (approved):

cuihtlauac:
Looks good to me; much needed. Thanks @sabine

PR #865, opened 2023-01-25
Frontend: Icon Sharing

cuihtlauac:


sabine:
Currently, there are only 4 occurrences of `Bullet`, and no occurrences of `Illustration`.

I would refactor that as
```
let downloads cla =
  <svg xmlns="http://www.w3.org/2000/svg" class="<%s cla %>" fill="none" viewBox="0 0 24 24" stroke="currentColor">
    <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" id="downloads" d="M4 16v1a3 3 0 003 3h10a3 3 0 003-3v-1m-4-4l-4 4m0 0l-4-4m4 4V4" />
  </svg>
```
and remove `kind` altogether. The size and padding must be determined within the context the icon is used.

Then, replace all the duplicated svg icons with `<%s! Icon.(downloads classes) %>` (respectively).

In my experience, this gives the needed flexibility at the location where the icon is used while also not duplicating the SVG everywhere.

cuihtlauac:
Sorry @sabine , I forgot to turn this into a draft. Many refactorings will come.

I will do what you have in mind, it's planned. However, I need to begin by identify each svg snippet inlined into the code and name it accordingly. Once this is done, I'll do the sharing and at the end, I'll do the simplification refactorings automatically

sabine (approved):

sabine:
Rebased. I think we should merge. Even if there's a small mess-up somewhere (which I don't think there is), we can fix that. Quality of life (for the developer) when adding icons is so much improved by this patch.

cuihtlauac:
Can't you wait until Monday? Never put in production on Fridays :-)

sabine:
Ok I'll avoid icon PRs till monday, that's fair.

cuihtlauac:
Thanks for merging! 👍

Issue #820, opened 2023-01-16
Bad redirections in URLs with trailing slashes


TheLortex:
The trailing slash middleware https://github.com/ocaml/ocaml.org/pull/170 introduced problems in some situations:
- https://ocaml.org//github.com/ocaml/ocaml.org/pull/170/ redirects to an arbitrary URL
- https://ocaml.org/packages/////////////////////////////////////////////////////////////// leads to too many redirections

Thanks to @reynir who identified the issue in mirage.io

cuihtlauac:
A bit of context:

1. Initial issue: https://github.com/ocaml/ocaml.org/issues/166
2. Faulty PR: https://github.com/ocaml/ocaml.org/pull/170
3. Issue in mirage-www: https://github.com/mirage/mirage-www/issues/772
4. PR in mirage-www: https://github.com/mirage/mirage-www/pull/788
5. New (this) issue: https://github.com/mirage/mirage-www/pull/788#issuecomment-1373455519

Summary: the code introduced in https://github.com/ocaml/ocaml.org/pull/170 is faulty. It went unnoticed until it was copied into mirage-www. Thanks to @TheLortex and @reynir for reporting this.

Questions:

1. @TheLortex are you working on fixing this, or do you want to fix it first on our side?
2. Isn't the _copying_ suggesting this should go into dream?

TheLortex:
@cuihtlauac I'm not actively working on it, so I would be happy if you can fix it on you side. 

> Isn't the copying suggesting this should go into dream?

That's a complex matter because we don't know if the canonical path for given route should end by a slash or not. _not ending by a slash_ every path is an opinionated decision so it might have difficulties going in dream. But I agree that would be useful.

reynir:
I think it could make sense to upstream functions to Dream that help deal with the somewhat complex semantics of the `Location` header, at least. I agree about getting it into dream could be difficult - I personally don't like it much as it can break semantics of relative links in the html and I made a library with a different approach (though still suffering from the same bug) https://github.com/reynir/dream-normalize-route. Another point against upstreaming is that dream hasn't seen seen any activity in a while.

cuihtlauac:
Fixed by https://github.com/ocaml/ocaml.org/pull/866

PR #866, opened 2023-01-27
Drop extra slashes

cuihtlauac:
Handle the request unchanged if the target seems canonical, otherwise:

  * Turns repeated slashes into a single one
  * Drop the last slash
  * Redirect to the resulting target

Issue: https://github.com/ocaml/ocaml.org/issues/820

Note: Is `Dream.split_target` bugged? When passed a string beginning with exactly two slashes, `split_target` drops everything before the third slash. See: https://github.com/aantron/dream/issues/248 As a workaround; three slashes are prepended to all targets

### DIFF
--- a/src/ocamlorg_web/lib/middleware.ml
+++ b/src/ocamlorg_web/lib/middleware.ml
@@ -1,11 +1,12 @@
 let no_trailing_slash next_handler request =
-  let target = Dream.target request in
-  match target with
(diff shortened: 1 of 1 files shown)


cuihtlauac:
Dream issue may be upstream: https://github.com/mirage/ocaml-uri/issues/167

PR #867, opened 2023-01-27
Ood: Use ppx_show to generate pretty printers

cuihtlauac:


cuihtlauac:
One possible way to check if this is not introducing regressions is to compare the contents of `_build/default/src/ocamlorg_data` after `make build` in this branch (`ood-ppx-show`) and `main`. However, the generated OCaml sources would have to be formatted using `ocamlformat`. I have some code doing this, if needed I can turn it into a PR.

Manual observation on `academic_institution.ml` of the outlined processes shows:
* String literals in `main` are surrounded by `{js|` and `|js}` while regular double quotes are used in `ood`
* Newlines remain in string literals in `ood-ppx-show` while they are stripped of in `main`

Otherwise, it is same

sabine:
I like this, it removes a lot of noise from the files and makes it more obvious to newcomers how to make changes to the actual logic.

sabine (approved):

sabine:
I looked at the new and the old `ocamlorg_data` folders.

Observations:

In the old `ocamlorg_data`, `workshop.ml` is already using some quotes `"`, instead of `{js|` and `|js}`, and `workflow.ml` uses only quotes. For `workflow.ml`, the content even appears identical (modulo formatting).

All the HTML is all properly escaped (`\"` and `&quot;`), so there are no problems to be expected there, when using quotes.

To me this is a strong enough signal that we should go ahead. The worst that can happen is that we revert and fix some data - the gain in consistency and clarity is well worth it.

cuihtlauac:
Thanks, @sabine for this. It's better to have a second opinion for such a merge. Let's proceed after I resolve the conflict

PR #870, opened 2023-01-31
Fix the broken link

diohabara:
Previous link(`#Runlengthencodingofalistdirectsolutionmedium`) was directed to <https://ocaml.org/problems#Runlengthencodingofalistdirectsolutionmedium>, but actual link to the run-length problem was <https://ocaml.org/problems#10>.


cuihtlauac:
Nice catch. Thanks a lot

PR #878, opened 2023-02-01
Upgrade package documentation to odoc 2.2.0

### LINKED ISSUE #848: Bring back missing preambles (content up to first heading) on package documentation
- [x] upstream patched https://github.com/ocaml/odoc/pull/908 (the `--content-only` option would omit the preambles, and ocaml.org is currently reconstructing metadata from the URL).
- [x] ocaml-docs-ci pipeline has built documentation for all packages with odoc 2.2.0
- [x] patch ocaml.org (patch needs to be rebased, but otherwise good to go)

sabine:
This patch makes ocaml.org consume the new `.html.json` format (HTML preamble/content+metadata) produced by `odoc html --as-json`, starting from version 2.2.0.

Patch attempts to fetch the new documentation from https://docs-data.ocaml.org/current/ (hardcoded URL), and, if that is not found, falls back to the old documentation in the regular, configurable location (currently https://docs-data.ocaml.org/live/).

The included fallback will be removed ASAP after it is clear that the breaking documentation format upgrade went fine.

Can be tested together with voodoo by uncommenting the hardcoded localhost path.

Resolves https://github.com/ocaml/ocaml.org/issues/848.

Prerequisite PR on voodoo: https://github.com/ocaml-doc/voodoo/pull/52 (upgrades to odoc 2.2.0)

### DIFF
--- a/src/ocamlorg_package/lib/module_map.ml
+++ b/src/ocamlorg_package/lib/module_map.ml
@@ -1,19 +1,20 @@
 module String_map = Map.Make (String)
 
 type kind =
-  | Library
   | Module
   | Page
   | Leaf_page
   | Module_type
-  | Parameter
+  | Parameter of int
+  | OLDParameter (* FALLBACK, DEPRECATED*)
   | Class
   | Class_type
   | File
 
 let prefix_of_kind = function
   | Module_type -> "module-type-"
-  | Parameter -> "argument-"
+  | Parameter i -> "argument-" ^ Int.to_string i ^ "-"
+  | OLDParameter -> "argument-" (* FALLBACK, DEPRECATED*)
   | Class -> "class-"
   | Class_type -> "class-type-"
   | _ -> ""
@@ -57,10 +58,16 @@ let kind_of_yojson v =
   | "module" -> Module
   | "leaf-page" -> Leaf_page
   | "module-type" -> Module_type
-  | "argument" -> Parameter
   | "class" -> Class
   | "class-type" -> Class_type
   | "file" -> File
+  | "argument" ->
+      OLDParameter
+      (* TODO: DEPRECATED, REMOVE when we no longer need to fall back to the old
+         format*)
+  | s when String.starts_with ~prefix:"argument-" s ->
+      let i = List.hd (List.tl (String.split_on_char '-' s)) in
+      Parameter (int_of_string i)
   | _ -> raise (Type_error ("Variant not supported", v))
 
 let rec module_of_yojson ?parent v : Module.t =
--- a/src/ocamlorg_package/lib/ocamlorg_package.ml
+++ b/src/ocamlorg_package/lib/ocamlorg_package.ml
@@ -456,17 +456,37 @@ let changes_filename ~kind t =
   | None -> None
   | Some (filename, _content) -> Some filename
 
-let documentation_status ~kind t =
+type documentation_status = Success | Failure | Unknown
+
+let old_documentation_status ~kind t =
+  let open Lwt.Syntax in
+  let old_package_url =
+    old_package_url ~kind (Name.to_string t.name) (Version.to_string t.version)
+  in
+  let url = old_package_url ^ "status.json" in
+  let+ content = http_get url in
+  match content with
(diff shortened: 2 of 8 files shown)


src/ocamlorg_package/lib/module_map.ml
```diff
@@ -57,11 +58,18 @@ let kind_of_yojson v =
...
-  | _ -> raise (Type_error ("Variant not supported", v))
+  | "argument" ->
+      OLDParameter
+      (* TODO: DEPRECATED, REMOVE when we no longer need to fall back to the old
+         format*)
+  | s ->
+      if String.starts_with ~prefix:"argument-" s then
+        let i = List.hd (List.tl (String.split_on_char '-' s)) in
+        Parameter (int_of_string i)
+      else raise (Type_error ("Variant not supported", v))
```
cuihtlauac:
```suggestion
  | s when String.starts_with ~prefix:"argument-" s ->
      let i = List.hd (List.tl (String.split_on_char '-' s)) in
      Parameter (int_of_string i)
  | _ -> raise (Type_error ("Variant not supported", v))
```

src/ocamlorg_package/lib/ocamlorg_package.ml
```diff
@@ -219,44 +245,111 @@ module Documentation = struct
...
+          ("breadcrumbs", `List json_breadcrumbs);
+          ("toc", `List json_toc);
+          ("preamble", `String preamble);
+          ("content", `String content);
+        ] ->
+        let breadcrumbs =
+          List.tl
+            (List.tl
+               (List.tl
+                  (List.tl (List.map breadcrumb_from_json json_breadcrumbs))))
```
cuihtlauac:
```suggestion
        let breadcrumbs = match List.map breadcrumb_from_json json_breadcrumbs with
        | _ :: _ :: _ :: _ :: breadcrumbs -> breadcrumbs
        | _ -> failwith "Not enough bread"
```

src/ocamlorg_package/lib/ocamlorg_package.ml
```diff
@@ -202,14 +202,40 @@ let featured_packages t = t.featured_packages
...
+    | "module" -> Module
+    | "module-type" -> ModuleType
+    | "class" -> Class
+    | "class-type" -> ClassType
+    | "file" -> File
+    | _ ->
+        if String.starts_with ~prefix:"argument-" s then
+          let i = List.hd (List.tl (String.split_on_char '-' s)) in
+          Parameter (int_of_string i)
+        else raise (Invalid_argument ("kind not recognized: " ^ s))
```
cuihtlauac, line 223: Isn't this almost the same as `Module_map.kind_of_json`? Isn't some sharing possible?

cuihtlauac (changes requested):
This is very impressive.

I've inlined a couple of minor comments. 

I need some more time to digest the whole thing

src/ocamlorg_package/lib/module_map.ml
sabine: ty, I didn't know about the `when` syntax! Very cool!

src/ocamlorg_package/lib/ocamlorg_package.ml
sabine: After we update how voodoo generates `package.json`, which contains the module map, sharing makes sense. Before that, there are subtle differences and keeping them separate is the cleaner route.

src/ocamlorg_package/lib/ocamlorg_package.ml
sabine: ty!

src/ocamlorg_package/lib/ocamlorg_package.ml
sabine: Sorry, I was being too hasty here, yeah, looks like they are now actually the same, except for that one fallback case. I tried not to do too many cleanups in this patch to keep the diff small. :see_no_evil:

Definitely looking forward to refactor and clean that up. Probably best to do it at the same time as removing the fallback, though.

Issue #821, opened 2023-01-16
Problems: Add Copy Code Button


cuihtlauac:
It would be nice to have a button allowing to copy the problem code fragments, either in statement and solution.

sabine:
What needs to be done:
1. extend `problem.ml` to provide `statement_ml` and `solution_ml` (or maybe there are better names) string fields that contain the strings that should be copied to clipboard. I think that means stripping the markdown code blocks and turning all non-code text into comments because it would be nice to have the full problem statement copied, not only the code block.
2. add a button in the UI to copy to clipboard.

sabine:
After getting some feedback from the community it became clear that it is not that clear what should happen here: Some users would greatly prefer that exercises are also available for offline consumption, e.g. in a git repository or with a command line application that also checks that the exercise is solved correctly.

IMO it would be nice to have an "open in playground" button because the same code that is copied to the playground could also be made available separately.

PR #885, opened 2023-02-06
Add a copy button to Exercises

SaySayo:
This is a draft PR for (#821). There's currently a dummy copy button but there's still a lot of work to be done for functionality. cc @cuihtlauac

### DIFF
--- a/src/ocamlorg_data/ood.mli
+++ b/src/ocamlorg_data/ood.mli
@@ -128,6 +128,8 @@ module Problem : sig
     tags : string list;
     statement : string;
     solution : string;
+    statement_ml : string;
+    solution_ml : string;
   }
 
   val all : t list
--- a/src/ocamlorg_frontend/pages/problems.eml
+++ b/src/ocamlorg_frontend/pages/problems.eml
@@ -111,10 +111,85 @@ Learn_layout.render
         </div>
       </div>
 
-      <div x-show="statement"><%s! problem.statement %></div>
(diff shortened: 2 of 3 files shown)


tmattio:
From offline conversations:

- The approach to having a global copy button does not work well with statements that contain multiple code blocks. Instead, we could have a copy button on each code block. If this is done upstream, this would also be available on the other sections of the site (learn and package docs)
- One problem with copying the code blocks from the Exercices section is that they are not pastable anywhere: they include the top-level line prompt and the type result. We want users to be able to copy only the code to run in a REPL for instance.
  - mdx probably has logic for this.
  - We can explore how to split the type result from the code block. IIRC mdx does that for the errors already.
 
Using mdx might not be compatible with working upstream, however, so these seem to be two alternative solutions. I'm personally leaning toward looking into what we can do upstream and maybe changing how we use mdx in the exercises markdown.

sabine:
Hmm... what if every exercise had additional sections "Statement ML" and "Solution ML" in the markdown files of the individual problems that contain the exact text that should be copied into an editor?

Then, there's artistic freedom between how to present the problem statement on the website and what will be copied to clipboard (which should be a string that can be copied directly into a code editor or the playground).

Obviously, coping statements for a REPL line by line is also a nice feature, but I would not choose a REPL to work on example problems. The playground or my code editor seems a much more natural fit for solving exercises.

Opinions?

PR #887, opened 2023-02-06
standard library API link in search dropdown

sabine:
Experimental draft: When the search box is focused, a quick link to the Standard Library API pops up.

Multiple people were asking for a quick way to go to the Standard Library. Since we don't have space left in the top nav bar, we extend the search box with a dropdown that opens when we focus within the search box of the top nav bar. 

I suppose, in the longer run, we might want to make the search box search more than just packages:
A dropdown like this could allow us to specify where to search (e.g. "... in packages", "... in learn area", "... in all of ocaml.org"). It's also a place where we could have for more shortcut links (like the one to Standard Library API)? We could filter an extended list of shortcut links and display them in the dropdown. :shrug:

|before|after|
|-|-|
|Screenshot from 2023-02-09 13-21-00|Screenshot from 2023-02-09 13-21-06|

### DIFF
--- a/src/ocamlorg_frontend/components/header.eml
+++ b/src/ocamlorg_frontend/components/header.eml
@@ -21,7 +21,7 @@ let render
 ()
 =
 <header 
-  class="h-20 flex items-center text-white dark:bg-[#171717]"
+  class="h-20 flex items-center dark:bg-[#171717]"
   x-data="{ open: false, sidebar: window.innerWidth > 1024 && true, showOnMobile: false}"
   @resize.window="sidebar = window.innerWidth > 1024">
   <nav class="container-fluid wide header flex justify-between items-center">
@@ -43,8 +43,8 @@ let render
     <ul class="hidden lg:flex items-center space-x-4 xl:space-x-8">
       <li class="relative">
         <form action="/packages/search" method="GET">
-          <div class="flex items-center justify-center">
-            <input type="search" name="q" class="header__search focus:border-gray-800 text-gray-800 border-primary-600 h-10 rounded-l-md appearance-none px-4 py-1 lg:w-56 xl:w-80" placeholder="Search OCaml Packages">
+          <div class="header__search relative flex items-center justify-center">
+            <input type="search" name="q" class="header__search__input focus:border-gray-800 text-gray-800 border-primary-600 h-10 rounded-l-md appearance-none px-4 py-1 lg:w-56 xl:w-80" placeholder="Search OCaml Packages">
             <button aria-label="search" class="h-10 rounded-r-md bg-primary-600 text-white flex items-center justify-center px-4">
               <svg class="w-6 h-6 text-white" fill="currentColor" xmlns="http://www.w3.org/2000/svg"
                 viewBox="0 0 24 24">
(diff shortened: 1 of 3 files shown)


src/ocamlorg_frontend/components/header.eml
```diff
@@ -52,8 +52,14 @@ let render
...
               </svg>
             </button>
-            <div class="header__search__dropdown absolute w-full mt-16 px-4 py-1 mt-20 bg-white text-primary-600 font-semibold border b-1 border-primary-600">
-              <a class="hover:underline" href="<%s Url.api %>">Standard Library API</a>
+            <div class="header__search__dropdown absolute rounded-b-md w-full top-0 mt-10 p-2 bg-white font-semibold border border-primary-600">
+              <span class="pl-2">Or go to:</span>
+              <a class="flex py-2 px-4 gap-4 hover:bg-primary-100 font-regular hover:font-semibold text-primary-600" href="<%s Url.api %>">
+                Standard Library API
+                <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor" class="w-6 h-6">
+                  <path stroke-linecap="round" stroke-linejoin="round" d="M13.5 6H5.25A2.25 2.25 0 003 8.25v10.5A2.25 2.25 0 005.25 21h10.5A2.25 2.25 0 0018 18.75V10.5m-10.5 6L21 3m0 0h-5.25M21 3v5.25" />
```
cuihtlauac: Can't you already put it `Icons.ml`? I prefer resolving a conflict than searching for SVG tags again :-)

cuihtlauac (changes requested):

src/ocamlorg_frontend/components/header.eml
sabine: I will rebase the icons patch, ok?

sabine: Reason being that we might already have this icon in #865

sabine: Or we merge #865 first. The benefits of that cleanup are worth it.

src/ocamlorg_frontend/components/header.eml
```diff
@@ -43,15 +43,24 @@ let render
...
                 <path
                     d="M16.32 14.9l5.39 5.4a1 1 0 0 1-1.42 1.4l-5.38-5.38a8 8 0 1 1 1.41-1.41zM10 16a6 6 0 1 0 0-12 6 6 0 0 0 0 12z" />
               </svg>
             </button>
+            <div class="header__search__dropdown z-10 absolute rounded-b-md w-full top-0 mt-10 p-2 bg-white font-semibold border border-primary-600">
+              <span class="pl-2">Or go to:</span>
+              <a class="flex py-2 px-4 gap-4 hover:bg-primary-100 font-regular hover:font-semibold text-primary-600" href="<%s Url.api %>">
+                Standard Library API
+                <svg xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24" stroke-width="1.5" stroke="currentColor" class="w-6 h-6">
+                  <path stroke-linecap="round" stroke-linejoin="round" d="M13.5 6H5.25A2.25 2.25 0 003 8.25v10.5A2.25 2.25 0 005.25 21h10.5A2.25 2.25 0 0018 18.75V10.5m-10.5 6L21 3m0 0h-5.25M21 3v5.25" />
```
cuihtlauac:
```suggestion
                <%s! arrow_top_right_on_square "w-6 h-6" %>
```

cuihtlauac (changes requested):
How about adding the following definition to `Icons.ml` in this PR (together with the inlined suggestion) ?
```ocaml
let arrow_top_right_on_square class_ =
  <svg xmlns="http://www.w3.org/2000/svg" class="<%s class_ %>" fill="none" viewBox="0 0 24 24" stroke="currentColor" aria-hidden="true">
    <path stroke-linecap="round" stroke-linejoin="round" stroke-width="2" d="M10 6H6a2 2 0 00-2 2v10a2 2 0 002 2h10a2 2 0 002-2v-4M14 4h6m0 0v6m0-6L10 14"></path>
  </svg>

```
We'll have a merge conflict when rebasing  https://github.com/ocaml/ocaml.org/pull/865
but I think it will be easy to resolve.

As a side remark the path used in 865 for this icon is not the same as the one in this PR. There's something I don't understand

sabine:
When there's two icons that look the same and both are meeting quality standards (i.e. looks good), I would keep the one with the shorter path text.

Moved the icon into icons.eml and rebased on main.

cuihtlauac (approved):

PR #897, opened 2023-02-07
Reverse result of get_package_versions

cuihtlauac:
Function was called in three different places:

* In Ocamlorg_package.latest_documented_version, its result was reversed
* In Handler.package_doc, its result was ignored
* In Handler.package_version, its result was sorted

Therefore, it is simpler to always returned a reversed result.

### DIFF
--- a/src/ocamlorg_package/lib/ocamlorg_package.mli
+++ b/src/ocamlorg_package/lib/ocamlorg_package.mli
@@ -145,7 +145,7 @@ val get_packages_with_name : state -> Name.t -> t list option
 (** Get the list of packages with the given name. *)
 
 val get_package_versions : state -> Name.t -> Version.t list option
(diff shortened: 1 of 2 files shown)


src/ocamlorg_package/lib/ocamlorg_package.mli
```diff
@@ -145,7 +145,7 @@ val get_packages_with_name : state -> Name.t -> t list option
 (** Get the list of packages with the given name. *)
 
 val get_package_versions : state -> Name.t -> Version.t list option
-(** Get the list of versions for a package name. *)
+(** Get the reversed list of versions for a package name. *)
```
sabine: I think it would be helpful to state what the ordering is. I.e. from latest to newest, or newest to latest.

sabine (approved):

PR #904, opened 2023-02-08
Use >>= instead of let* for Lwt.bind

cuihtlauac:


src/ocamlorg_package/lib/ocamlorg_package.ml
```diff
@@ -367,15 +367,12 @@ let doc_exists t name version =
...
     | [] -> Lwt.return None
-    | _ -> (
-        let version = List.hd vlist in
-        let* doc = doc_exists t name version in
-        match doc with
-        | Some version -> Lwt.return (Some version)
-        | None -> aux (List.tl vlist))
+    | version :: vlist ->
+        doc_exists t name version
+        >>= Option.fold ~none:(aux vlist) ~some:Lwt.return_some
```
sabine, line 375: I know it's a thing people commonly do... but maybe we can try not to reuse/overshadow parameter names in match-expressions?

```suggestion
    | version :: vs ->
        doc_exists t name version
        >>= Option.fold ~none:(aux vs) ~some:Lwt.return_some
```

sabine: Also not sure if we want to keep the code more verbose and obvious, to make it easier for beginners to contribute. :thinking:

tmattio:
I'm not sure I understand this PR, or why it was merged: @sabine's reviews don't seem to have been addressed (I also think the use of infix operators instead of let bindings is hurting readability, and it's not the convention used in the code base), the name of the PR isn't representative of the changes (most changes are added documentation from what I can see, and not related to the change of the operator), the changes in the documentation haven't been reviewed by @christinerose and seem incomplete, and there's no description to describe the changes.

Is it an accidental merge?

cuihtlauac:
Yes I made a mistake, sorry about that. Neither the >>= (personal experimentation) nor the docs (work in progress) should have been merged. I will revert both changes.

PR #905, opened 2023-02-08
add the last publication date to package search results

sabine:
When looking for packages, it can be helpful to know when the last version was published.

This patch adds the last publication date of a package to the package search results.

|before|after|
|-|-|
|Screenshot 2023-02-08 at 16-04-02 OCaml Packages · Search Result|Screenshot 2023-02-08 at 16-04-06 OCaml Packages · Search Result|


cuihtlauac (approved):
Great, that's useful. Maybe the wording should be the same there and in the package metadata summary. BTW, "Last published" feels better than "Published".

PR #908, opened 2023-02-09
Slim playground

jonludlam:
The main aim of this PR is to cut down the playground size a lot.

Previously:
```sh
$ du -h -c *.js
4.6M    merlin-v1.js
3.4M    playground-v1.js
 21M    worker-v1.js
 29M    total
```

Now:
```sh
du -h -c *.js
2.7M    merlin-v1.js
3.4M    playground-v1.js
3.6M    worker-v1.js
9.7M    total
```

Not quite the whole story though! the toplevel and merlin will also have to dynamically load cmis from the server, and these are found in misnamed `playground/asset/stdlib` directory, which contains cmis for both stdlib and domainslib.

This cuts down on the functionality somewhat in the sense that the cmis for `compiler-libs`, `js_of_ocaml` are no longer usable, but I don't think anyone using this toplevel will miss them.

I suspect there's still some slimming that can be done as `playground.js` is still very big for what it does, but that's a task for another PR.

This PR also fixes merlin, which hasn't been working properly since the 5.0 update.


playground/ocamlorg-playground.opam
```diff
@@ -48,6 +50,6 @@ dev-repo: "git+https://github.com/ocaml/ocaml.org.git"
 pin-depends: [
   ["esbuild.dev" "git+https://github.com/tmattio/opam-esbuild#7738af5912cc9f554d31da81d2334e88052102fb"]
   ["code-mirror.dev" "git+https://github.com/patricoferris/jsoo-code-mirror#8fe48910e265ff87f9fc94ceb7b3d19fac102a96"]
-  ["merlin-js.dev" "git+https://github.com/tmattio/merlin-js#74edff465942cc4f0b3b51b3f8264db53a460aa2"]
+  ["merlin-js.dev" "git+https://github.com/jonludlam/merlin-js#276d0af663db6aa0bc1bf8733eca206cf91dbcb7"]
```
tmattio: Testing and reviewing is clearly not blocked by this, but should we aim to upstream merlin-js patches (from both my fork and yours) before we merge?

cc @voodoos

jonludlam: We should definitely have a longer term plan for this, rather than playing 'pass the repo' :-D

jonludlam:
Domainslib wasn't usable as it wasn't referenced and thus unlinked. I've added `-linkall` to the dune build of the worker and now it's working as expected, at the expense of worker.js going from 3.6M to 3.9M.

playground/ocamlorg-playground.opam
voodoos: Yes, it would be best to upstream the patches, and I can even cut a release :-)

voodoos:
> The main aim of this PR is to cut down the playground size a lot.
> 
> Previously:
> 
> ```shell
> $ du -h -c *.js
> 4.6M    merlin-v1.js
> 3.4M    playground-v1.js
>  21M    worker-v1.js
>  29M    total
> ```
> 
> Now:
> 
> ```shell
> du -h -c *.js
> 2.7M    merlin-v1.js
> 3.4M    playground-v1.js
> 3.6M    worker-v1.js
> 9.7M    total
> ```
> This PR also fixes merlin, which hasn't been working properly since the 5.0 update.

Just a remark: 29M was (fortunately!) not the amount of data actually sent to the clients, the gzipped size of these artifacts is 6.32MB according to my browser:

```
1.22MB merlin-v1.js 
1.08MB playground-v1.js
4.02MB worker-v1.js
```

It is still is a nice improvement and loading the cmis of the stdlib only when they are needed makes a lot of sense. Also this should open the door to providing the cmis of other opam packages... did you also experiment with this ? In any case I will be happy to upstream these patches to `merlin-js` 🙂 !

jonludlam:
> this should open the door to providing the cmis of other opam packages..

Indeed, this work comes from some previous experiments where we were adapting the docs pipeline to build jsoo toplevels for every package in opam. I'd like to finish that off!

sabine:
After a short time of sending them uncompressed to the clients, we fixed that. Now, we're going one step further and will serve the playground files under immutable digest URLs: https://github.com/ocaml/ocaml.org/pull/930.

What's the status on this PR? Should #930 wait until after this PR is merged?

tmattio:
I think the last missing piece is to merge https://github.com/voodoos/merlin-js/pull/4 and update the PR to use a released version of merlin-js. I wouldn't wait on this PR to merge #930 though

jonludlam:
the PR to merlin-js has been merged, so I've rebased this. I've removed the domainslib bit for now - we need to figure out the threads problem there.

jonludlam:
Also, at some point between `js_of_ocaml` 5.0.0 and 5.2.0 the size of the playground js went down a lot, so the new sizes are:

```
2.7M    merlin.min.js
1.0M    playground.min.js
3.4M    worker.min.js
7.1M    total
```

and due to compression, the transferred sizes are now:

```
playground.js: 340k
merlin.js: 844k
worker.js: 890k
```

The improvement in playground.js gains has been in `main` since @voodoos fixed merlin (and upgraded to js_of_ocaml 5.1.1+2f5be3f0-dirty)

tmattio:
I notice 404 in the browser console because the playground tries to load internal modules (`std_exit.cmi`, `camlinternalAtomic.cmi`, etc.). Is this expected?

Otherwise this works great!

jonludlam:
Ah good spot! I've synced the list of toplevel modules now, so that should be fixed.

tmattio (approved):
LGTM! Thanks a ton for the awesome work @jonludlam, this is a great foundation for very exciting work, like loading OCaml packages in the playground, or OCaml toplevels in the package documentation. Thank you! 🎉

sabine (approved):
Thank you, fully agree with @tmattio, this is nice!

gpetiot:
I needed to install `domainslib` to build the playground since this commit. Maybe a dependency is missing in playground/ocamlorg-playground.opam?

jonludlam:
Oh, I thought I'd nuked mention of domainslib. I must have left something behind that mentions it...

jonludlam:
https://github.com/ocaml/ocaml.org/pull/1248

PR #909, opened 2023-02-10
update ahrefs job

Khady:
The https://ahrefs.com/jobs/ocaml-developer position is evergreen, publication date is not relevant in that case.


cuihtlauac:
Thanks for letting us know about that @Khady. We need to find a way to mark positions as always available.

cuihtlauac (approved):

Issue #910, opened 2023-02-10
Job freshness


cuihtlauac:
Jobs should have a freshness label such as (not all options make sense, this is just illustrative):
- “Posted today”
- “Posted X days ago”
- “Posted this week”
- “Posted X weeks ago”
- “Posted this month”
- “Posted X months ago“

However, there are two special/special cases:
- “Evergreen” (such as discussed in https://github.com/ocaml/ocaml.org/pull/909)
- “Freshness status unknown”

sabine:
I think this is closely related to categorizing jobs and how we deal with new incoming listings and what sorting and filtering options there will be.

IMO the existing publication date is fine as a freshness label, but it would be helpful to establish categories so it is clear who the job is addressed to.

E.g.
- category of job (employee vs. contractor/freelancer)
- expectation on the worker's time commitment (part-time, full-time)
- duration of the job (short-term vs. long-term)
- type of listing (spontaneous application vs a specific job posting)

:thinking: Something like that.

PR #911, opened 2023-02-10
add OCamlPro spontaneous application job listing

sabine:
Adds OCamlPro listing from https://discuss.ocaml.org/t/list-your-open-ocaml-positions-on-the-ocaml-org-job-board/11377/5

Also removes publication date on Tarides' spontaneous application listing.

I think, in general, we should eventually not show publication dates on spontaneous application listings. However, it is nice to see the new spontaneous application listing at the top of the list for a while. That makes people aware that this is a listing that recently came in.

After some point it doesn't matter anymore when a spontaneous application listing was published - all that matters is that it is still relevant. So we should consider checking maybe every 6 months if the spontaneous application listings are still active.

### DIFF
--- a/data/jobs.yml
+++ b/data/jobs.yml
@@ -1,4 +1,10 @@
 jobs:
+  - title: Spontaneous Application
+    link: https://ocamlpro.com/jobs/
+    location: France
+    publication_date: 2023-02-10
+    company: OCamlPro
+    company_logo: https://ocamlpro.com/assets/img/logo_ocamlpro.png
   - title: Cryptography Engineer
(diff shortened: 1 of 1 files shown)


cuihtlauac:
I have opened an issue on 'Job Freshness': https://github.com/ocaml/ocaml.org/issues/910

I believe the date is not key. What matters is providing clues to help the user answer the question: "is this offer still relevant". The older the smaller the changes, unless it is "spontaneous/evergreen" offer

cuihtlauac (approved):

PR #915, opened 2023-02-10
Add a debugging back trace example

jmid:
To improve the debugging guide, this PR adds an example of how to print a back trace for an uncaught exception.
This is briefly mentioned in https://ocaml.org/docs/error-handling#stacktraces too, but I figured a spelled-out example could be useful to a newcomer.

Nit: The error handling page above calls these `stacktraces` (one word) where as the manual page https://v2.ocaml.org/manual/runtime.html calls them both `backtraces` (8 occurrences) and `back traces` (2 occurrences). 

I went with the `back trace` for now
- to side with the manual
- as it seems more correct English, and 
- because it also agrees with `OCAMLRUNPARAM`s `b` option

### DIFF
--- a/data/tutorials/wf_01_debugging.md
+++ b/data/tutorials/wf_01_debugging.md
@@ -9,12 +9,16 @@ date: 2021-05-27T21:07:30-00:00
 
 # Debugging
 
-This tutorial presents two techniques for debugging OCaml programs:
+This tutorial presents three techniques for debugging OCaml programs:
 
 * [Tracing functions calls](#tracing-functions-calls-in-the-toplevel),
   which works in the interactive toplevel.
 * The [OCaml debugger](#the-ocaml-debugger), which allows analysing programs
   compiled with `ocamlc`.
+* [How to get a back trace for an uncaught
+  exception](#printing-a-back-trace-for-an-uncaught-exception) in an
+  OCaml program
+
 
 ## Tracing Functions Calls in the Toplevel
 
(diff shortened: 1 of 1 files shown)


cuihtlauac:
This is great @jmid, thanks a lot.

cuihtlauac (approved):

data/tutorials/wf_01_debugging.md
```diff
@@ -9,12 +9,16 @@ date: 2021-05-27T21:07:30-00:00
 
 # Debugging
 
-This tutorial presents two techniques for debugging OCaml programs:
+This tutorial presents three techniques for debugging OCaml programs:
 
 * [Tracing functions calls](#tracing-functions-calls-in-the-toplevel),
   which works in the interactive toplevel.
```
christinerose, line 15:
```suggestion
  which work in the interactive toplevel.
```

data/tutorials/wf_01_debugging.md
```diff
@@ -339,3 +343,52 @@ Under Emacs you call the debugger using `ESC-x` `ocamldebug a.out`. Then Emacs
 will send you directly to the file and character reported by the debugger, and
 you can step back and forth using `ESC-b` and `ESC-s`, you can set up break
```
christinerose, line 344:
```suggestion
you can step back and forth using `ESC-b` and `ESC-s`. Plus, you can set up break
```

data/tutorials/wf_01_debugging.md
```diff
@@ -339,3 +343,52 @@ Under Emacs you call the debugger using `ESC-x` `ocamldebug a.out`. Then Emacs
...
 points using `CTRL-X space`, and so on...
+
+## Printing a back trace for an uncaught exception
```
christinerose:
```suggestion
## Printing A Back Trace for an Uncaught Exception
```

data/tutorials/wf_01_debugging.md
```diff
@@ -339,3 +343,52 @@ Under Emacs you call the debugger using `ESC-x` `ocamldebug a.out`. Then Emacs
...
+
+Getting a back trace for an uncaught exception can be informative to
+understand in which context a problem occurs. However, by default
```
christinerose:
```suggestion
understand in which context a problem occurs. However, by default,
```

christinerose (approved):
Minor formatting suggetions

data/tutorials/wf_01_debugging.md
jmid: Ah, yes. Good catch!

data/tutorials/wf_01_debugging.md
jmid: OK :+1:

jmid:
I've made the two corrections relevant to the current PR.

I think the other two suggested text changes are better fit for another PR. In particular, for the first suggestion:
```diff
-  which works in the interactive toplevel.
+  which work in the interactive toplevel.
```
it is the *tracing* (singular) that works in the toplevel, not the 'function calls' (plural).
Hence a different formulation to more clearly convey the intended meaning would be to rephrase it to
```
* [Function call tracing](#tracing-functions-calls-in-the-toplevel),
  which works in the interactive toplevel.
```
Anyway - this is heading into an argument of rephrasing a sentence which is neither authored nor changed in this PR, so I recommend addressing it elsewhere.

PR #917, opened 2023-02-11
add singapore location for ahrefs job

Khady:
Thanks for the quick improvements :)


cuihtlauac (approved):

sabine:
Always happy to. Small but useful tasks we try to do immediately when they're brought up! :grin:

PR #929, opened 2023-02-17
serve css/js assets under digest urls, cache-control

### LINKED ISSUE #839: Enable HTTP headers to allow browser caching of assets (Meta Issue)
Problems with caching of assets (CSS, JavaScript, Fonts) are mentioned in these issues:
- https://github.com/ocaml/ocaml.org/issues/704
- https://github.com/ocaml/ocaml.org/issues/838
- https://github.com/ocaml/ocaml.org/issues/797

Underlying problem: We cannot currently enable asset caching via HTTP headers because assets are being served under their exact filenames.

What needs to be done: serve assets under filenames that include a hash (file content or last modified time). This enables us to allow caching via HTTP headers indiscriminately while still maintaining the ability to make changes that will reach our users.

Idea: use a PPX that, at compile time hashes the file content (or inspects the last-modified time) and copies the file accordingly, then returns the path of the hash-named file. Then, we can use, e.g. `[%asset "css/main.css" ]` in our templates.

sabine:
Changes the caching strategy for assets.

All assets `{PATH}` (e.g. `css/main.css`) are now served under two URLs:
1. `/{PATH}` (Cache-control header allows caching for 1 day)
2. `/_/{DIGEST}/{PATH}` (Cache-control header allows caching for a very long time and declares the file immutable)

Digests are looked up via `assets.ml` (produced by `ocaml-crunch` with mode `plain`). They are a hash of the file content.

In the templates, there is now a parameter `digest_url : string -> string` which renders the path containing the digest (only for assets from `assets.ml`, as we do not expect files from `media.ml` to change).

Note: `Cache-control: immutable` tells the browser that checking whether the file was modified is unnecessary.

Resolves https://github.com/ocaml/ocaml.org/issues/839 except for the playground.

Note that, since `inter.css` has a relative import for the font files, all the font files are served under `inter.css`'s digest. So, if anyone ever needs to update the font files, they have to modify `inter.css` to change its digest.


cuihtlauac (changes requested):
See: https://github.com/ocaml/ocaml.org/tree/cuihtlauac-929

src/ocamlorg_static/file.ml
```diff
@@ -0,0 +1,18 @@
...
+let of_url_path path =
+  let xs = String.split_on_char '/' path in
+  match xs with
+  | [] -> raise (Invalid_argument "invalid Digest_url: path cannot be empty")
+  | "_" :: [] | [ "_"; _ ] ->
+      raise
+        (Invalid_argument
+           "invalid Digest_url: paths starting with '_' must be followed by a \
+            digest and filepath")
+  | "_" :: x :: xs -> { digest = Some x; filepath = String.concat "/" xs }
```
cuihtlauac: Are you sure we can't simplify this one? Usually, sorting match branches in decreasing pattern length order allow maximum sharing. Something like that:
 
```suggestion
  | "_" :: x :: xs -> { digest = Some x; filepath = String.concat "/" xs }
  | "_" :: _ ->
      raise
        (Invalid_argument
           "invalid Digest_url: paths starting with '_' must be followed by a \
            digest and filepath")
```

cuihtlauac (changes requested):
I'm a bit tired, I've left a question w/o testing, sorry

src/ocamlorg_static/file.ml
sabine: Thank you, good suggestion!

src/ocamlorg_web/lib/static.ml
```diff
@@ -38,31 +38,37 @@ let not_modified ~last_modified request =
...
-        match static_url.digest with
-        | None -> Fmt.str "max-age=86400" (* one day *)
-        | Some _ -> "max-age=31536000, immutable"
-      in
-      Dream.respond
-        ~headers:([ ("Cache-Control", cache_control) ] @ Dream.mime_lookup path)
-        asset
+          let cache_control =
+            match static_file.digest with
+            | None -> Fmt.str "max-age=86400" (* one day *)
```
cuihtlauac:
```suggestion
            | None -> "max-age=86400" (* one day *)
```

src/ocamlorg_static/file.ml
```diff
@@ -8,11 +8,7 @@ type t = { filepath : string; digest : string option }
...
-  | "_" :: [] | [ "_"; _ ] ->
-      raise
-        (Invalid_argument
-           "invalid Digest_url: paths starting with '_' must be followed by a \
-            digest and filepath")
-  | "_" :: x :: xs -> { digest = Some x; filepath = String.concat "/" xs }
-  | _ -> { digest = None; filepath = path }
+  | "_" :: x :: xs -> Some { digest = Some x; filepath = String.concat "/" xs }
+  | "_" :: _ -> None
+  | [] -> None
```
cuihtlauac:
```suggestion
  | "_" :: _ | [] -> None
```
If you switch to option, the double clause is back :-)

cuihtlauac (changes requested):
Pedantic remarks, sorry about that

cuihtlauac (approved):

PR #930, opened 2023-02-17
Improved Caching of Playground Assets (serve under digest-URLs)

sabine:
This patch aims to serve the playground assets behind a digest-URL, so that browsers can cache the playground assets indefinitely.

Idea:
1. similar to #929, we render a digest-URL for the playground in the `<script>`-tag in `playground.eml`.
2. since the main file of the playground imports two other files (`merlin.min.js`, `worker.min.js`), we need to make sure these files are also loaded from digest-URLs. This is done by adding a `data-merlin-url` and `data-worker-url` attributes to the `script`-tag that loads `playground.min.js`. From there, the URLs are picked up by the playground in `main.ml` when the playground runs.

We never have to be careful to version the playground assets again.


sabine:
@cuihtlauac the patch currently computes the playground file digest every time a playground URL is rendered.

I don't know what the elegant / obvious / short way to memoize the `digest` function in `Playground` is.

cuihtlauac:
Have pushed a branch on top of https://github.com/ocaml/ocaml.org/pull/929 with the stuff which is specific to this PR?

sabine:
I'm just being dense here, I think I can load all the digests at start into a hashmap. :see_no_evil:

cuihtlauac:
Winter has come. I feel exactly the same. Probably the mood of the times.

sabine:
Ok so it seems computing the digests in module initialisation means that digests of the playground are computed when running the tests......... which is neither what we want nor successful.

I think if the playground had a build step that crunched its assets, that would probably be most reasonable. Playground is only rarely being rebuilt, and then we check in the resulting crunched assets into the repo?

src/ocamlorg_static/ocamlorg_static.ml
```diff
@@ -1,18 +1,42 @@
...
 
-let of_url_path = File.of_url_path
+module Asset = struct
+  include Asset
+
+  (* given the path of a file from `assets.ml`: 1. looks up the file's digest in
+     and 2. returns the corresponding digest URL for use in templates *)
+  let url filepath =
+    let digest =
+      Option.map (fun d -> Dream.to_base64url d) (Asset.hash filepath)
```
cuihtlauac:
```suggestion
      Option.map Dream.to_base64url (Asset.hash filepath)
```

src/ocamlorg_static/ocamlorg_static.ml
```diff
@@ -1,18 +1,42 @@
...
-         (Fmt.str
-            "ERROR: '%s' is rendered via asset_url, but it is not an asset!"
-            filepath));
-  File.to_url_path ?digest filepath
+  (* given the path of a file from `assets.ml`: 1. looks up the file's digest in
+     and 2. returns the corresponding digest URL for use in templates *)
+  let url filepath =
+    let digest =
+      Option.map
+        (fun d -> Dream.to_base64url d)
```
cuihtlauac:
```suggestion
        Dream.to_base64url
```

cuihtlauac (changes requested):
Peanuts

src/ocamlorg_static/dune
```diff
@@ -25,4 +25,18 @@
...
+  (source_tree %{workspace_root}/playground/asset))
+ (action
+  (with-stdout-to
+   %{null}
+   (run
+    %{bin:static-file-digest}
+    %{workspace_root}/playground/asset/
+    -s
+    -o
+    %{target}))))
```
cuihtlauac: I know this is ugly:

```suggestion
     %{target}
    (chdir %{workspace_root}/playground/asset
    (pipe-stdout
      (run %{bin:md5sum} %{deps})
      (run %{bin:awk} "{ print \"| \\\"\"$2\"\\\" -> Some (from_hex \\\"\"$1\"\\\")\"}")
      (system "(echo \"let digest = Digest.(function\"; cat -; echo \"| _ -> None)\")"))))))
```

But it turns an ad-hoc dependency into a system dependency.

What do you think?

cuihtlauac (changes requested):

src/ocamlorg_static/dune
sabine: I can see the appeal of having no dependency and this is really concise too, but if we do that we should do it as a separate script (with comments what it does).

cuihtlauac: You are right; it should be a script. Then if we have a script, it doesn't have to be a shell-script, it could be a OCaml-script.

PR #931, opened 2023-02-19
Fix editor swiper's width

### LINKED ISSUE #836: Swiper takes up too much space on Home page, and causes horizontal scroll on small screen in Chrome 109.0.5414.74 (Official Build) (64-bit)
Swiper takes up more space than expected on lg screen:
Screenshot from 2023-01-18 17-44-02

Swiper causes horizontal overflow on sm screen:
Screenshot from 2023-01-18 17-44-16

Found while working on making the HTML more semantic.

Solution: need to adjust styles to prevent the swiper from taking up more space than intended. Might want to switch to a different swiper (or remove swiper altogether!) if it doesn't let itself be contained easily.

For reference: This is what it looks like when I hide overflow and force the swiper to a max-width of 100% of its surrounding container (i.e. the intended width according to the layout):
Screenshot from 2023-01-18 17-49-36

YassineHaouzane:
Fixes #836 by removing fixed width on images and disabling swiper's resizing.

### DIFF
--- a/src/ocamlorg_frontend/css/styles.css
+++ b/src/ocamlorg_frontend/css/styles.css
@@ -78,10 +78,6 @@ body {
   width: 2px;
 }
 
-.mySwiper2 {
-  width: 793px;
-}
-
 .codebox {
   height: 462px;
 }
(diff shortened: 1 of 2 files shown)


cuihtlauac (approved):
Thanks @YassineHaouzane, this is a nice and helpful contribution

sabine (approved):
LGTM

PR #932, opened 2023-02-20
remove fallback to old docs-data

sabine:
Removes the fallback to old (pre-odoc-2.2.0) documentation.

Todo:
- [x] fix problem with placement of CHANGELOG, README, LICENSE (see https://github.com/ocaml/ocaml.org/pull/985)
- [x] promote package documentation to `live` folder

### DIFF
--- a/src/ocamlorg_frontend/components/navmap.eml
+++ b/src/ocamlorg_frontend/components/navmap.eml
@@ -5,7 +5,6 @@ type kind =
   | Leaf_page
   | Module_type
   | Parameter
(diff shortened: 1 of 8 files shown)


cuihtlauac (commented):
Looks good to me.

sabine:
Status of this:

Current voodoo has a bug where it places README, CHANGELOG and LICENSE files in the root directory of the package (instead of in `docs/`). So currently, we fall back the the old README, CHANGELOG and LICENSE data 100% of the time.

PR #939, opened 2023-02-21
Adds package autocomplete search to the search dropdown

sabine:
Patch adds interactive search functionality for packages on the top navbar search box dropdown and to the packages page.

|before|after|
|-|-|
|Screenshot from 2023-02-22 10-46-12|Screenshot from 2023-02-22 10-46-19|
|-|Screenshot from 2023-02-22 10-46-41|
|-|Screenshot from 2023-02-21 17-06-09 indicator displays when search is happening|
|Screenshot from 2023-02-22 10-45-45|Screenshot from 2023-02-22 10-45-57 packages page has basic interactive search|


src/ocamlorg_frontend/url.ml
```diff
@@ -1,6 +1,7 @@
 let index = "/"
 let packages = "/packages"
 let packages_search = "/packages/search"
+let packages_autocomplete_f = "/packages/autocomplete"
```
cuihtlauac: Is `_f` required? I don't get what it means, what it's used for.

src/ocamlorg_web/lib/handler.ml
```diff
@@ -346,14 +346,27 @@ let packages_search t req =
...
-        Ocamlorg_package.search_package ~by_popularity:true t search
+        Ocamlorg_package.search_package ~sort_by_popularity:true t search
       in
       let total = List.length packages in
       let results = List.map (package_meta t) packages in
       let search = Dream.from_percent_encoded search in
       Dream.html (Ocamlorg_frontend.packages_search ~total ~search results)
   | None -> Dream.redirect req Ocamlorg_frontend.Url.packages
 
+let packages_autocomplete_f t req =
```
cuihtlauac: Same remark as in `Url` regarding `_f`

src/ocamlorg_web/lib/handler.ml
```diff
@@ -346,14 +346,27 @@ let packages_search t req =
...
+  | Some search ->
+      if search <> "" then
+        let packages =
+          Ocamlorg_package.search_package ~sort_by_popularity:true t search
+        in
+        let results = List.map (package_meta t) packages |> List.take 5 in
+        let search = Dream.from_percent_encoded search in
+        Dream.html (Ocamlorg_frontend.packages_autocomplete_f ~search results)
+      else Dream.html ""
+  | None -> Dream.html "No query..."
```
cuihtlauac:
```suggestion
  | Some search when search <> "" ->
      let packages =
        Ocamlorg_package.search_package ~sort_by_popularity:true t search
      in
      let results = List.map (package_meta t) packages |> List.take 5 in
      let search = Dream.from_percent_encoded search in
      Dream.html (Ocamlorg_frontend.packages_autocomplete_f ~search results)
  | _ -> Dream.html ""
```

src/ocamlorg_frontend/pages/packages_autocomplete_f.eml
```diff
@@ -0,0 +1,39 @@
...
+~search
+(text: string)
+=
+  let render_item = function 
+    | Str.Delim s -> {|<span class="bg-background-light-blue text-gray-800 font-normal">|} ^ Dream.html_escape s ^ {|</span>|}
+    | Text s -> Dream.html_escape s
+  in
+  let r = Str.global_replace (Str.regexp "[ \t]+") "\\|" search in
+  let split = Str.full_split (Str.regexp_case_fold r) text in
+  List.fold_left (fun a b -> a ^ render_item b) "" split
```
cuihtlauac: Where can we share this and its copy in `package_search.eml`? Maybe in some file in `src/ocamlorg_frontend/component`?

cuihtlauac (changes requested):
Incantations found in `header.eml` should also be used in `packages.eml`?

src/ocamlorg_frontend/url.ml
sabine: intention is to signify that this is not a full HTML page but only a fragment. Could possibly mark that better :thinking:

src/ocamlorg_frontend/pages/packages_autocomplete_f.eml
sabine: In this case, I think everything except the styles themselves could be shared. That's a good idea.

cuihtlauac (approved):
I like `Search.highlight_search_terms`. It makes me think of what we've done with icons. Could it be the beginning of a pattern? We would gradually extract small pieces of common or look-alike stuff and share them that way.

sabine:
Yes, exactly, factoring such components out when we notice them is good.

The criterion for factoring out is:
"Is this used in a similar context?"

There is a fine line to walk between duplication and wrong abstraction in the frontend world. It always has to be obvious which other locations may be affected by a change. Else it can end up in a "whack-a-mole situation" - fixing one thing breaks another.

It's helpful to give the caller a sensible amount of control over the styles.

cuihtlauac:
> There is a fine line to walk between duplication and wrong abstraction in the frontend world. It always has to be obvious which other locations may be affected by a change. Else it can end up in a "whack-a-mole situation" - fixing one thing breaks another.

100% agree. That's why I believe working in a bottom-up way makes more sense. Don't factor in until it is obviously needed or beneficial. When factoring, prefer small and stupid things rather than trying to be smart. If lucky, in the long run, we may end up factoring big things, but it will come in time, if the need ever appears.

cuihtlauac (approved):

cuihtlauac (approved):

PR #943, opened 2023-02-22
Create an ocamlorg.global library

cuihtlauac:
Directory: src/global
Library Module Name: Ocamlorg
Library Dependency Name: ocamlorg.global
Submodules:
  - Import
  - Url

The goal of this library is to group in a single place stuff used
across several libraries and have a single place for extending Stdlib.

Many files touched, but it's only:
* moving definitions
* making definition visible
* shortening paths

In a second stage, this should be the place where /data types definitions
would be shared. Today, if my count is correct, each is copied thrice:
* src/ocamlorg_data/XXX.mli
* tool/ood-gen/lib/XXX.ml (twice, one as a string literal)


sabine (approved):
I strongly agree with this. Good cleanups.

Indeed, `Url` should not be coming from `Ocamlorg_frontend` when it is a shared resource between `Ocamlorg_frontend` and `Ocamlorg_web`.

PR #949, opened 2023-02-23
package.version is either a `Specific` version or `Latest` (mostly a refactor)

sabine:
This patch
* renames and refactors `Ocamlorg_frontend.Package_intf` to `Ocamlorg_frontend.Package`, which has a notion of whether we are viewing this package through a 'latest`-URL or a specific-version-URL and
* adds `latest_version` to `Ocamlorg_frontend.Package`, so that we can display the specific latest version in the version switcher dropdown
* renders "latest" entry in the version dropdown next to the specific latest version (I don't know a package that has an `avoid-version`-flagged newest version off the top of my head, but on such a package you would really see the difference in a pleasant way).

|before|after|
|-|-|
|Screenshot from 2023-02-23 18-03-20 latest version is displayed as "latest" only|Screenshot from 2023-02-23 18-03-10 latest version entry includes specific version and appears in front of specific latest version|

### DIFF
--- a/src/ocamlorg_frontend/package_intf.ml
+++ b/src/ocamlorg_frontend/package_intf.ml
@@ -1,15 +1,28 @@
+type version = Latest | Version of string
+
 type package = {
   name : string;
   description : string;
   license : string;
-  version : string;
+  version : string option;
+      (* None = we are looking at the latest version of the package on a
+         /latest-URL, Some v = we are looking at version v *)
   versions : string list;
+  latest_version : string;
   tags : string list;
   rev_deps : string list;
   authors : Ood.Opam_user.t list;
   maintainers : Ood.Opam_user.t list;
   publication : float;
 }
+
+let exact_version package =
+  match package.version with None -> package.latest_version | Some v -> v
+
+let render_package_version package =
+  match package.version with
+  | None -> Fmt.str "%s (latest)" package.latest_version
+  | Some v -> v
 
 type packages_stats = {
   nb_packages : int;
--- a/src/ocamlorg_frontend/components/navmap.eml
+++ b/src/ocamlorg_frontend/components/navmap.eml
@@ -104,15 +104,13 @@ let rec nested_render ~path (item : toc) =
 let render
 ~(package: Package_intf.package)
 ~path
-~is_latest_url
 (maptoc : t)
 =
-  let version = if is_latest_url then None else Some package.version in
   <div class="navmap pb-12">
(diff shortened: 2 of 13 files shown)


src/ocamlorg_frontend/package_intf.ml
```diff
@@ -1,16 +1,29 @@
+type version = Latest | Version of string
```
cuihtlauac: Are you using this? I can't find where. If you do, consider renaming `Version` into something like `Numbered`, `Released` or `Named` to avoid using the same word twice.

src/ocamlorg_frontend/package_intf.ml
```diff
@@ -1,16 +1,29 @@
...
+
 type package = {
   name : string;
   description : string;
   license : string;
-  version : string;
+  version : string option;
+      (* None = we are looking at the latest version of the package on a
+         /latest-URL, Some v = we are looking at version v *)
```
cuihtlauac: Maybe using your `version` type would be more explicit/safer than a comment.

src/ocamlorg_frontend/package_intf.ml
```diff
@@ -1,16 +1,29 @@
...
   publication : float;
 }
 
+let exact_version package =
+  match package.version with None -> package.latest_version | Some v -> v
+
+let render_package_version package =
+  match package.version with
+  | None -> Fmt.str "%s (latest)" package.latest_version
+  | Some v -> v
```
cuihtlauac: I don't think function definitions should go into an `_intf.ml` file. To my understanding, they're meant to host type definitions only.

cuihtlauac (changes requested):

src/ocamlorg_frontend/package_intf.ml
sabine: Oversight. Removed because matching on this was too verbose in the templates.

src/ocamlorg_frontend/package_intf.ml
sabine: Ok that makes sense. This module is essentially "what a package looks like to the templates", so renaming to `package.ml` would make sense.

src/ocamlorg_frontend/package_intf.ml
sabine: I'll give it another try to see if I can reduce the verbosity I saw when I did this. With the helper functions it may not be too bad, and I agree that the explicitness is helpful.

PR #952, opened 2023-02-24
Get rid of Url.package_with_version

cuihtlauac:
* Remove old Url.package
* Rename old Url.package_with_version into Url.package
* Update Url.with_version
  - None -> "/latest" (as in old Url.package_with_version)
  - Some "" -> "" (as in old Url.package)
  - Some x -> "/x" (as in old Url.package_with_version)
* Use new Url.with_version in new Url.package
* At call sites of old Url.package, add ~version:"" to get the
  same behaviour
* At call sites of old Url.package_with_version, replace by Url.package

A bit convoluted, but did not managed to observe a change of behaviour.

### DIFF
--- a/src/global/url.ml
+++ b/src/global/url.ml
@@ -3,15 +3,15 @@ let packages = "/packages"
 let packages_search = "/packages/search"
 let packages_autocomplete_fragment = "/packages/autocomplete"
 let with_hash = Option.fold ~none:"/p" ~some:(( ^ ) "/u/")
-let package ?hash v = with_hash hash ^ "/" ^ v
 let package_docs v = "/p/" ^ v ^ "/doc"
-let with_version = Option.value ~default:"latest"
 
-let package_with_version ?version ?hash v =
-  with_hash hash ^ "/" ^ v ^ "/" ^ with_version version
+let with_version =
+  Option.fold ~none:"/latest" ~some:(fun v -> if v = "" then v else "/" ^ v)
+
+let package ?version ?hash v = with_hash hash ^ "/" ^ v ^ with_version version
 
 let package_doc ?hash ?version ?(page = "index.html") v =
-  with_hash hash ^ "/" ^ v ^ "/" ^ with_version version ^ "/doc/" ^ page
+  with_hash hash ^ "/" ^ v ^ with_version version ^ "/doc/" ^ page
 
 let community = "/community"
 let success_story v = "/success-stories/" ^ v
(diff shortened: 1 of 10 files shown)


src/ocamlorg_frontend/url.ml
```diff
@@ -3,15 +3,14 @@ let packages = "/packages"
...
 let with_hash = Option.fold ~none:"/p" ~some:(( ^ ) "/u/")
-let package ?hash v = with_hash hash ^ "/" ^ v
 let package_docs v = "/p/" ^ v ^ "/doc"
-let with_version = Option.value ~default:"latest"
+let with_version = Option.fold ~none:"/latest" ~some:(fun v -> if v = "" then v else "/" ^ v)
 
-let package_with_version ?version ?hash v =
-  with_hash hash ^ "/" ^ v ^ "/" ^ with_version version
+let package ?version ?hash v =
+  with_hash hash ^ "/" ^ v ^ with_version version
```
sabine: IMO all the URLs that only redirect somewhere else should be separate from those intended for use in templates.

So, for URLs used in the templates and as redirection target, we should always require either a version or "latest".

cuihtlauac: Yes, I had something along those lines in mind as a second stage/commit. It makes sense to use an optional argument
- Avoids writing Some
- Avoids passing None

Since there are three cases: specific, latest and not specified; there are two possible implementations:
1. None for latest and Some "" for not specified
2. None for not specified and Some "latest" for latest

I picked the first because it was the implicit one earlier. Do you thing we should go for the second or do you have another idea?

sabine: No, I think we should make sure that we never use `/p/PACKAGE`-URLs in the templates and that this also cannot happen by accident.

So, there should be:
- one function that renders a package URL (which always is either a versioned URL `/p/package/VERSION` or latest-URL `/p/PACKAGE/latest`)
- one redirect /p/PACKAGE -> /p/PACKAGE/latest

cuihtlauac: > No, I think we should make sure that we never use `/p/PACKAGE`-URLs in the templates and that this also cannot happen by accident.

Then we need to update dependencies and conflicts in package overview, it's happening there :-)

> * one function that renders a package URL (which always is either a versioned URL `/p/package/VERSION` or latest-URL `/p/PACKAGE/latest`)

Let's see what I can do.

cuihtlauac:
Subsumed by https://github.com/ocaml/ocaml.org/pull/999

Issue #890, opened 2023-02-06
Display Exercise Difficulty


cuihtlauac:
Exercises have a difficulty field: beginner, intermediate, advanced

It would be nice to:

- [x] Display difficulty
- [x] Filter exercises by difficulty

Beginner: no mark
Intermediate: one mark (☡)
Advanced: two marks (☡☡)

To honour tradition, use something as https://en.wikipedia.org/wiki/Bourbaki_dangerous_bend_symbol as difficulty mark

PR #955, opened 2023-02-24
Display exercise difficulty (#890)

YassineHaouzane:
This PR implements the first task of #890 displaying exercises difficulty.
We are displaying the difficulty marks next to the problem name.

### DIFF
--- a/src/ocamlorg_frontend/pages/problems.eml
+++ b/src/ocamlorg_frontend/pages/problems.eml
@@ -63,7 +63,7 @@ Learn_layout.render
     <div x-data="{statement: true}" id="<%s" problem.number %>
       >
       <div class="flex space-y-4 lg:space-y-0 flex-col lg:flex-row lg:justify-between">
-        <span aria-hidden="true" class="danger invisible lg:visible"><%s danger problem.difficulty %></span>
+        <span aria-hidden="true" class="danger invisible lg:visible" title="<%s Ood.Problem.difficulty_to_string problem.difficulty ^ " difficulty" %>"><%s danger problem.difficulty %></span>
         <h4 class="font-bold text-body-600">
           <%s problem.title %>
           <span aria-hidden="true" class="ml-2 visible lg:invisible" > <%s danger problem.difficulty %></span>
--- a/src/ocamlorg_frontend/pages/problems.eml
+++ b/src/ocamlorg_frontend/pages/problems.eml
@@ -21,6 +21,11 @@ in
   <%s! problem_sidebar_links_by_tag ~tag:"graph" "Graphs" %>
   <%s! problem_sidebar_links_by_tag "Miscellaneous" %>
 
+let danger = function
+  | `Beginner -> ""
+  | `Intermediate -> "☡"
+  | `Advanced -> "☡☡"
+  
 let render
 (problems : Ood.Problem.t list)
 =
@@ -48,12 +53,22 @@ Learn_layout.render
         problems is available on
         <a href="https://github.com/ocaml/ocaml.org/tree/main/data/problems">GitHub</a>.
       </p>
+      <p>
+        Every exercise has a difficulty level, ranging from beginner to advanced. 
+        This difficulty level is indicated with the symbol "☡".
+        The absence of marks represents the beginner level, one mark the intermediate level and two marks the advanced level.
+      </p>
     </div>
     <% problems |> List.iter (fun (problem : Ood.Problem.t) -> %>
     <div x-data="{statement: true}" id="<%s" problem.number %>
       >
       <div class="flex space-y-4 lg:space-y-0 lg:items-center flex-col lg:flex-row lg:justify-between">
-        <h4 class="font-bold text-body-600"><%s problem.title %></h4>
+        <span aria-hidden="true" class="danger invisible lg:visible"><%s danger problem.difficulty %></span>
+        <h4 class="font-bold text-body-600">
+          <%s problem.title %>
+          <span aria-hidden="true" class="visible lg:invisible" > <%s danger problem.difficulty %></span>
+        </h4>
+        <div class="w-0 h-0 overflow-hidden"><%s Ood.Problem.difficulty_to_string problem.difficulty ^ " difficulty" %></div>
         <div class="flex bg-body-600 bg-opacity-5 rounded-md p-1 items-center">
           <button
             class="px-3.5 h-9 font-medium flex space-x-2 items-center rounded-lg"
--- a/src/ocamlorg_frontend/pages/problems.eml
+++ b/src/ocamlorg_frontend/pages/problems.eml
@@ -66,7 +66,7 @@ Learn_layout.render
         <span aria-hidden="true" class="danger invisible lg:visible" title="<%s Ood.Problem.difficulty_to_string problem.difficulty ^ " difficulty" %>"><%s danger problem.difficulty %></span>
         <h4 class="font-bold text-body-600">
           <%s problem.title %>
-          <span aria-hidden="true" class="ml-2 visible lg:invisible" > <%s danger problem.difficulty %></span>
+          <span aria-hidden="true" class="ml-2 visible lg:invisible" title="<%s Ood.Problem.difficulty_to_string problem.difficulty ^ " difficulty" %>"> <%s danger problem.difficulty %></span>
         </h4>
         <div class="w-0 h-0 overflow-hidden"><%s Ood.Problem.difficulty_to_string problem.difficulty ^ " difficulty" %></div>
         <div class="flex bg-body-600 bg-opacity-5 rounded-md p-1 items-center">
--- a/src/ocamlorg_frontend/pages/problems.eml
+++ b/src/ocamlorg_frontend/pages/problems.eml
@@ -62,7 +62,7 @@ Learn_layout.render
     <% problems |> List.iter (fun (problem : Ood.Problem.t) -> %>
     <div x-data="{statement: true}" id="<%s" problem.number %>
       >
-      <div class="flex space-y-4 lg:space-y-0 lg:items-center flex-col lg:flex-row lg:justify-between">
+      <div class="flex space-y-4 lg:space-y-0 flex-col lg:flex-row lg:justify-between">
         <span aria-hidden="true" class="danger invisible lg:visible"><%s danger problem.difficulty %></span>
         <h4 class="font-bold text-body-600">
           <%s problem.title %>
--- a/src/ocamlorg_frontend/pages/problems.eml
+++ b/src/ocamlorg_frontend/pages/problems.eml
@@ -66,7 +66,7 @@ Learn_layout.render
         <span aria-hidden="true" class="danger invisible lg:visible"><%s danger problem.difficulty %></span>
         <h4 class="font-bold text-body-600">
           <%s problem.title %>
-          <span aria-hidden="true" class="visible lg:invisible" > <%s danger problem.difficulty %></span>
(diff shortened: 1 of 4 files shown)


src/ocamlorg_frontend/pages/problems.eml
```diff
@@ -21,6 +21,12 @@ in
   <%s! problem_sidebar_links_by_tag ~tag:"graph" "Graphs" %>
   <%s! problem_sidebar_links_by_tag "Miscellaneous" %>
 
+let difficulty_marks difficulty =
+  match difficulty with
```
cuihtlauac:
```suggestion
let difficulty_marks = function
```
Don't you prefer something like this?

cuihtlauac (changes requested):
Thanks for this contribution @YassineHaouzane, this is nice

I see two issues, which arguably are more relevant to the initial request than your code (which is fine BTW). 
- Sequences of parenthesis: “Binary Search Trees (Dictionaries) (☡)”
- Marks not visible enough, in books they're often written in the margin which eases understanding it's “meta”

YassineHaouzane:
Thank you for your fast feedback !
I have committed the changes that you requested.

These are just suggestions but here's what i think of the issues that you mentioned.

> Sequences of parenthesis: “Binary Search Trees (Dictionaries) (☡)”

I don't really know how to tackle this, maybe not displaying the difficulty marks in the title ?

> Marks not visible enough, in books they're often written in the margin which eases understanding it's “meta”

If it's because there's a lot of beginner problems maybe displaying one mark for beginner's exercises would be a solution or again, not displaying them in the title block.

src/ocamlorg_frontend/pages/problems.eml
```diff
@@ -21,6 +21,11 @@ in
   <%s! problem_sidebar_links_by_tag ~tag:"graph" "Graphs" %>
   <%s! problem_sidebar_links_by_tag "Miscellaneous" %>
 
+let difficulty_marks = function
+  | `Beginner -> ""
+  | `Intermediate -> "(☡)"
+  | `Advanced -> "(☡☡)"
```
cuihtlauac:
```suggestion
let danger = function
  | `Beginner -> ""
  | `Intermediate -> "☡"
  | `Advanced -> "☡☡"
```

src/ocamlorg_frontend/pages/problems.eml
```diff
@@ -53,7 +58,7 @@ Learn_layout.render
     <div x-data="{statement: true}" id="<%s" problem.number %>
       >
       <div class="flex space-y-4 lg:space-y-0 lg:items-center flex-col lg:flex-row lg:justify-between">
```
cuihtlauac:
```suggestion
      <div class="flex space-y-4 lg:space-y-0 lg:items-center flex-col lg:flex-row lg:justify-between">
                <span class="danger"><%s danger problem.difficulty %></span>

```

src/ocamlorg_frontend/pages/problems.eml
```diff
@@ -53,7 +58,7 @@ Learn_layout.render
...
-        <h4 class="font-bold text-body-600"><%s problem.title %></h4>
+        <h4 class="font-bold text-body-600"><%s problem.title %> <%s difficulty_marks problem.difficulty %></h4>
```
cuihtlauac:
```suggestion
        <h4 class="font-bold text-body-600"><%s problem.title %> </h4>
```

cuihtlauac (changes requested):
Somewhere in `src/ocamlorg_frontend/css/styles.css` I'd also add this:

```
.prose span.danger {
  position: absolute;
  width: 3em;
  margin-left: -2.9em;
  font-weight: 600;
  font-size: 18px;
  color: #111827;
  text-align: center;
}
```

That would allow the danger marks to be displayed in the left margin, without parenthesis.

What do you think?

YassineHaouzane:
The `danger` class works well in desktop but in mobile the marks overflow the layout.
I could add a media query that adjust the values of the margins and the font-size it would look like this :
image
For reference here's the css that I used in mobile: 
```css
@media (max-width: 1023px) {
  .prose span.danger {
    width: 1.5rem;
    margin-left: -1.5rem;
    margin-top: 2.0em;
    font-size: 10px;
  }
}
```

cuihtlauac:
Thanks @YassineHaouzane, that sounds like a sound approach, can you commit this?

YassineHaouzane:
I have pushed these changes.
I reversed the logic the desktop version is in the media query so that i can use the exact breakpoint value 
`max-width: 1024px` instead of `min-width: 1023px`.

Also i noticed an issue with this approach the layout is broken when then window has exactly a width of 1024px.
I think it's because the mobile breakpoints occurs in 2 steps when we reach 1024px we hide the sidebar in `learn_layout.eml` then when the width is under 1024px the tailwind's breakpoints gets applied.

I think in `learn_layout.eml` it should be : 
```html
  <div class="bg-white" x-data="{ open: false, sidebar: window.innerWidth >= 1024 && true, showOnMobile: false}" @resize.window="sidebar = window.innerWidth >= 1024"  x-on:close-sidebar="sidebar=window.innerWidth >= 1024 && true">
```
(i.e. all the checks should be inclusive to match the tailwind breakpoint)
What do you think @cuihtlauac ?

cuihtlauac (approved):
This PR looks good to me. 

Regarding the glitch you found when the width is exactly 1024px, can you file a fresh issue to describe it? I'd like to check if that issue doesn't appear on other pages

cuihtlauac (changes requested):
Screenshot_20230302-080538

The danger mark looks very small on my phone.

sabine (commented):
Thanks a lot for the contribution @YassineHaouzane!

I see these issues here:

1. On mobile, as you correctly assessed, there is no space in the margins to put the difficulty level.
2. It is not obvious what the symbol means to anyone who has not read books that make use of it.
3. The meaning of the symbol is not discernible via screen reader. We're currently trying to improve overall accessibility of the site.

I recommend...
* for (1) we need to place the difficulty marks in a different position (e.g. after the title), both on `md` and `sm` screen size
* for (2) there needs to be a paragraph at the start of the exercises section that explains the meaning of the bend symbol. Additionally, to help people discover the meaning when they see the symbols, a title attribute that spells out the meaning in human readable terms, e.g. "Intermediate difficulty", "Advanced difficulty" is likely to help discoverability.
* for (3) one solution is to spell out the human readable meaning (could be visually hidden by placing it into a div with height/width set to zero and `overflow: hidden`) and to hide the symbols from screen readers via `aria-hidden="true"`. Another solution would be to find a design where both the sighted and the sight-impaired user both get the difficulty rating spelled out (which make remedies regarding (2) obsolete as there are no longer any concerns about discoverability of meaning of the symbols).

YassineHaouzane:
>Regarding the glitch you found when the width is exactly 1024px, can you file a fresh issue to describe it? I'd like to check if that issue doesn't appear on other pages

I created an issue: #960 

>The danger mark looks very small on my phone.

This is intended otherwise there is not enough space in the margins with a `font-size` of 18px : 

image

>for (1) we need to place the difficulty marks in a different position (e.g. after the title), both on md and sm screen size

I initially placed the difficulty marks after the title but this involves sequences of parentheses with title's name that ends with parentheses. Should I still do that ? Or display the difficulty marks without parentheses ?

image

YassineHaouzane:
* I removed the parentheses from the difficulty marks.

* We are now displaying the difficulty marks in the title only in `md` and `sm` screens otherwise they will be displayed in the margins.

* I added a paragraph at the beginning of the exercise section that explains the meaning of the difficulty marks.

* I added a hidden div that contains the human readable difficulty level for screen readers.

src/ocamlorg_frontend/pages/problems.eml
```diff
@@ -48,12 +53,22 @@ Learn_layout.render
...
+      <p>
+        Every exercise has a difficulty level, ranging from beginner to advanced. 
+        This difficulty level is indicated with the symbol "☡".
+        The absence of marks represents the beginner level, one mark the intermediate level and two marks the advanced level.
+      </p>
     </div>
     <% problems |> List.iter (fun (problem : Ood.Problem.t) -> %>
     <div x-data="{statement: true}" id="<%s" problem.number %>
       >
       <div class="flex space-y-4 lg:space-y-0 lg:items-center flex-col lg:flex-row lg:justify-between">
```
sabine:
```suggestion
      <div class="flex space-y-4 lg:space-y-0 flex-col lg:flex-row lg:justify-between">
```

|before|after|
|-|-|
|Screenshot 2023-03-15 at 12-27-23 Exercises|Screenshot 2023-03-15 at 12-27-38 Exercises align difficulty symbol at the top|

src/ocamlorg_frontend/pages/problems.eml
```diff
@@ -48,12 +53,22 @@ Learn_layout.render
...
-        <h4 class="font-bold text-body-600"><%s problem.title %></h4>
+        <span aria-hidden="true" class="danger invisible lg:visible"><%s danger problem.difficulty %></span>
+        <h4 class="font-bold text-body-600">
+          <%s problem.title %>
+          <span aria-hidden="true" class="visible lg:invisible" > <%s danger problem.difficulty %></span>
```
sabine:
```suggestion
          <span aria-hidden="true" class="ml-2 visible lg:invisible" > <%s danger problem.difficulty %></span>
```

|before|after|
|-|-|
|Screenshot 2023-03-15 at 12-18-17 Exercises|Screenshot 2023-03-15 at 12-18-25 Exercises add spacing between title and symbol|

sabine (commented):
Sorry we took so long to get back to this @YassineHaouzane. I think you addressed all the constraints and there were just two alignment/spacing issues.

So thanks a lot for the help! :+1:

src/ocamlorg_frontend/pages/problems.eml
```diff
@@ -48,12 +53,22 @@ Learn_layout.render
...
+        The absence of marks represents the beginner level, one mark the intermediate level and two marks the advanced level.
+      </p>
     </div>
     <% problems |> List.iter (fun (problem : Ood.Problem.t) -> %>
     <div x-data="{statement: true}" id="<%s" problem.number %>
       >
-      <div class="flex space-y-4 lg:space-y-0 lg:items-center flex-col lg:flex-row lg:justify-between">
-        <h4 class="font-bold text-body-600"><%s problem.title %></h4>
+      <div class="flex space-y-4 lg:space-y-0 flex-col lg:flex-row lg:justify-between">
+        <span aria-hidden="true" class="danger invisible lg:visible"><%s danger problem.difficulty %></span>
```
sabine:
```suggestion
        <span aria-hidden="true" class="danger invisible lg:visible" title="<%s Ood.Problem.difficulty_to_string problem.difficulty ^ " difficulty" %>"><%s danger problem.difficulty %></span>
```

Maybe we add a title too, so that people who don't read (like me) can still discover by hovering.

src/ocamlorg_frontend/pages/problems.eml
```diff
@@ -48,12 +53,22 @@ Learn_layout.render
...
+        <h4 class="font-bold text-body-600">
+          <%s problem.title %>
+          <span aria-hidden="true" class="ml-2 visible lg:invisible" > <%s danger problem.difficulty %></span>
```
sabine:
```suggestion
          <span aria-hidden="true" class="ml-2 visible lg:invisible" title="<%s Ood.Problem.difficulty_to_string problem.difficulty ^ " difficulty" %>"> <%s danger problem.difficulty %></span>
```

sabine (approved):

PR #958, opened 2023-03-01
No crunching

sabine:
Patch removes crunching of assets and data - compute static file digests for assets.

On my machine:
before:
```
$ time make
opam exec -- dune build --root .
Done in 382ms.
Done in 1421ms.
                                          
real	0m12,948s
user	0m40,899s
sys	0m7,429s
```

after:
```
$ time make
opam exec -- dune build --root .
Done in 344ms.
Done in 1286ms.
                                          
real	0m8,282s
user	0m28,940s
sys	0m6,731s
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -31,6 +31,8 @@ RUN chmod -R 755 /var
 COPY --from=build /home/opam/package.state /var/package.state
 COPY --from=build /home/opam/opam-repository /var/opam-repository
 COPY --from=build /home/opam/_build/default/src/ocamlorg_web/bin/main.exe /bin/server
+COPY --from=build /home/opam/_build/default/asset _build/default/asset
+COPY --from=build /home/opam/_build/default/data/media _build/default/data/media
 
 COPY playground/asset playground/asset
 
--- a/src/ocamlorg_static/ocamlorg_static.ml
+++ b/src/ocamlorg_static/ocamlorg_static.ml
@@ -1,18 +1,36 @@
 let of_url_path = File.of_url_path
 
+let read_file file =
+  let open Lwt.Syntax in
+  Lwt.catch
+    (fun () ->
+      Lwt_io.(with_file ~mode:Input file) (fun channel ->
+          let* content = Lwt_io.read channel in
+          Some content |> Lwt.return))
+    (fun _exn ->
+      Dream.log "failed to read file %s" file;
+      None |> Lwt.return)
+
 module Media = struct
-  let digest = Media.hash
-  let read = Media.read
+  let file_root = Sys.getcwd () ^ "/_build/default/data/media/"
+  let url_root = "/media"
+  let digest = Media_digests.digest
+
+  let read filepath =
+    let file = Filename.concat file_root filepath in
+    read_file file
 end
 
 module Asset = struct
-  let digest = Asset.hash
-  let read = Asset.read
+  let file_root = Sys.getcwd () ^ "/_build/default/asset/"
+  let digest = Asset_digests.digest
 
-  (* given the path of a file from `assets.ml`: 1. looks up the file's digest in
-     and 2. returns the corresponding digest URL for use in templates *)
+  let read filepath =
+    let file = Filename.concat file_root filepath in
+    read_file file
+
   let url filepath =
-    let digest = Option.map Dream.to_base64url (Asset.hash filepath) in
+    let digest = Option.map Dream.to_base64url (digest filepath) in
     if digest = None then
       raise
         (Invalid_argument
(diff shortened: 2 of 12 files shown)


avsm:
If you do get rid of crunching, then the utilities you've written could live in the new https://github.com/ocaml-web org that @samoht has created (for the purpose of giving a home to the ancillary utilities and libraries that drive this website).

sabine:
I didn't know about that repository. That sounds like an excellent place for this.

avsm:
We just created it last week, so not actually been activated yet!

Dockerfile
```diff
@@ -23,6 +23,8 @@ RUN apk update && apk add --update libev gmp git
 RUN chmod -R 755 /var
 
 COPY --from=build /home/opam/_build/default/src/ocamlorg_web/bin/main.exe /bin/server
+COPY --from=build /home/opam/_build/default/asset _build/default/asset
+COPY --from=build /home/opam/_build/default/data/media _build/default/data/media
```
sabine, line 35: I need to test whether this is correct. I was thinking, I should probably try to `docker build` and `docker run` the docker container. However, this is what I get:

```
$ sudo docker build .
Sending build context to Docker daemon  52.88MB
Step 1/22 : FROM ocaml/opam:alpine-3.17-ocaml-4.14 as build
 ---> d6b3acb10d6c
Step 2/22 : RUN sudo apk update && sudo apk add --update libev-dev openssl-dev gmp-dev oniguruma-dev
 ---> Running in 07a17d876c02
fetch https://dl-cdn.alpinelinux.org/alpine/v3.17/main/x86_64/APKINDEX.tar.gz
ERROR: https://dl-cdn.alpinelinux.org/alpine/v3.17/main: temporary error (try again later)
fetch https://dl-cdn.alpinelinux.org/alpine/v3.17/community/x86_64/APKINDEX.tar.gz
ERROR: https://dl-cdn.alpinelinux.org/alpine/v3.17/community: temporary error (try again later)
fetch https://dl-cdn.alpinelinux.org/alpine/edge/main/x86_64/APKINDEX.tar.gz
ERROR: https://dl-cdn.alpinelinux.org/alpine/edge/main: temporary error (try again later)
fetch https://dl-cdn.alpinelinux.org/alpine/edge/community/x86_64/APKINDEX.tar.gz
ERROR: https://dl-cdn.alpinelinux.org/alpine/edge/community: temporary error (try again later)
fetch https://dl-cdn.alpinelinux.org/alpine/edge/testing/x86_64/APKINDEX.tar.gz
ERROR: https://dl-cdn.alpinelinux.org/alpine/edge/testing: temporary error (try again later)
v3.17.2-24-g9bb746ae153 [https://dl-cdn.alpinelinux.org/alpine/v3.17/main]
v3.17.2-35-g33db3ffa533 [https://dl-cdn.alpinelinux.org/alpine/v3.17/community]
v20230208-795-g69e308aa808 [https://dl-cdn.alpinelinux.org/alpine/edge/main]
v20230208-798-ge0d946eda73 [https://dl-cdn.alpinelinux.org/alpine/edge/community]
v20230208-799-gb4c6cbb034d [https://dl-cdn.alpinelinux.org/alpine/edge/testing]
5 errors; 40350 distinct packages available
```

sabine: Ok my bad, updating the Docker version fixed that :laughing:

sabine:
Status: since serving the files from memory is quite useful for production, this patch either needs to
1) build without crunching for development purposes, and build with crunching for production
2) include a http server that serves the files in such a way that they are cached in memory in production

tmattio:
> build without crunching for development purposes, and build with crunching for production

That's what we did at one point IIRC, but we changed that because the added maintenance on the build system configuration wasn't worth the time it saved for build during development.

Is serving from the filesystem really _that_ bad in production? (especially if we have a caching system in place for assets)

cuihtlauac:
I also wonder about the dune set up. Crunched stuff is mostly static. Are we sure dune isn't doing useless re-processing of untouched files?

sabine:
> Is serving from the filesystem really that bad in production?

While it is slower to read from the filesystem, there's also OS-level file system caching which mitigates this. It is unclear whether we will run into problems with the server's network bandwidth first or with the speed of the filesystem access.

> I also wonder about the dune set up. Crunched stuff is mostly static. Are we sure dune isn't doing useless re-processing of untouched files?

Dune seems to be looking at the deps and only rebuild if the deps change, right?

But even if not... linking the crunched files into the binary happens every build and appears to be a more time-consuming operation than computing the digests.

Considering that
(1) proper benchmarks would take up valuable time that can be spent on implementing a caching layer or other more useful features, and 
(2) it's very likely that serving from file system will be fast enough for practical purposes,
I'm inclined to merge this because it speeds up the build and removes a dependency.

src/ocamlorg_static/ocamlorg_static.ml
```diff
@@ -1,18 +1,36 @@
...
+          let* content = Lwt_io.read channel in
+          Some content |> Lwt.return))
+    (fun _exn ->
+      Dream.log "failed to read file %s" file;
+      None |> Lwt.return)
+
 module Media = struct
-  let digest = Media.hash
-  let read = Media.read
+  let file_root = Sys.getcwd () ^ "/_build/default/data/media/"
```
gpetiot, line 15: Could we use `Fpath.t` paths instead of strings?

sabine: In principle yes, but since there is no need for cross-platform compatibility it feels more like obfuscating the code than an actually helpful abstraction.

edwintorok:
FYI this got reverted because it caused all documentation pages to reply with HTTP 500 on ocaml.org: https://github.com/ocaml/ocaml.org/issues/1170#issuecomment-1545307948

PR #964, opened 2023-03-03
Make HTML on package overview/search/autocomplete more semantic

sabine:
Improves accessibility without changing styles.


cuihtlauac (approved):
This looks good to me.

PR #969, opened 2023-03-07
Simplify and uniformize enumerated kinds of data

cuihtlauac:
- Isolate in modules
- Uniform parsing, of_string and of_yaml
- Define metadata field with ad-hox type


sabine (approved):
LGTM

PR #975, opened 2023-03-13
Update Citrix entry, and add XenServer job

edwintorok:
Rename Citrix entry to XenServer. XenServer used to be part of Citrix (and as far as I know the only group using OCaml within Citrix), however after the merger Citrix is now part of Cloud Software Group, and XenServer is its own business unit inside CSG. As far as I know there is no OCaml usage left within Citrix itself, so replace this with XenServer.

(XenServer also got renamed to Citrix Hypervisor, and now it is being renamed back to XenServer, although the new XenServer website is still under development).

We also have a job opening which involves mostly OCaml work (according to GitHub [our repo is 84% OCaml code](https://github.com/xapi-project/xen-api), so add an entry for that.

Draft PR because I'm not sure how to check how this page would render.


cuihtlauac (changes requested):
On Citrix + Tibco -> Cloud:

Thanks for updating the text. That's great. And needed, those mergers, acquisitions and rebrandings are always so confusing, deBruijn indices would be so easier :-)

We need to do our homework. The text on each industrial user isn't displayed anywhere as it used to in v2.ocaml.org

On job post:

Up to my understanding, the company is Cloud Software Group. Inside this, XenServer is a business unit and a product. I suggest:

- company: Cloud Software Group, XenServer

cuihtlauac:
Corresponding issue on industrial users descriptions:

https://github.com/ocaml/ocaml.org/issues/976

edwintorok:
I've asked about this internally, and it is preferred to emphasize the business unit first. Would "XenServer, Cloud Software Group" be a suitable alternative (remove the 'part of' from the text)

cuihtlauac (approved):

edwintorok:
Thanks for merging my PR. I'm not familiar with the ocaml.org workflow, will the change eventually be deployed automatically, or is there something else that I should do to trigger that?

Looking here I see that a deployment seems to have successfully built an image but then timed out https://deploy.ci.ocaml.org/job/2023-03-13/210954-ocluster-build-4d52b9

cuihtlauac:
You're welcome. I love merging job offers. They are my favourites. 

Sorry, I didn't notice it didn't go online as it should have. Once merged, PR go live automatically.

cuihtlauac:
That must have been a transient issue. It's now online.

Issue #42, opened 2021-07-28
Integrate http://check.ocamllabs.io/


tmattio:
Can also integrate checks in package overview - e.g https://community.chocolatey.org/packages/KB2919355#testingResults

cc @patricoferris

cuihtlauac:
Show “Latest Successful Build Using: <ocaml compiler version>”

![Screenshot_2022-12-06_14-23-31](https://user-images.githubusercontent.com/1917592/205924600-4a1e7c16-5f74-4500-97b9-7c505bdd47ba.png)

PR #977, opened 2023-03-13
Expose check.ci.ocaml.org in package overview

cuihtlauac:
This is a work in progress. The goal is to expose the opam package build results stored in check.ocamllabs.io in the corresponding package overview page

- [X] Display "OCaml 5 Ready" in package overview if build good
- [X] Process build result json
- [X] Store build statuses in state
- [X] Scrape json from check.ocamllabs.io
- [X] Expose package build check status in GraphQL API

To test, store the build statuses json in the cache dir:

```shell
wget https://check.ci.ocaml.org/api/v1/latest/packages -P ~/.cache/ocamlorg/
```

Corresponding Issue: https://github.com/ocaml/ocaml.org/issues/42

### DIFF
--- a/src/ocamlorg_package/lib/check.ml
+++ b/src/ocamlorg_package/lib/check.ml
@@ -1,0 +1,44 @@
+open Ocamlorg.Import
+
+let error kind str =
+  Error (`Msg (Printf.sprintf "Unrecognized %s: %s" kind str))
+
+let ( let* ) = Result.bind
+let ( <@> ) f = Result.fold ~ok:Result.map ~error:(fun e _ -> Error e) f
+let ok_cons f x u = Ok List.cons <@> f x <@> u
+
+type status = string option
+
+module Json = struct
+  let error kind json =
+    Error
+      (`Msg
+        (Printf.sprintf "Unrecognized %s: %s" kind (Yojson.Safe.to_string json)))
+
+  let status_of_string = function
+    | "good" -> Ok None
+    | ("bad" | "internal-failure" | "not-available" | "partial") as s ->
+        Ok (Some s)
+    | str -> Error (`Msg (Printf.sprintf "Unrecognized status: %s" str))
+
+  let to_build = function
+    | `Assoc [ ("compiler", `String version); ("status", `String status) ] ->
+        let* status = status_of_string status in
+        Ok (version, status)
+    | json -> error "build" json
+
+  let to_release = function
+    | `Assoc [ ("name", `String release); ("statuses", `List builds) ] ->
+        let* builds = List.fold_right (ok_cons to_build) builds (Ok []) in
+        Ok (release, builds)
+    | json -> error "release" json
+
+  let to_repo = function
+    | `List releases ->
+        List.fold_left (Fun.flip (ok_cons to_release)) (Ok []) releases
+    | json -> error "repo" json
+
+  let of_string json =
+    let of_list u = u |> List.to_seq |> String.Map.of_seq in
+    json |> to_repo |> Result.map of_list
+end
--- a/src/ocamlorg_package/lib/check.ml
+++ b/src/ocamlorg_package/lib/check.ml
@@ -1,4 +1,4 @@
-open Ocamlorg.Import
+open Import
 
 let error kind str =
   Error (`Msg (Printf.sprintf "Unrecognized %s: %s" kind str))
--- a/src/ocamlorg_package/lib/info.ml
+++ b/src/ocamlorg_package/lib/info.ml
@@ -2,7 +2,7 @@ type url = { uri : string; checksum : string list }
 
 (* This is used to invalidate the package state cache if the type [Info.t]
    changes. *)
-let version = "3"
+let version = "4"
 
 type t = {
   synopsis : string;
--- a/src/ocamlorg_package/lib/info.ml
+++ b/src/ocamlorg_package/lib/info.ml
@@ -187,13 +187,13 @@ let of_opamfiles
     in
     List.fold_left f OpamPackage.Set.empty names
   in
-  Logs.info (fun f -> f "Dependencies...");
+  Logs.info (fun f -> f "Opam repo: Dependencies...");
   let dependencies = get_dependency_set packages opams in
-  Logs.info (fun f -> f "Reverse dependencies...");
+  Logs.info (fun f -> f "Opam repo: Reverse dependencies...");
   let* rev_deps = rev_depends dependencies in
-  Logs.info (fun f -> f "Publication dates...");
+  Logs.info (fun f -> f "Opam repo: Publication dates...");
   let* timestamps = Opam_repository.create_package_to_timestamp () in
-  Logs.info (fun f -> f "Generate package info");
+  Logs.info (fun f -> f "Opam repo: Generate package info");
   Lwt_fold.package_name_map
     (fun name vmap acc ->
       let+ vs =
--- a/src/ocamlorg_package/lib/ocamlorg_package.ml
+++ b/src/ocamlorg_package/lib/ocamlorg_package.ml
@@ -122,7 +122,7 @@ let get_latest' packages name =
          in
          { version; info; name })
 
-let update_repo commit t =
+let update_repo t commit =
   let open Lwt.Syntax in
   Logs.info (fun m -> m "Opam repo: Update");
   t.opam_repository_commit <- Some commit;
@@ -134,12 +134,11 @@ let update_repo commit t =
   let+ stats = Statistics.compute packages in
   t.packages <- packages;
   t.stats <- Some stats;
-  Logs.info (fun m -> m "Loaded %d packages" (Name.Map.cardinal packages))
   Logs.info (fun m ->
       m "Opam repo: Loaded %d packages" (Name.Map.cardinal packages));
   true
 
-let update_build_check (data, digest) t =
+let update_build_check t (data, digest) =
   Logs.info (fun m -> m "Build check: Update");
   let json = Yojson.Safe.from_string data in
   let build_check = Check.Json.of_string json in
@@ -176,12 +175,14 @@ let http_get url =
 
 let ( let>& ) opt some = Option.fold ~none:(Lwt.return false) ~some opt
 
+let if_fresher new_x old_x = if Some new_x = old_x then None else Some new_x
+
 let maybe_update_repo state =
   let open Lwt.Syntax in
   let* commit = Opam_repository.(if exists () then pull else clone) () in
   Logs.info (fun m -> m "Opam repo: Commit %s" commit);
-  let>& commit = Option.update_with commit state.opam_repository_commit in
-  update_repo commit state
+  let>& commit = if_fresher commit state.opam_repository_commit in
+  update_repo state commit
 
 let maybe_update_build_check state =
   let open Lwt.Syntax in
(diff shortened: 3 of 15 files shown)


tmattio:
By the way, looping back to this, similar to sherlocode, we'd have to deploy the service as part of the OCaml.org infrastructure to use it. @cuihtlauac do you want to open an issue on `ocaml/infrastructure` to start discussions on this?

cuihtlauac:
https://github.com/ocaml/infrastructure/issues/40

src/ocamlorg_frontend/dune
```diff
@@ -1,6 +1,12 @@
 (library
  (name ocamlorg_frontend)
- (libraries ocamlorg.global dream ood timedesc ocamlorg_static))
+ (libraries
+  ocamlorg.global
+  dream
+  ood
+  timedesc
+  ocamlorg_static
+  ocamlorg_package))
```
sabine: `ocamlorg_frontend` has its own `Package` module that represents what the package information looks like, for rendering purposes. I think it would be better to add the build status, as well as its `to_string` function to `Ocamlorg_frontend.Package`, and refrain from adding a dependency here.

tmattio: Ah yes! That was a design principle used when creating the website: the front end should be self-contained and not depend on anything, ensuring that it remains only a small wrapper around HTML and that all the logic lives elsewhere.

cuihtlauac: Thanks, @sabine for pointing this out. Indeed, there's a situation here.

Here is a diagram of the internal dependencies inside ocamlorg.

https://drive.google.com/file/d/1_kBOQx-0naKhGcVc0JL1kdevL0fA8_IW/view?usp=sharing

Drawing an arrow from `package` to `frontend` isn't great. I did it to provide to the `frontend` the build check status (from `check.ci.ocaml.org`) as a datatype rather than a `string`, which would have put logic in `frontend`, which would have been unquestionably wrong. 

I don't think using `package.ml` from the `frontend` is a great solution either. If I was to put the definition of `Status.t` that is currently in `ocamlorg_package/lib/build.ml` there, it would revert the offending arrow. 

Therefore, I see two ways to go (but there are certainly more):

1. Use a `Stdlib` predefined type to encode the build check status `Build.Status.t`. For instance, the JSON coming from `check.ci.ocaml.org` could be parsed such that the statuses would have type `string option` where `None` would mean success and `Some cause` would mean failure.
1. Move that build check status type up to either `ood`, `global` or something else which would lie at the side of those in the diagram.

To me, the second one makes more sense. The first one feels bizarre. Both `package` and `frontend` already rely on types defined in `odd`. Data coming from `check.ci.ocaml.org` is yet another kind of data that needs to be dealt with. The only difference is encoding: JSON instead of Yaml.

What's your view on this?

sabine: `Ocamlorg_frontend.Package` provides only rendering-related logic and types. So we indeed cannot move the definition of `Status.t` - as it currently is - in there. The current definition of `Status.t` contains both data import logic and rendering logic. So that would have to be split up into these two parts.

(2) sounds fine to me because this is indeed just another data import from an external source.

I think we should disentangle the different data sources, in general.

As I see it, `Ocamlorg_package` is on the same "level" as `ood`: it's a module that provides data to ocamlorg web. However, among all the data modules, it's the most convoluted one because it imports from several data sources: (1) from opam repository, (2) from the documentation pipeline.

In a traditional Model-View-Controller architecture, `ood`, `package` would be models, while `web` is the controller, and `frontend` is the view.

The special situation about `ood` is that there's no need for `frontend` to repeat the types in `ood` because they're already in the shape that we need to render a template. The architecturally clean thing to do here would be to only share the types of `ood` with the frontend, but not the values - the controller is who determines which values get rendered.

sabine: Tiny addendum: the critical thing not to share with the frontend IMO is all the data import logic

avsm:
This is a great feature to have! I do think we need to get migrate the service [check.ocamllabs.io](http://check.ocamllabs.io/) -> [check.ocaml.org](http://check.ocaml.org/).

One of the features I liked about the (now deprecated) [ocaml-ci tool](https://github.com/ocaml/infrastructure/wiki/Using-the-opam-ci-tool) is that it separated the results database from the service, by pushing the results regularly to a [git repository](https://github.com/ocaml/obi-logs/tree/builds). This way, any tool could clone the repository and work over the results. It would simplify the OCaml.org website to do this as well, as it would just need to watch a git repository for updated results, and services like [check.ocaml.org](http://check.ocaml.org/) could push to it (along with other ones in the future as well).

cuihtlauac:
> It would simplify the OCaml.org website to do this as well, as it would just need to watch a git repository for updated results, and services like [check.ocaml.org](http://check.ocaml.org/) could push to it (along with other ones in the future as well).

I don't think processing a git repo or parsing JSON doesn't make such a big difference. What would help is a subscribe mechanism. We'd be pushed instead of watching, simpler and nicer to the planet.

src/global/type.ml
```diff
@@ -0,0 +1,21 @@
+module Build = struct
+  type status = Bad | Internal_failure | Not_available | Partial | Good
+
+  let status_of_string = function
+    | "bad" -> Ok Bad
+    | "internal-failure" -> Ok Internal_failure
+    | "not-available" -> Ok Not_available
+    | "partial" -> Ok Partial
+    | "good" -> Ok Good
+    | str -> Error (`Msg (Printf.sprintf "Unreconized status: %s" str))
```
gpetiot: "Unrecognized"

src/ocamlorg_package/lib/check.ml
```diff
@@ -0,0 +1,34 @@
+open Ocamlorg.Import
+
+let error kind str = Error (`Msg (Printf.sprintf "Unreconized %s: %s" kind str))
```
gpetiot: "Unrecognized"

src/ocamlorg_package/lib/check.ml
```diff
@@ -0,0 +1,34 @@
...
+let ( let* ) = Result.bind
+let ( <@> ) f = Result.fold ~ok:Result.map ~error:(fun e _ -> Error e) f
+let ok_cons f x u = Ok List.cons <@> f x <@> u
+
+module Json = struct
+  let error kind json =
+    Error
+      (`Msg
+        (Printf.sprintf "Unreconized %s: %s" kind (Yojson.Safe.to_string json)))
```
gpetiot: "Unrecognized", also could re-use `error` defined above

src/ocamlorg_package/lib/ocamlorg_package.mli
```diff
@@ -193,3 +193,9 @@ val search : ?sort_by_popularity:bool -> state -> string -> t list
 
 val featured : state -> t list option
 (** A list of packages to highlight on the Packages page. *)
+
+module Build : sig
+  val find : state -> t -> (string * Ocamlorg.Type.Build.status) list
+  (** Returns a list of compiler version and build check status pairs
+      correponding to the package release. *)
```
gpetiot: "corresponding"

gpetiot (commented):
Just spotted a few typos, but looks good to me otherwise.

src/ocamlorg_package/lib/check.ml
cuihtlauac: You missed the 3rd one there: https://github.com/ocaml/ocaml.org/blob/ffced3d38232efc5b3ddb80b919e444cc457d4cb/src/global/type.ml#L10 ;-)

avsm:
> I don't think processing a git repo or parsing JSON doesn't make such a big difference. 

It makes a difference in that you don't need the check service to be available and online to build the website. Having all the inputs in the filesystem is very helpful to have a more deterministic infrastructure.  You can build a subscription over that by just diffing old and new trees.

tmattio:
Rebased and pushed. Only missing the UI changes from what it seems @cuihtlauac.

Re. calling check.ci.ocaml.org service vs having the state on the filesystem: I'm personally happy to go with the approach in the current PR:

- The website can be built locally even if there is no access to check.ci.ocaml.org, we could display a placeholder in the UI if the service can't be reached. 
- We're already relying on external services for the package documentation. This just adds another API call when loading the package doc, which I think is reasonable
- The data is going to grow outdated quickly and doesn't have "historical value"

If check.ci.ocaml.org was pushing to a git repository we could definitely use that instead, but I suggest we don't let that block the current PR: it is going to be very useful to have these "OCaml 5-ready" status badges!

src/ocamlorg_package/lib/info.ml
```diff
@@ -2,7 +2,7 @@ type url = { uri : string; checksum : string list }
 
 (* This is used to invalidate the package state cache if the type [Info.t]
    changes. *)
-let version = "2"
+let version = "3"
```
tmattio: Let's not change the package state, we can afford to make another API call when loading the package docs IMO and it's closer to the current information flow since we rely on the external package doc service.

cuihtlauac: Without this (and if `package.state` is not deleted), there's a segmentation fault at startup. Marshalling of `Ocamlorg_package.state`, same problem as #1229.

tmcgilchrist:
> I do think we need to get migrate the service [check.ocamllabs.io](http://check.ocamllabs.io/) -> [check.ocaml.org](http://check.ocaml.org/).

That service got moved to [check.ci.ocaml.org](https://check.ci.ocaml.org).

> It makes a difference in that you don't need the check service to be available and online to build the website. Having all the inputs in the filesystem is very helpful to have a more deterministic infrastructure. You can build a subscription over that by just diffing old and new trees.

I like this idea of having a store of the checks results, even if we put a web service in front of it. We would like to make check.ci.ocaml.org work more incrementally rather than as a batch operation as it does now.

src/ocamlorg_package/lib/info.ml
tmattio: Yes, I meant let's not change the package state type so we don't have to bump the version. We can make the queries when loading the page instead

src/ocamlorg_package/lib/ocamlorg_package.ml
```diff
@@ -17,6 +17,8 @@ type state = {
   mutable packages : Info.t Version.Map.t Name.Map.t;
   mutable stats : Statistics.t option;
   mutable featured : t list option;
+  mutable build_check_digest : Digest.t option;
+  mutable build_check : (string * Check.status) list String.Map.t;
```
tmattio, line 21: Can we expose this in the GraphQL API also?

cuihtlauac: Why not. The only trouble is we've overloaded `/api` route at some point, it no longer serves graphql requests but redirects to https://v2.ocaml.org/api/

Both list and digest or only list?

cuihtlauac: Screenshot_2023-06-02_11-24-24

cuihtlauac: Screenshot_2023-06-02_13-10-19

tmattio:
As per our last offline discussions, it seems that the functionalities provided by this PR can be implemented by looking at the opam metadata instead of calling an external service.

I believe we agree on introducing this change when we find a strong use case for it. Should we close the PR in the meantime? cc @sabine

sabine:
Yes, it makes sense to close. I think a CI-based setup would likely make sense for "Windows compatible" badges for the upcoming opam 2.2 release. What do you think?

tmattio:
> I think a CI-based setup would likely make sense for "Windows compatible" badges for the upcoming opam 2.2 release. What do you think?

That sounds like an excellent use case yes!

PR #978, opened 2023-03-14
Keyboard controls on package search dropdown

### LINKED ISSUE #944: Use arrows keys in the dropdown search menu
First, thanks a lot for #939! It's super nice in a UX point of view, it saves several clicks and page loading, eg to go to the docs of a known package.

However, since we were just typing, it would be an improvement if we could navigate the dropdown menu using the arrows keys instead of the mouse.

sabine:
Patch adds keyboard controls to the package search dropdowns.

When there are more than 5 search results, pressing the enter key in the dropdown sends you to the full list of results. When there are 5 or less search results, pressing the enter key in the dropdown sends you to the first search result.

Resolves #944.

What this patch does not do:
- doesn't keyboard-navigate to Standard Library API link (could be added, though, but this here is pretty much the simplest implementation, so I'd like to have this as a reference point in the repo for adding keyboard-navigation to elements) - we will refine when there's enough feedback as this is going to be probably the most used control on the site
- css styles are not final. Claire prepared a really nice UI that a later patch should apply

|before|after|
|-|-|
|Screenshot from 2023-03-14 12-58-24|Screenshot from 2023-03-14 12-58-20|

### DIFF
--- a/src/ocamlorg_frontend/pages/packages_autocomplete_fragment.eml
+++ b/src/ocamlorg_frontend/pages/packages_autocomplete_fragment.eml
@@ -1,12 +1,18 @@
 let render
 ~search
+~total
 (packages : Package_intf.package list)
 =
 <div class="mx-2 mb-2">
   <span class="font-semibold">Packages:</span>
-  <% if List.length packages = 0 then ( %>
+  <% if total = 0 then ( %>
   <p class="break-words">
     We didn't find a match for "<%s search %>".
+  </p>
+  <% ) else ( %>
+  <p class="break-words">
+    <%s (if total = List.length packages then "" else "Showing top " ^ string_of_int (List.length packages) ^ " out of ")
+    ^ string_of_int total ^ (if total = 1 then " search result" else " search results") %>:
   </p>
   <% ); %>
 
--- a/src/ocamlorg_frontend/pages/packages_autocomplete_fragment.eml
+++ b/src/ocamlorg_frontend/pages/packages_autocomplete_fragment.eml
@@ -1,3 +1,41 @@
+(*
+  Package autocomplete works like this:
+  1. user types search terms into the search input
+  2. htmx fetches the fragment below (see function [render])
+     and inserts it into the HTMLElement selected by [target_sel].
+     While htmx's HTTP request is in progress, the HTMLElement
+     with selected by [indicator_sel] is shown.
+  3. user uses arrow keys to navigate the top 5 search results,
+     which present quick access to both the package overview page
+     and the documentation.
+
+  Note:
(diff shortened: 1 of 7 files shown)


src/ocamlorg_frontend/pages/packages_autocomplete_fragment.eml
```diff
@@ -1,31 +1,84 @@
...
+     with selected by [indicator_sel] is shown.
+  3. user uses arrow keys to navigate the top 5 search results,
+     which present quick access to both the package overview page
+     and the documentation.
+
+  Note:
+  - when there are at most five search results, pressing the
+    Enter key in the input field sends you to the overview
+    page of the first search results
+ *)
```
cuihtlauac: Are there any other places where this could be stored? Some doc, the commit message? Not sure about this

src/ocamlorg_frontend/pages/packages_autocomplete_fragment.eml
```diff
@@ -1,31 +1,84 @@
...
+   and the search results HTML fragment (see function [render]). *)
+let form_attributes =
+  {js| x-data="{ row: null, col: 0, max: 0, total: 0 }"
+       @submit="if (row === null && total <= 5) row = 0; if (row !== null) { window.location = document.getElementById('package-autocomplete-'+row+'-'+col).getAttribute('href'); $event.stopPropagation(); $event.preventDefault(); return false }" |js}
+
+let input_attributes ~target_sel ~indicator_sel =
+  {js|  @keyup.down="if (row === null) { row = 0; col = 0; } else { row +=1; if (row > max) { row = max } }"
+        @keyup.up="if (row !== null) { row -=1; if (row < 0) { row = null } }"
+        @keyup.right="col = (col + 1)%2"
+        @keyup.left="col = (col + 1)%2"
```
cuihtlauac: `col + 1` in both cases? I have mixed feelings, on the one hand since it's modulo 2 it's the same. But...

cuihtlauac (changes requested):
Looks good to me. The only thing is the big comment, I wish we could save it elsewhere.

src/ocamlorg_frontend/pages/packages_autocomplete_fragment.eml
sabine: Let's just not wrap-around and block on the borders instead.

src/ocamlorg_frontend/pages/packages_autocomplete_fragment.eml
sabine: I kept the one part that you need to know when you read the code as a comment and everything else is a commit message.

PR #979, opened 2023-03-14
Opam repository polling changes

cuihtlauac:
- Remove "Already up-to-date." message emmited by git pull
- Move git clone repo from polling into init
- Start polling by sleeping
- Add poll/pull log
- Add package update log

Note: This will intended to prepare polling of check.ocamllabs.io in
maybe_update.

### DIFF
--- a/src/ocamlorg_package/lib/ocamlorg_package.ml
+++ b/src/ocamlorg_package/lib/ocamlorg_package.ml
@@ -142,16 +142,18 @@ let maybe_update t =
   match t.opam_repository_commit with
   | Some v when v = commit -> Lwt.return ()
   | _ ->
+      Logs.info (fun m -> m "Update server state");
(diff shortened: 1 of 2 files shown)


sabine (approved):
Looks good to me. Instead of cloning on every poll, we only clone when the repository has not already been checked out. Now, polling is a git pull, and instead of waiting before calling `updater` again, `updater` waits when it is called.

cuihtlauac:
> Instead of cloning on every poll, we only clone when the repository has not already been checked out. Now, polling is a git pull, and instead of waiting before calling `updater` again, `updater` waits when it is called.

Actually, cloning was already only done once. But it was triggered from a function intended to do repetitive stuff. The proposed change only makes it obvious that cloning only happens once, at least to me.

PR #983, opened 2023-03-17
stay on current page when using version switch in package docs

### LINKED ISSUE #138: Redirect to the current page when changing the version of a package
If the page exists, we can redirect to it, otherwise, we can redirect to the package overview (or the nearest parent?).

sabine:
Resolves #138 by changing the version select such that it stays on the current package documentation URL when switching to another version of the package.

### DIFF
--- a/src/ocamlorg_web/lib/handler.ml
+++ b/src/ocamlorg_web/lib/handler.ml
@@ -497,7 +497,7 @@ let package_doc t kind req =
       match docs with
       | None ->
           Dream.html
-            (Ocamlorg_frontend.package_documentation_not_found
+            (Ocamlorg_frontend.package_documentation_not_found ~page:(Some path)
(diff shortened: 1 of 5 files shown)


src/ocamlorg_web/lib/handler.ml
cuihtlauac, line 1: That file has a couple of dreadfully long functions, `package_doc` for instance. We'll have to do something about that.

cuihtlauac (changes requested):
I've updated a branch with some changes, mostly related to opt args.

https://github.com/ocaml/ocaml.org/pull/new/cuihtlauac-983

src/ocamlorg_web/lib/handler.ml
sabine: Agreed. I want to remove the fallback first.

PR #986, opened 2023-03-20
Improved Source Code Download on Package Overview

sabine:
Better UX for copying checksums, cleaner layout for the source code download section on package overview.

Ty @clairevanden

|before|after|
|-|-|
|Screenshot 2023-03-20 at 13-04-01 streaming 0 8 0 (latest) · OCaml Package| Screenshot 2023-03-20 at 13-02-52 streaming 0 8 0 (latest) · OCaml Package|


cuihtlauac (approved):
After icons, using named colors should be our next move?

sabine:
Not sure about how far down a color-naming rabbit hole we should go (if at all). Ideally, we have a limited set of colors in the tailwind definitions file, and that will be all colors we use.

PR #989, opened 2023-03-20
don't crash on packages that have avoid-version on all versions

sabine:
I'm pretty sure there's a clever way to do this more efficiently. However, for now I just want packages that have all versions tagged `avoid-version` not crash the search. (Example: search for "base".)


tmattio:
Doesn't that create inconsistent results? Versions with avoid-version of a package are usually not visible, but if the package only contains versions to avoid, then we serve all of them.
I think we should either

- Always include the avoid-version, or, if we don't want that,
- Not serve the package at all if it only has versions to avoid.

sabine:
This only determines which version is considered "latest". (Used to be max_bindings over all versions, before checking for avoid-version was added.)

All package versions are served, including the ones tagged `avoid-version`.

PR #991, opened 2023-03-21
Shorten Ocaml_package public identifiers

cuihtlauac:
- Renaming
- get_package_versions -> get_versions
- get_package -> get
- search_package -> search
- featured_packages -> featured
- state_of_package_list -> state_of_list
- Remove "package" from Ocamlorg_package identifiers


src/ocamlorg_package/lib/ocamlorg_package.ml
```diff
@@ -15,11 +15,11 @@ type state = {
...
   mutable opam_repository_commit : string option;
   mutable packages : Info.t Version.Map.t Name.Map.t;
-  mutable stats : Packages_stats.t option;
-  mutable featured_packages : t list option;
+  mutable stats : Stats.t option;
+  mutable featured : t list option;
 }
 
-let state_of_package_list (pkgs : t list) =
+let state_of_list (pkgs : t list) =
```
sabine: Here I'm not sold on shortening the name, however, I suppose that's because the old name is already quite nondescript. Maybe a better name is `opam_repository_state`?

(At least that's what I got from the comment - that this computes the opam repository state of a list of packages.)

sabine (changes requested):
Let's make it so that we never `open Ocamlorg_package` (there's currently three `open Ocamlorg_package` in the code). The new names are really clear and obvious when we mention the module at all call sites.

sabine:
Sorry, this got caught up in conflicts, but other than that it's merge-ready. :+1:

PR #999, opened 2023-03-21
refactor package urls, routes, and handlers

### LINKED ISSUE #995: Dream errors poping-up in logs
Noticed this in my local build:

```
dream.router ERROR REQ 5 Dream.param: missing path parameter "page"
```

@mtelvers confirms there're also in the live build.

sabine:
This patch
* renames package urls and handlers to better reflect their purpose and be more consistently named
* introduces dedicated url and handler for package files (README, CHANGELOG, LICENSE)
* moves all non-handler package-related functions into a module `PackageHelpers` to prevent confusion with handlers
* extends Ocamlorg_frontend.Package by `homepages` and `source`

Resolves https://github.com/ocaml/ocaml.org/issues/995.

There's more potential for tidying up / clarifying that this patch does not include, e.g.
1) sidebar state of `package_documentation` handler
2) extend voodoo to inspect for README, CHANGELOG, LICENSE and add info about these files to `package.json`
3) make voodoo emit `documentation_status` as part of `package.json`
4) after 2+3) make Ocamlorg_frontend.Package pick up this info from `package.json`

### DIFF
--- a/src/ocamlorg_web/lib/router.ml
+++ b/src/ocamlorg_web/lib/router.ml
@@ -66,26 +66,19 @@ let package_route t =
         (Url.package_overview ~hash:":hash" ":name" ~version:":version")
         ((Handler.package_overview t) Handler.Universe);
       Dream.get
-        (Url.package_file ":name" ~version:":version" ~filepath:"**")
-        ((Handler.package_file t) Handler.Package);
-      Dream.get
-        (Url.package_file ~hash:":hash" ":name" ~version:":version"
-           ~filepath:"**")
-        ((Handler.package_file t) Handler.Package);
-      Dream.get
         (Url.package_documentation ":name" ~version:":version" ~page:"**")
         ((Handler.package_documentation t) Handler.Package);
       Dream.get
         (Url.package_documentation ~hash:":hash" ~page:"**" ":name"
            ~version:":version")
         ((Handler.package_documentation t) Handler.Universe);
-      Dream.get (Url.package_redirect ":name") (Handler.package_redirect t);
       Dream.get
-        (Url.package_redirect ~hash:":hash" ":name")
-        (Handler.package_redirect t);
+        (Url.package_file ":name" ~version:":version" ~filepath:"**")
+        ((Handler.package_file t) Handler.Package);
       Dream.get
-        (Url.package_docs_redirect ":name")
-        (Handler.package_docs_redirect t);
+        (Url.package_file ~hash:":hash" ":name" ~version:":version"
+           ~filepath:"**")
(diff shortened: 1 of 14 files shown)


src/ocamlorg_web/lib/router.ml
```diff
@@ -59,25 +59,33 @@ let package_route t =
...
+        ((Handler.package_file t) Handler.Package);
       Dream.get
-        (Url.package_doc ~hash:":hash" ~page:"**" ":name" ~version:":version")
-        ((Handler.package_doc t) Handler.Universe);
+        (Url.package_documentation ":name" ~version:":version" ~page:"**")
+        ((Handler.package_documentation t) Handler.Package);
+      Dream.get
+        (Url.package_documentation ~hash:":hash" ~page:"**" ":name"
+           ~version:":version")
+        ((Handler.package_documentation t) Handler.Universe);
```
cuihtlauac: Is this the only new route?

cuihtlauac (changes requested):
I've uploaded a branch with some changes. Nothing significant, just helped me review. Feel free to pick and choose. 

https://github.com/ocaml/ocaml.org/tree/cuihtlauac-999

It may have been possible to split this in several PR.  

In `url.ml` I'm not so convinced by the names:
```
package_redirect
package_docs_redirect 
```
To me, redirection is an implementation option which is not relevant to URL string processing and should not be exposed. Maybe `package_overview_without_version` and `package_documentation_without_version` would be better. I'm not sure.

Overall, it goes in the right direction, this is what matters.

src/ocamlorg_web/lib/router.ml
sabine: This is `package_doc` renamed. The new routes are the `package_file` ones.

cuihtlauac: You're right. Sorry for the noise.

sabine:
> In url.ml I'm not so convinced by the names:
> `package_redirect`
> `package_docs_redirect`
> To me, redirection is an implementation option which is not relevant to URL string processing and should not be exposed. Maybe package_overview_without_version and package_documentation_without_version would be better. I'm not sure.

I'm pretty sure I don't want to name it in such a way that someone will be tempted to use that URL in a template. I considered moving it to the `redirection.ml` file, as it's just a redirect. However, there were no "dynamic redirects" happening in that file and that discouraged me. Now that I think about it again, I think I should move the redirects to `redirection.ml`.

cuihtlauac:
> Now that I think about it again, I think I should move the redirects to `redirection.ml`.

I had the same thought. Let's do that and remove them from `Url`

PR #1006, opened 2023-03-22
splice out development dependencies from dependencies

sabine:
* split dependencies in development dependencies (anything depending `with-test`, `with-doc`, etc) and regular dependencies

100% not sure if the logic of checking for a substring `"with-"` here is correct.

|before|after|
|-|-|
|Screenshot 2023-03-22 at 18-06-55 irmin 3 6 1 (latest) · OCaml Package|Screenshot 2023-03-22 at 18-06-59 irmin 3 6 1 (latest) · OCaml Package|

### DIFF
--- a/src/ocamlorg_frontend/pages/package_overview.eml
+++ b/src/ocamlorg_frontend/pages/package_overview.eml
@@ -18,6 +18,7 @@ let render
 ~content
 ~content_title
 ~dependencies
+~dev_dependencies
 ~rev_dependencies
 ~conflicts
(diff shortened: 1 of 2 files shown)


samoht:
That seems like a great idea! But it seems you forgot to move the `dev` dependencies to the new section?

sabine:
I am clueless actually, this is someone else's great idea (probably Ambre or Enguerrand) and they had to educate me about this. So yes, I'll move over the dependencies marked `dev` as well!

PR #1010, opened 2023-03-23
Return 404 on not found

cuihtlauac:
- Add missing 404 response code when
  - Package documentation not found
  - Package file not found
- Define and use helper function html_opt. Streamlined
  processing of optinal data into Dream responses
- Use html_opt as custom binder.
- Formatting, this indents to the left a lot of code

### DIFF
--- a/src/ocamlorg_web/lib/handler.ml
+++ b/src/ocamlorg_web/lib/handler.ml
@@ -1,11 +1,11 @@
 open Ocamlorg
 open Ocamlorg.Import
 
-let http_opt ?(not_found = Ocamlorg_frontend.not_found) opt f =
+let http_or_404 ?(not_found = Ocamlorg_frontend.not_found) opt f =
   Option.fold ~none:(Dream.html ~code:404 (not_found ())) ~some:f opt
 
 (* short-circuiting 404 error operator *)
-let ( let</>? ) opt = http_opt opt
+let ( let</>? ) opt = http_or_404 opt
 let index _req = Dream.html (Ocamlorg_frontend.home ())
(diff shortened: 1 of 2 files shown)


sabine (commented):
I love this, this is like a short-circuiting "if Some, continue and return a HTTP response or if None 404 right here" operator.

I put some suggestions in https://github.com/ocaml/ocaml.org/tree/not-found-going-further. I felt it is a good idea to
1. make the naming a bit clearer, `let</>?` and `http_or_404` (because it's expecting a http response and not some HTML)
2. use the new operator everywhere and be more consistent with how we apply `Dream.html` (it just looks so inconsistent with sometimes the `|>` operator being used but most of the time it's a normal application).
3. not return any HTML from 404 on static file paths (I think I probably introduced that mistake earlier this year).

cuihtlauac:
> I love this, this is like a short-circuiting "if Some, continue and return a HTTP response or if None 404 right here" operator.

Exactly! As usual, you described it even better than me 

> I put some suggestions in https://github.com/ocaml/ocaml.org/tree/not-found-going-further. I felt it is a good idea to
> 
> 1. make the naming a bit clearer, `let</>?` and `http_or_404` (because it's expecting a http response and not some HTML)

Excellent. I'm loving this too.

> 2. use the new operator everywhere and be more consistent with how we apply `Dream.html` (it just looks so inconsistent with sometimes the `|>` operator being used but most of the time it's a normal application).
> 3. not return any HTML from 404 on static file paths (I think I probably introduced that mistake earlier this year).

Fine.

Issue #980, opened 2023-03-15
When package info is regenerating, package info is unavailable


sabine:
Sometimes, package info (which powers package search and all other package-related functionality) is unavailable for a few minutes. We need to figure out how to have a better "uptime" here.

Usually happens at deploy time or shortly after.

sabine:
Resolved by #1011

PR #1011, opened 2023-03-24
Speedup docker run startup

cuihtlauac:
- Create init-cache script
- Reduce startup time

Meant to address: https://github.com/ocaml/ocaml.org/issues/980

### DIFF
--- a/init-cache
+++ b/init-cache
@@ -1,0 +1,11 @@
+#!/bin/sh
+
+CACHE=$HOME/.cache/ocamlorg
+
+rm -f $CACHE/package.state
+touch  $CACHE/package.state
+opam exec -- dune exec src/ocamlorg_web/bin/main.exe &
+PID=$!
+echo $PID
+inotifywait -e close_write $CACHE/package.state
+kill $PID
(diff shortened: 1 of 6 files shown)


init-cache
```diff
@@ -0,0 +1,10 @@
+#!/bin/sh
+
+STATE=$1
+
+rm -f $STATE
+touch  $STATE
+opam exec -- dune exec src/ocamlorg_web/bin/main.exe &
```
tmattio, line 7: Instead of running the entire server, we could create a small wrapper around the function that initialises the state and run this, this would avoid having to kill the process manually.

tmattio (commented):
This is great!! Having a good 20s where packages are unavailable when we push (and we push a lot!) is very frustrating, thanks for doing this 🙂

tmattio:
Btw the cache is supposed to be serialized on a mounted filesystem (outside of the docker container), so that we don't have to re-generate it at launch time. Clearly that's not working correctly, but maybe we should have a look at why it's not doing this if we're trying to improve the startup time?

init-cache
cuihtlauac: Yes, that would be another option. I also considered adding a command line option to `main.exe` that would make it stop once its initialization is complete. Honestly, I had never used `inotifywait` and wanted to play with it.

PR #1015, opened 2023-03-27
Unify package overview and documentation layout

sabine:
Rearranges package overview and documentation such that:
* `package_layout.eml` defines the two sidebars and the content area
* there is a navigation element to switch between Overview (About) and Documentation (Docs)
* there is a placeholder element for the upcoming in-package search (to remind us how important this feature is and to show users that we have a plan where it goes)

Consequences:
* package overview page now has a collapsing sidebar with a button to slide it in on small screens
* package documentation now has a tablet (`md`) layout that shows the sidebar on-screen, instead of collapsed

|page|before|after|
|-|-|-|
|docs xl| Screenshot 2023-03-27 at 15-30-33 irmin 3 6 1 (latest) · OCaml Package|Screenshot 2023-03-27 at 15-30-37 irmin 3 6 1 (latest) · OCaml Package|
| docs lg |Screenshot 2023-03-27 at 15-31-49 irmin 3 6 1 (latest) · OCaml Package|Screenshot 2023-03-27 at 15-31-56 irmin 3 6 1 (latest) · OCaml Package|
| docs md | Screenshot 2023-03-27 at 15-32-38 irmin 3 6 1 (latest) · OCaml Package|Screenshot 2023-03-27 at 15-32-43 irmin 3 6 1 (latest) · OCaml Package|
| docs sm |Screenshot 2023-03-27 at 15-33-34 irmin 3 6 1 (latest) · OCaml Package|Screenshot 2023-03-27 at 15-33-43 irmin 3 6 1 (latest) · OCaml Package|
|-|-|-|
| overview xl |Screenshot 2023-03-27 at 16-00-59 irmin 3 6 1 (latest) · OCaml Package|Screenshot 2023-03-27 at 16-01-05 irmin 3 6 1 (latest) · OCaml Package|
| overview lg with docs fail | Screenshot 2023-03-27 at 16-06-42 ocaml-base-compiler 5 0 0 (latest) · OCaml Package|Screenshot 2023-03-27 at 16-06-45 ocaml-base-compiler 5 0 0 (latest) · OCaml Package|
| overview md |Screenshot 2023-03-27 at 16-07-43 irmin 3 6 1 (latest) · OCaml Package|Screenshot 2023-03-27 at 16-07-46 irmin 3 6 1 (latest) · OCaml Package|
|overview sm| Screenshot 2023-03-27 at 15-46-52 irmin 3 6 1 (latest) · OCaml Package|Screenshot 2023-03-27 at 15-46-57 irmin 3 6 1 (latest) · OCaml Package|

Note: on the `sm` version of package overview, the "install" sidebar slides in on menu button tap, that's why we can see the content area after this patch. ETA: Claire has prepared a dedicated layout for the `sm` screen, so not too much emphasis needs to be put on this one here.

Note: docs failure has a `title` tooltip:
Screenshot from 2023-03-27 15-52-00

### DIFF
--- a/src/ocamlorg_frontend/layouts/package_layout.eml
+++ b/src/ocamlorg_frontend/layouts/package_layout.eml
@@ -7,6 +7,9 @@ let render
 ?hash
 ?page
 ~(package : Package.package)
+~(documentation_status: Package.documentation_status)
+~left_sidebar_html
+~right_sidebar_html
 inner =
 Layout.render
 ?styles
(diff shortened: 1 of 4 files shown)


cuihtlauac:
This is great. Moving back and forth between the API reference and the overview feels very natural. It gives a “that's the way it should be” impression, which is nice. It may make sense to allow round trips between them to return to their starting place, whichever it may be (what's the isomorphism there?)

I preferred the "Modules" label over the "Docs" label. I find the latter too overloaded to be helpful. “Modules” was nice as it had an OCaml flavour, but it may be confusing to newcomers. “API“ or “API reference“ may be better. “About” also sounds a bit too narrow, I preferred “Overview”

I also like having “On this page” and “package xxx” (library index) both on the left. That helped better understand the issue I see in the relationship between these two things. Navigating the `frama-c` package illustrates my point. There we end up with two copies of the same list, back to back, with contents listed in different orders. I believe, from a UX perspective, a single merged list would be better: `List.combine` ;-)

sabine:
> I preferred the "Modules" label over the "Docs" label. I find the latter too overloaded to be helpful. “Modules” was nice as it had OCaml flavour, but it may be confusing to newcomers. “API“ or “API reference“ may be better. “About” also sounds a bit too narrow, I preferred “Overview”

"Docs" is going to split into "Docs" and "Modules", or into "Docs" and "API" when there is both of these. Currently, we only have the index page for "Docs" (which is `index.mld`). For "API", I currently believe the index page will be the full index of the package (waiting on upstream data).

Reason why I tried "About" and "Docs" is that the three of "About", "Docs", "API" do not require collapsing the search bar already on `md` screen. I.e. with names this short, we only need to collapse on `sm` screen. Whereas with the names of "Overview", "Docs", "Modules", we need to collapse the nav and search bar already on `md` screens.

But maybe it is better to already fully collapse on `md` screens, so that the search results will be able to take up the full width of the page. :thinking:

> I also like having “On this page” and “package xxx” (library index) both on the left. That helped better understand the issue I see in the relationship between these two things. Navigating the `frama-c` package illustrates my point. There we end up with two copies of the same list, back to back, with contents listed in different orders. I believe, from a UX perspective, a single merged list would be better: `List.combine` ;-)

The problem currently is that the auto generated `doc/index.html`s are 100% useless. We should not be generating a document with lists of libraries in the first place via `voodoo`. When the package author provides no docs, it would be better to show a page that mentions this and invites the community to reach out to the author or check in the README whether contributions to the documentation are welcome.

ETA: ok, so basically regarding available data:
1. we have the navigation for the "API" or "Modules" tab, but no navigation yet for the "Docs" tab.
2. if the package author provides an `index.mld` file, we have the "home page" of the "Docs" tab. If the package author doesn't, we get this auto-generated list of all libraries and their root modules (which is a bit similar to a root page of an index). The home page of the "API" or "Modules" tab would be the full index of the package, but we're waiting on upstream for that data.

Consequence is: we're in this muddled situation where "Docs" means both "Documentation pages" and "API"/"Modules" until we have the data for both navigations and both home pages.

PR #1017, opened 2023-03-28
Table of contents on package overview pages

sabine:
Patch adds a table of contents to the `xl` screen of package overview pages (reflecting the different sections in the main content area - to give a quick overview of number of dependencies and conflicts) and prepares the addition of table of contents to the package file pages README/CHANGELOG.

|before|after|
|-|-|
|Screenshot 2023-03-28 at 13-54-15 caqti 1 9 0 (latest) · OCaml Package|Screenshot 2023-03-28 at 13-54-19 caqti 1 9 0 (latest) · OCaml Package|

### DIFF
--- a/src/ocamlorg_frontend/components/toc.eml
+++ b/src/ocamlorg_frontend/components/toc.eml
@@ -10,26 +10,26 @@ let render (t : t) =
   <div>
     <% (match t with [] ->  %>
     <% | _ -> %>
-    <div class="font-semibold text-gray-500 text-sm mb-4">ON THIS PAGE</div>
-    <ul class="leading-6 space-y-3 text-sm">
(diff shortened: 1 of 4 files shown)


tmattio:
Interesting! With this change, can we go back to a wide layout?

sabine:
I tested it: going to wide would make the middle content area significantly wider than recommended for prose (reading manuals).

That could be counteracted by making the sidebars wider or increasing the gap between the columns. However, that negatively affects the smaller screens that get the `xl` layout because it reduces their content area width.

My conclusion: the way to make a wide layout work properly on the package pages - without adversely affecting the smaller end of the `xl` screen width - is to introduce a fifth (`xxl`) breakpoint. And that should be a different patch.

cuihtlauac (approved):

PR #1018, opened 2023-03-28
Introduce small fixed footer for pages with sticky sidebars

### LINKED ISSUE #850: Too much scrolling required on the "Learn" page
People need to scroll too much to see the content.

Scrolling down the page does not show the tutorials and guides section.

Scrolling down the navbar on the left it is hard to scroll down to "common error messages", "best practices" and "OCaml platform".

sabine:
This patch solves the following problems:

On all pages with sticky sidebars (learn area + package info area):
1.  when you scroll all the way to the bottom and into the footer, suddenly the sticky sidebars become "unstuck" and move up. That moves some links you may want to use to navigate off screen.
2. when you try to scroll the sidebars on the learn area, you cannot see all content in the sidebar when you scroll the sidebar all the way to the bottom - only after scrolling the overall page a bit until the sidebars become "sticky", you can see that content.

Resolves #850.

How?
* patch introduces a small footer fixed to the bottom of the screen to provide quick access to the most critical links within the Learn and Package areas (replaces the regular footer)
* all three columns (left sidebar, content area, right sidebar) get enough bottom padding in order to scroll them enough to see all the content, even when we are still at the top of the page (where we have a "header" which does not "stick")
* paddings are carefully adjusted to prevent things from moving around when you scroll all the way to the bottom

Alternative possible fixes:
1. don't have a footer at all on pages with sticky sidebars
2. have a footer only in the main content area (which may look stranger than a fixed footer)
3. don't have sticky sidebars
4. have a non-sticky small footer and accept some jankiness in the sticky sidebars when hitting the bottom of the page

|before|after|
|-|-|
|Screenshot 2023-03-28 at 17-22-23 caqti 1 9 0 (latest) · OCaml Package|Screenshot 2023-03-28 at 17-22-18 caqti 1 9 0 (latest) · OCaml Package|
|Screenshot 2023-03-28 at 17-30-11 caqti 1 9 0 (latest) · OCaml Package|Screenshot 2023-03-28 at 17-30-15 caqti 1 9 0 (latest) · OCaml Package|
|Screenshot 2023-03-28 at 17-30-28 caqti 1 9 0 (latest) · OCaml Package|Screenshot 2023-03-28 at 17-30-56 caqti 1 9 0 (latest) · OCaml Package|


cuihtlauac (approved):

cuihtlauac:
I love the idea. I hate having to scroll all the way down to find stuff. I'm tempted to think it should be there on all the pages, although it may be too extreme.

The discrepancy with the content usually found at the bottom of the pages makes me think we may need four more pages, one for each of the four categories usually found at the bottom: “About Us”, “Resources”, “Community” and “Policies”.

It reminds me some documentations have a sticky header. That's the case of https://developer.android.com/reference. We may have to consider this too.

sabine:
Hmmmmm, https://developer.android.com/reference has such a similar layout, but with some serious black magic they managed to make the sticky sidebar on the right shrink when the footer spanning the two right columns comes up! It looks like they're calculating the max height to make the sidebar shrink when the footer appears. :thinking: 

Regarding the footer being fixed to the bottom: I'm actually conflicted about that because, as you mention, it might make sense to have a sticky header to provide access to the search bar and such - having both would take up quite a bit of screen space on smaller screens. Also, I do not know about legal requirements in France: do we need to provide a privacy policy or other legal links on every screen? :thinking:

Having pages for the different footer section is also an interesting idea.

All I know for certain is that the current state is really painful. :see_no_evil: I think this patch needs a bit more work / I need to explore the solution with a shrinking sticky sidebar on the right.

sabine:
I have an exploration of a "shrinking sidebars" layout here: https://github.com/ocaml/ocaml.org/compare/main...sabine:shrinking_sidebars?expand=1. The layout is only implemented on the Learn area (not on the package docs). Both sidebars shrink as the footer comes up - in contrast to the current situation this means that the sidebar content doesn't "float away to the top" when we scroll into the footer, but instead the sidebar content stays in position: it appears as if the footer "overlaps" the bottom of the sidebars.

To achieve this, I had to position the shrinking sidebars using `position:fixed`. However, the consequence of that is that the logic for calculating the positions, margins and padding gets a bit complex.

PR #1021, opened 2023-03-29
Fix missing sidebar on htmx navigation on learn area

sabine:
In order to sometimes have a sidebar on the learn area, and sometimes not, HTMX needs to be used in `outerHTML` swap mode so that the target div gets replaced fully.

All pages need to render the div with id `htmx-right-sidebar` so that there's always a swap target for the right sidebar.

Currently: When you navigate to the learn area and then navigate to a tutorial, the table of contents doesn't appear :see_no_evil:. This patch fixes this problem.

### DIFF
--- a/src/ocamlorg_frontend/layouts/learn_layout.eml
+++ b/src/ocamlorg_frontend/layouts/learn_layout.eml
@@ -43,7 +43,7 @@ let sidebar_icon_link_block
 
 let htmx_attributes href
 =
-  <% if String.get href 0 != '#' then ( %> hx-boost="true" hx-ext="multi-swap" hx-swap="multi:#htmx-head,#htmx-sidebar,#htmx-content,#htmx-right-sidebar" hx-push-url="true" <% ); %>
+  <% if String.get href 0 != '#' then ( %> hx-boost="true" hx-ext="multi-swap" hx-swap="outerHTML multi:#htmx-head,#htmx-sidebar,#htmx-content,#htmx-right-sidebar" hx-push-url="true" <% ); %>
 
 let sidebar_link
 ~title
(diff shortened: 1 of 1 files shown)


src/ocamlorg_frontend/layouts/learn_layout.eml
```diff
@@ -111,13 +111,11 @@ inner_html
...
-            | Some (html) -> %>
-          <div class="hidden xl:flex top-0 sticky h-screen flex-col w-60 overflow-auto lg:pt-6 right-sidebar"
-               id="htmx-right-sidebar">
-            <%s! html %>
-          </div>
-        <% | None -> () ); %>
+        <div class="hidden <%s if Option.is_some right_sidebar_html then "xl:flex top-0 sticky h-screen flex-col w-60 overflow-auto lg:pt-6 right-sidebar" else "" %>"
+            id="htmx-right-sidebar">
+          <%s! match right_sidebar_html with
+          | Some (html) -> html | None -> "" %>
```
cuihtlauac:
```suggestion
          <%s! Option.value ~default:"" (right_sidebar_html) %>
```

src/ocamlorg_frontend/layouts/learn_layout.eml
```diff
@@ -111,13 +111,11 @@ inner_html
...
           <%s! inner_html %>
         </div>
-        <% (match right_sidebar_html with
-            | Some (html) -> %>
-          <div class="hidden xl:flex top-0 sticky h-screen flex-col w-60 overflow-auto lg:pt-6 right-sidebar"
-               id="htmx-right-sidebar">
-            <%s! html %>
-          </div>
-        <% | None -> () ); %>
+        <div class="hidden <%s if Option.is_some right_sidebar_html then "xl:flex top-0 sticky h-screen flex-col w-60 overflow-auto lg:pt-6 right-sidebar" else "" %>"
```
cuihtlauac:
```suggestion
        <div class="hidden <%s Option.fold right_sidebar_html ~none:"" ~some:(Fun.const "xl:flex top-0 sticky h-screen flex-col w-60 overflow-auto lg:pt-6 right-sidebar") %>"
```

I wonder if we should have specialized versions of `Option.fold` and `Option.value` on `string option` and with `""` as default? How often do we return `""` on `string option` throughout `.eml` files?
 
```ocaml
        <div class="hidden <%s Html.maybe right_sidebar_html ~html:(Fun.const "xl:flex top-0 sticky h-screen flex-col w-60 overflow-auto lg:pt-6 right-sidebar") %>"
```

cuihtlauac (changes requested):

src/ocamlorg_frontend/layouts/learn_layout.eml
sabine: I think `Option.fold` and `Option.value` are sufficiently short and have the advantage that they're clearer because newcomers to the code don't have to look up the definition.. It's just I need to actually use them :thinking:

PR #1030, opened 2023-03-30
advance opam-repository to March 30, commit id 9cfac26185d76f082bfdbf6f92f8e1a5aacbc589

sabine:
Advance the opam-repository pin and resolve breaking changes in `hyper` and `mirage-crypto-rng-lwt` (see https://discuss.ocaml.org/t/mirage-crypto-rng-unix-initialize/11505/3).

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@ FROM ocaml/opam:alpine-3.17-ocaml-4.14 as build
 RUN sudo apk update && sudo apk add --update libev-dev openssl-dev gmp-dev oniguruma-dev
 
 # freeze branch was opam-repo HEAD at the time of commit
-RUN cd ~/opam-repository && git checkout -b freeze b457e9f3d6 && opam update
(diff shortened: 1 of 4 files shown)


cuihtlauac:
I suggest we pick commit `c03ceac6872a8ff0ebed2c4fa28b0013fb1984e2` (from march 20) which is the latest in the current `ocaml/opam:alpine-3.17-ocaml-4.14` docker image. That would avoid having to `git fetch`, which will remove some networking from the docker build and save some time with an equivalent result.

Another option, would be to pick the release of ocaml 4.14.1 in the opam-repository log. That would allow the compiler to be the same in all builds.

sabine:
`hyper` was released on March 29. :thinking:

sabine:
Hm, I guess we can keep this around as a draft until we advance opam pin beyond 9cfac26185d76f082bfdbf6f92f8e1a5aacbc589.

cuihtlauac:
That makes sense. It shouldn't take long, @mtelvers told me images are built weekly.

sabine:
Closing in favor of https://github.com/ocaml/ocaml.org/pull/1079

Issue #1022, opened 2023-03-29
Improve https://ocaml.org/docs/ocaml-on-windows


rubyFeedback:
I am trying to compile weidu, which is for games such as Baldur's Gate 2.

Weidu on github says we need ocaml on windows. I am a linux guy mostly so I don't
know windows that well (aside from running stuff, but not really as a developer; I
hate compiling on windows so I am already annoyed that I have to compile something,
but latest release of weidu is from 2019 I think).

Anyway.

I go to:

https://ocaml.org/docs/ocaml-on-windows

And I am trying to find the DOWNLOAD.

Instead I get some random text that babbles about WSL2. Well, do I need WSL2 really?
Can't I just download some binary? Or a .zip file and extract it?

I can't find the thing. Who wrote the homepage? Some 100 years old? It's such a 
confusing page. I think what I would prefer is "Downloads" link that shows the
list of available downloads. I don't want to have to read random content and
the whole page linked above seems to have been autogenerated by Chat-GTP or
something. So please, for the love of life, someone make the webpage better
AND PROVIDE DOWNLOAD LINKS. Even github manages that; you typically
see a "Release" links on the middle right side. That is  useful in general (even
though on weidu it is outdated, but the general functionality is good - why
can't the ocaml web designers learn from github here?)

By the way the page also looks as if it is meant for smartphone users only.
On my widescreen I have only about 50% occupied. I get that not everyone
has a widescreen, no problem - but still, on my system you waste tons of
space. So this is a combination of fail by not-so-great content, combined
with not-so-great webdesign.

I get only very little screen in the middle as the following image shows:

https://i.imgur.com/lpz8VIv.png

After firing the guy who designed the homepage, hopefully someone can fix
this thing and make it more useful. Fancy CSS is great, but first the content
should be CLEAR, and USEFUL, to the POINT. Right now it is not so this should
be improved.

Thank you for reading and hoepfully considering improving the chaotic 
Chat-GTP design of ocaml.org subpage there (I did not check other links
but I guess you guys may have similar issues - the design is just horrible).

rubyFeedback:
Alright, I found the download here:

https://github.com/diskuv/dkml-installer-ocaml/releases

Why is it that the github pages are better than the official ocaml homepage?

Also it is not clear to me why this is the best release. Perhaps in the intro
don't discourage new users saying that this is not the best release? I have
no idea if it is not - I only want to download something that runs, so I can
compile weidu. It's sad panda day, today ...

[...]

PR #1032, opened 2023-03-30
use em breakpoints instead of px breakpoints

sabine:
There are problems with font scaling vs breakpoints when the user sets a custom font size in their OS or browser settings.

Possible solutions are either to
1. use `em` breakpoints in the media queries. Consequence: there may be bugs in ancient browsers that do not properly support relative units in media queries.
2. to set a 16px base font size on the `html` element. Consequence: some browsers will not apply the custom font size setting, resulting in a smaller font size than the user intended. And, since everything scales with the base font size, there will be a lot of white space on the sides for people with wide screens.

This patch changes the breakpoints to use the `em` unit.

Should hopefully fix the problem encountered in https://github.com/ocaml/ocaml.org/issues/1022.

|setting|before|after|
|-|-|-|
|Firefox, scale to 50%, font-size 24|Screenshot 2023-03-30 at 12-39-36 OCaml on Windows · OCaml Tutorials|Screenshot 2023-03-30 at 12-39-41 OCaml on Windows · OCaml Tutorials|
|Firefox, scale to 80%, font-size 24|Screenshot 2023-03-30 at 12-40-51 OCaml on Windows · OCaml Tutorials|Screenshot 2023-03-30 at 12-40-56 OCaml on Windows · OCaml Tutorials|
|Firefox, scale 80%, font-size 16|Screenshot 2023-03-30 at 12-44-00 OCaml on Windows · OCaml Tutorials|Screenshot 2023-03-30 at 12-44-04 OCaml on Windows · OCaml Tutorials|
|Firefox, scale 120%, font-size 16| Screenshot 2023-03-30 at 12-44-16 OCaml on Windows · OCaml Tutorials|Screenshot 2023-03-30 at 12-44-22 OCaml on Windows · OCaml Tutorials|
|Chrome, scale 80%, font-size "Large"| Screenshot from 2023-03-30 13-06-48|Screenshot from 2023-03-30 13-06-54|

### DIFF
--- a/src/ocamlorg_frontend/components/header.eml
+++ b/src/ocamlorg_frontend/components/header.eml
@@ -22,8 +22,8 @@ let render
 =
 <header 
   class="h-20 flex items-center dark:bg-[#171717]"
-  x-data="{ open: false, sidebar: window.innerWidth >= 1024 && true, showOnMobile: false}"
-  @resize.window="sidebar = window.innerWidth >= 1024">
+  x-data="{ open: false, sidebar: window.matchMedia('(min-width: 64em)') && true, showOnMobile: false}"
+  @resize.window="sidebar = window.matchMedia('(min-width: 64em)')">
   <nav class="container-fluid wide header flex justify-between items-center">
     <ul class="space space-x-5 xl:space-x-8 items-center flex text-body-400 font-medium dark:text-white dark:text-opacity-60 dark:font-semibold">
       <li style="width:132px">
--- a/src/ocamlorg_frontend/css/styles.css
+++ b/src/ocamlorg_frontend/css/styles.css
@@ -30,7 +30,7 @@ body {
 }
 
 .container-fluid.wide {
-  max-width: 1440px;
+  @apply max-w-8xl
 }
 
 .bg-pattern {
@@ -152,7 +152,7 @@ body {
   width: 3em;
   margin-left: -2.9em;
   margin-top: 0;
-  font-size: 18px;
+  font-size: 1rem;
 }
 
 .sidebar a.active {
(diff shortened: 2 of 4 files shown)


cuihtlauac:
I've tested on most of the machines I have access to. It looks nice on all of them (including phones and tablets) but my son's. He's using a 2560×1440 27" screen, windows 11 and Edge. He also has the habit of always maximising all windows to full screen. Here is the result.

Capture d’écran (25)

I'm using an even wider screen where I never set windows to full screen. But if I do, I observe the same white areas on the sides. Not sure it qualifies as an issue.

sabine:
Looks fairly reasonable to me on the 2560×1440 screen. The goal is not to eliminate all white space, but keep it in moderation.

What we could do is introduce another breakpoint for really wide screens where we scale up the font size. But that means, you can actually see less of the content as everything expands downwards. I suppose that for people on wide screens, they would rather set a higher browser font setting (which this patch allows them to do, in contrast to the `px`-font-size-on-`html` solution).

2560×1440 has a reasonable height, but there's apparently wide screens that have a really low height (think 2560x1080, for example) where scaling the layout up more would make things a lot worse.

cuihtlauac:
I have a 3440×1440 screen. It doesn't make sense to maximize a browser window on such a screen. It's designed to have three A4-size or letter-size windows side-by-side (without scaling). I don't think it's possible to have a nice layout for large screens with wide aspect ratios.

I don't use zooming and don't set custom fonts (I would if my screen was 5K or high-dpi). To me the white areas is ok. I prefer a slightly denser text, such as in Wikipedia, but I believe this is more a matter of taste.

sabine:
Makes sense, I expect the experience on a high-res screen to be really nice when you do side-by-side browser window - because you get a way above average browser window height in relation to the width.

cuihtlauac:
Both Windows 11 and MacOS have split-screen modes allowing several windows to be displayed side-by-side. To me, the widest aspect ratio we should support is a large tablet in landscape mode (such as the iPad pro). There's no point going further.

PR #1035, opened 2023-03-30
Bringup ocaml to 4.14.1

cuihtlauac:
Bringup existing switch using:
```
opam switch set-invariant ocaml-base-compiler.4.14.1
opam upgrade
```

There's no need to do `opam update`, the pinned opam-repo makes it a noop.

This will reduce the gap with Dockerfile build set-up.


sabine (approved):
Ah, that is nice.

PR #1051, opened 2023-04-01
Use Yoshi to generate OCaml modules from YAML files

tmattio:
This is a first shot at the idea of replacing the "hand-written" parsing and data generation in Ood with a tool that can take a configuration file that contains the data structure of the data as input and take care of the parsing + data generation automatically.
I've created a prototype of it at https://github.com/tmattio/yoshi, and this PR vendors the project and replaces the generation of `Ood.Watch` with it.

In terms, it could allow replacing Ood entirely with a single configuration file, which should be much more maintainable.

If you like the idea, I suggest that we could merge this PR early, even if Yoshi isn't ready to support all the data that we have in `data/`, and gradually extend it to add support for more data types and move the generation of the data modules step by step.


sabine:
Sounds great to me. This would be a useful-to-others tool that could live at https://github.com/ocaml-web and it should be published as an opam package when it is ready for that.

(Tangent: Which reminds me of the static file digests...I suppose that script could be expanded into a static-file-server library that both computes digests and serves compressed static files and caches the files in memory after first load from disk. That would make it a comprehensive drop-in solution for assets - ideally with minimal configuration.)

sabine (approved):
LGTM, we're merging this.

PR #1061, opened 2023-04-04
Clean up all sidebar alpinejs logic

sabine:
* remove useless alpinejs data (`open` on learn layout and package docs, `showOnMobile` on header, learn layout and package docs, and `sidebar` on header)
* use `window.matchMedia().matches`

### DIFF
--- a/src/ocamlorg_frontend/components/header.eml
+++ b/src/ocamlorg_frontend/components/header.eml
@@ -22,8 +22,7 @@ let render
 =
 <header
   class="h-20 flex items-center dark:bg-[#171717]"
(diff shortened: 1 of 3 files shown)


cuihtlauac (approved):

cuihtlauac:
What was the purpose of those `&& true`? Looks odd to me.

sabine:
> What was the purpose of those && true?

Could have potentially been used to force a type cast to bool, but even the old px-based logic didn't need that.

When I introduced em-based checks for the sidebar, though, I introduced a bug because I forgot to add `.matches`.

PR #1063, opened 2023-04-05
fork river

cuihtlauac:
- Bringup peter-evans/create-pull-request to v5
- Pin opam repo in scrape
- Use forked river

### DIFF
--- a/.github/workflows/scrape.yml
+++ b/.github/workflows/scrape.yml
@@ -28,6 +28,8 @@ jobs:
         with:
           ocaml-compiler: ${{ matrix.ocaml-compiler }}
           dune-cache: ${{ matrix.os != 'macos-latest' }}
+          opam-repositories: |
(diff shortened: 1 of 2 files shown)


tmattio:
Thanks @cuihtlauac! Can you open a PR in https://github.com/kayceesrk/river instead to avoid forking? Generally, we should try to reduce the number of forks we have and this one is easy: I have commit right to river.

sabine:
In an ideal world, the timeout in river would be configurable. Though, I think there will still be cases where timeouts will inevitably happen - sometimes the other website may just be offline while we scrape. So our scraper must gracefully recover from that and proceed to scrape the remaining sites.

cuihtlauac:
For sure I will open a pr on river as soon as I have a patch. The fork is an interim thing allowing me to debug issues in ci.

cuihtlauac:
Here is a quickfix PR: https://github.com/kayceesrk/river/pull/9 
A longer-term solution remains required.
I will revert this PR.

PR #1068, opened 2023-04-06
Allow workflows to run locally

cuihtlauac:
For instance, using nektos/act

* Disable sandboxing (we're already in a container)
* Don't wait for user confirmation when installing system packages


sabine (approved):
LGTM

PR #1090, opened 2023-04-12
Upgrade htmx to version 1.9.0

### LINKED ISSUE #877: HTMX doesn't like single quote characters in HTML attributes
HTMX errors when a single quote character is present in an attribute.

Example page: https://ocaml.org/p/ocaml-base-compiler/4.14.1/doc/Genprintval/Make/index.html. Navigating from any other documentation page to this one fails due to the single quote in `install_generic_printer'`.

In order to make navigation happen anyway, when HTMX errors, I opened https://github.com/ocaml/ocaml.org/pull/876.

One way to avoid this issue is to encode `'` as `&#39;` in odoc. Realistically, that would be the safest thing to do, since `'` is a valid string delimiter in HTML.

sabine:
Resolves https://github.com/ocaml/ocaml.org/issues/877.


tmattio:
Is it possible this broke the behaviour for the sidebars @sabine? The navbar gets reset when I change pages on the learn and package area.

sabine:
Yes, looks like it. I think there's an option for scroll preserving behavior that I need to use now. :thinking:

sabine:
No I have accidentally lost the extension that was doing the multi-swap.

PR #1096, opened 2023-04-14
Integrate Sherlodoc search directly via the Sherlodoc API

sabine:
This patch that makes Sherlodoc search available on the package documentation.
Calls the Sherlodoc API directly from the browser.

Current limitation: Sherlodoc has one version of every package indexed and that's the version being searched.

### DIFF
--- a/src/ocamlorg_frontend/layouts/package_layout.eml
+++ b/src/ocamlorg_frontend/layouts/package_layout.eml
@@ -19,7 +19,7 @@ Layout.render
 ?canonical
 ~active_top_nav_item:Header.Packages
 ~footer_html:(Fixed_footer.render ()) @@
-<div class="bg-white">
+<div class="bg-white" x-data="{ q : ''}">
   <div class="pt-6">
     <div class="container-fluid wide">
       <div class="flex justify-between flex-col md:flex-row">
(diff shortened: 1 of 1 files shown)


tmattio:
Wot wot! 🎉 

This is great! I do have one concern though: I'd strongly prefer that we don't rely on external services outside the control of the OCaml.org team. Can we deploy sherlocode on the ocamlorg infrastructure?

sabine:
We will when this has settled as the long-term solution.

Would it be okay to deploy this in a "building-in-public"-fashion so we can get early feedbacks from users?

I can also put it on staging if we want to be a bit more conservative wrt external calls on ocaml.org.

ETA: or I can try and add sherlodoc to ocaml.org's Docker container.

tmattio:
For sherlocode to be used in ocaml.org, it'd have to fall under the OCaml.org projects governance (c.f. https://ocaml.org/policies/governance), so move to an official org and be deployed as part of the OCaml infrastructure.

I don't have concerns deploying it in staging to facilitate user research though!

art-w:
Yes that makes sense!

- The main issue for deployment with the existing dockerfile for ocaml.org is the ~9go database file that will need to be accessible.. Do we have some place where it could be hosted? (It's read-only, do we have access to a docker volume to avoid copying it in the container?)

  In the future, the plan is that voodoo will produce this database but we are not there yet! It would be nice if UI improvements/feedbacks could be worked on in parallel to the integration with the existing documentation pipeline.

- Regarding official org, just let me know where I can move the repo! (or create an official fork if it's easier?) Changing the licence / Renaming sherlodoc is also up for grabs, for example `odoc-search` or whatever you see fit :)

  Note that this might be temporary: I have hope that the sherlodoc codebase will evaporate into `odoc` built-in support for search and that the initial project can be deprecated ^^

src/ocamlorg_frontend/layouts/package_layout.eml
```diff
@@ -60,14 +60,30 @@ Layout.render
...
+            >
+            <input id="in_package_search" type="search" name="q" class="focus:border-gray-800 text-gray-800 border-primary-600 h-10 rounded-l-md appearance-none px-4 flex-grow"
               autocomplete="off"
-              placeholder="Sorry, in-package search is not yet implemented, but this is where it's going to appear :-)">
-            <button disabled aria-label="search" class="h-10 rounded-r-md bg-gray-300 text-gray-800 flex items-center justify-center px-4">
+              placeholder="Search most recently-indexed version of this package..."
+              hx-get="https://doc.sherlocode.com/api"
+              hx-params="*"
+              hx-vals='js:{ "packages" : "<%s Dream.to_percent_encoded package.name %>", "limit" : 10, q: document.getElementById("in_package_search").value.trim()}'
+              hx-trigger="keyup[target.value.trim().length > 1] changed delay:300ms, search"
```
art-w: Can we drop the 300ms delay? It feels painfully slow for a service that responds in 10-20ms :P

sabine: We can sensibly reduce that to 150-200ms, as it can be assumed that the majority of users will be relatively fast typists.

In an ideal world, the debounce function would even take into account typing speed of the user, but that's not a thing that has entered mainstream web dev yet.

However, when we completely remove the debouncing, it means that we trigger a HTTP request for every keystroke. Even if the browser aborts HTTP requests via `AbortController`, it can only stop the HTTP request from hitting the server if it has not already been sent. How many HTTP requests per domain can be in-flight at the same time varies slightly between browsers, but apparently, we must currently assume that this number is 6 or more.

It's a conflict between saving power/bandwith and getting a response slightly faster.

sabine: Hm, I tried 200ms and it was already too fast for me: it would still send useless requests. :thinking:. Reducing to 250.

art-w: From a UX perspective, it doesn't feel right if the user has the time to wonder if the search even works :P It's faaar above the threshold for the system to feel reactive. Are we really saving more than a few bytes from not sending queries asap?

tmattio:
With the work on adding search support in odoc, powered by sherlodoc, I suspect the plan has changed quite significantly.
I suggest we pick this up once search support has landed upstream, closing in the meantime.

PR #1099, opened 2023-04-14
Update HACKING with Git LFS support

cuihtlauac:


HACKING.md
```diff
@@ -4,11 +4,22 @@
...
+
+The project [`Dockerfile`](./Dockerfile) contains up-to-date system configuration instructions, as used to ship into production. It is written for the Alpine Linux distribution but is meant to be adapted to other environments such as Ubuntu, macOS+brew or others. The GitHub workflow file [`.github/workflow/ci.yml`](.github/workflow/ci.yml) also contain useful commands for Ubuntu and macOS+brew. Since ocaml.org is mostly written in OCaml, a properly configured OCaml development environment is also required. Currently, ocaml.org doesn't yet compile using OCaml 5, version 4.14 of the language is used.
+
 The `Makefile` contains many commands that can get you up and running, a typical workflow will be to clone the repository after forking it.
 
 ```
 git clone https://github.com/<username>/ocaml.org.git
 cd ocaml.org
+git lfs install
+git lfs pull
```
sabine, line 22: Should we recommend `git lfs install` before `git clone`, then there's no need to `git lfs pull`? :thinking:

sabine: However, for people who have already checked out without lfs, that line could save them. :see_no_evil:

sabine (approved):

tmattio:
I'm late to the LFS party, but now that Dream has been released with support for OCaml 5, we could upgrade the codebase to OCaml 5 and go back to how we used to compile the playground (removing the need to upload the assets to GitHub).

sabine:
Hm, we could. This way saves some build time, though, because we don't have to build the playground on every build, right?

And in the long run, moving the media associated with ood data to LFS storage doesn't seem so bad either?

tmattio:
Right, I should say I don't think the previous solution is better, I was just pointing to an alternative solution in case it was of interest. Personally I think pushing the assets to the repo is best 🙂

PR #1115, opened 2023-04-19
Outreachy: fix Jyotsna Prakash' project homepage

avsm:
Linking to archive.org since the downstream page is no longer available.


sabine:
Thank you!

PR #1118, opened 2023-04-20
Scraper: in case of River__Http.Timeout, skip the source

sabine:
The scraper would crash when there's a timeout fetching a source's feed.

This patch catches the `River__Http.Timeout` exception to skip the unreachable source during scraping so that at least all the reachable sources will be scraped.

### DIFF
--- a/tool/ood-gen/lib/rss.ml
+++ b/tool/ood-gen/lib/rss.ml
@@ -25,7 +25,13 @@ let pp_meta ppf v =
 
 let feeds () =
   let sources = all_sources () in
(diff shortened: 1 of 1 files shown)


cuihtlauac:
Are we sure this isn't removing all the blog posts from that source when the timeout happens?

sabine:
Luckily, blog posts would never get removed, even when a source timeouts. Only the global feed did lose the entries, before https://github.com/ocaml/ocaml.org/pull/1144 stored the individual feeds.

PR #1126, opened 2023-04-22
WIP: Devcontainer

3Rafal:
It works, but there is redundancy between devcontainer's Dockerfile and project Dockerfile.
Other than that I think it should be good enough as a starting point.

### DIFF
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -1,0 +1,11 @@
+{
+    "name": "OCaml.org",
+    "dockerFile": "Dockerfile",
+    "customizations": {
+        "vscode": {
+            "extensions": ["ocamllabs.ocaml-platform"]
+        }
+    },
+    "containerUser": "opam",
+    "forwardPorts": [8080]
+}
--- a/.devcontainer/devcontainer.json
+++ b/.devcontainer/devcontainer.json
@@ -7,5 +7,6 @@
         }
     },
     "containerUser": "opam",
-    "forwardPorts": [8080]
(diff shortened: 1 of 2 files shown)


3Rafal:
I'm still unsure about project dependencies. Currently it's done via:
```
"onCreateCommand": "opam install --deps-only --with-doc --with-test --yes ."
```
so that when container is initialised, we can just run `make start` and have project being executed.

.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,12 @@
...
+    "name": "OCaml.org",
+    "dockerFile": "Dockerfile",
+    "customizations": {
+        "vscode": {
+            "extensions": ["ocamllabs.ocaml-platform"]
+        }
+    },
+    "containerUser": "opam",
+    "forwardPorts": [8080],
+    "onCreateCommand": "opam install --deps-only --with-doc --with-test --yes ."
```
cuihtlauac, line 11: Could we use `make deps` here? If needed, we could tweak the Makefile and split the `deps` target to share what is common.

cuihtlauac:
Thanks, @3Rafal, this is great. 

Your remark regarding redundancy between the Devcontainer Dockerfile and the project's Dockerfail is accurate. Actually, I was also considering evolving the docker setup to share the system and opam dependencies steps. Currently, these steps are executed for all PR checks, which is wasting time, CPU cycles, bandwidth and overall, energy.

We'd love to hear your opinion on options to achieve both goals simultaneously, which seems possible

cuihtlauac (approved):

PR #1134, opened 2023-04-26
Refine package search results design, include docs link

sabine:
Adjust the design of the search results to be more compact. For packages with more than 5 authors, list only the first five and "et al.".

Might want to compact the tags to a single line (expandable), or omit them (unless the search query matches), in the future. However, in contrast to the authors on some packages, the amount of tags seems to be still bearable so I refrain for now.

||before|after|
|-|-|-|
|xl|Screenshot 2023-04-26 at 14-53-38 OCaml Packages · Search Result|Screenshot 2023-04-26 at 14-53-41 OCaml Packages · Search Result|
|xl|Screenshot 2023-04-26 at 14-54-23 OCaml Packages · Search Result|Screenshot 2023-04-26 at 14-54-27 OCaml Packages · Search Result|
|sm|Screenshot 2023-04-26 at 14-54-52 OCaml Packages · Search Result|Screenshot 2023-04-26 at 14-54-58 OCaml Packages · Search Result|

### DIFF
--- a/src/ocamlorg_frontend/components/icons.eml
+++ b/src/ocamlorg_frontend/components/icons.eml
@@ -93,7 +93,7 @@ let document class_ =
 
 let documentation class_ =
   <svg xmlns="http://www.w3.org/2000/svg" class="<%s class_ %>" fill="none" viewBox="0 0 24 24" stroke-width="2" stroke="currentColor" aria-hidden="true">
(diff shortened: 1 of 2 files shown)


cuihtlauac (approved):
Much better indeed!

PR #1139, opened 2023-04-27
Fix Tutorial Contribution Link

cuihtlauac:
This allows the “contribute” link in tutorial to point to correct
gh blob from staging.ocaml.org

- Make git commit available globally
- Use git commit in tutorial contrib link


sabine (approved):
LGTM :+1:

PR #1144, opened 2023-04-28
Scrape OCaml Planet feeds individually

sabine:
* scrape feeds individually, so that one feed failing to be scraped doesn't make the whole run fail
* individual feeds are stored to `data/planet/*.xml`
* aggregate feed merges all feeds from `data/planet/*.xml` instead of just those that were successfully scraped in this run

This enables us to add many more feeds without being unable to scrape when feeds are temporarily unavailable.

Running the new script locally, I get this diff: https://github.com/ocaml/ocaml.org/commit/fcf2bd590308982b416633e4ec20468fc1939573

### DIFF
--- a/tool/ood-gen/lib/planet.ml
+++ b/tool/ood-gen/lib/planet.ml
@@ -24,17 +24,6 @@ let pp_meta ppf v =
 |}
     (metadata_to_yaml v |> Yaml.to_string |> Result.get_ok)
 
-let feeds () =
-  let sources = all_sources () in
-  sources.sources
-  |> List.filter_map (fun source ->
-         try Some (River.fetch source)
-         with e ->
-           print_endline
-             (Printf.sprintf "failed to scrape %s: %s" source.name
-                (Printexc.to_string e));
-           None)
-
 let validate_entries entries =
   let validate_author (author : Syndic.Atom.author) =
     match author with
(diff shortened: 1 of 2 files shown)


tool/ood-gen/lib/planet.ml
```diff
@@ -51,65 +40,96 @@ let validate_entries entries =
...
+                   date;
+                   preview_image;
+                   description;
+                   featured = None;
+                   authors = Some [ author ];
+                 }
+               in
+               let s = Format.asprintf "%a\n%s\n" pp_meta metadata content in
+               Printf.fprintf oc "%s" s;
+               close_out oc)
```
cuihtlauac, line 102: I think we can play our trick here ;-) (with a couple of simplifications as well):

```suggestion
         let ( let* ) (opt, err) some =
           Option.fold ~none:(Printf.printf "%s/%s: %s" name slug err) ~some opt
         in
         let* _ =
           ( (if Sys.file_exists output_file then None else Some ()),
             "already exist, not scraping again" )
         in
         let oc = open_out output_file in
         let content = River.content post in
         let* url = (River.link post, "skipping, item does not have a url") in
         let date = River.date post |> Option.map Syndic.Date.to_rfc3339 in
         let* date = (date, "skipping, item does not have a date") in
         let url = Some (Uri.to_string url) in
         let preview_image = River.seo_image post in
         let description = River.meta_description post in
         let authors = Some [ River.author post ] in
         let metadata =
           {
             title;
             url;
             date;
             preview_image;
             description;
             featured = None;
             authors;
           }
         in
         let s = Format.asprintf "%a\n%s\n" pp_meta metadata content in
         Printf.fprintf oc "%s" s;
         close_out oc)
```

cuihtlauac (changes requested):
Not sure it realy helps here. I really don't like indentation

tool/ood-gen/lib/planet.ml
```diff
@@ -51,65 +40,96 @@ let validate_entries entries =
...
+                   url = Some url;
+                   date;
+                   preview_image;
+                   description;
+                   featured = None;
+                   authors = Some [ author ];
+                 }
+               in
+               let s = Format.asprintf "%a\n%s\n" pp_meta metadata content in
+               Printf.fprintf oc "%s" s;
```
cuihtlauac:
```suggestion
         let ( let* ) (opt, err) some =
           Option.fold ~none:(Printf.printf "%s/%s: %s" name slug err) ~some opt
         in
         let* _ =
           ( (if Sys.file_exists output_file then None else Some ()),
             "already exist, not scraping again" )
         in
         let oc = open_out output_file in
         let content = River.content post in
         let* url = (River.link post, "skipping, item does not have a url") in
         let date = River.date post |> Option.map Syndic.Date.to_rfc3339 in
         let* date = (date, "skipping, item does not have a date") in
         let url = Some (Uri.to_string url) in
         let preview_image = River.seo_image post in
         let description = River.meta_description post in
         let authors = Some [ River.author post ] in
         let metadata =
           {
             title;
             url;
             date;
             preview_image;
             description;
             featured = None;
             authors;
           }
         in
         let s = Format.asprintf "%a\n%s\n" pp_meta metadata content in
         Printf.fprintf oc "%s" s;
         close_out oc)
```

cuihtlauac (changes requested):
We can play our trick here

tool/ood-gen/lib/planet.ml
sabine: Hm, one thing I greatly value is not to use the same operator name for different things. We're already using `let*` from `lwt` so that name is taken. As I see it, it is quite sad that we can't give custom let binding operators proper names - the set of characters we can use is so limited. :thinking:

I think with these kinds of operators, we have to carefully weigh whether the reduced obviousness of the code is worth it. The short-circuiting 404-binding-operator in the handlers makes sense to me: that one is getting enough use to make teaching the concept worthwhile, and it eliminates quite a bit of nesting for some handlers (which would otherwise have to be split in different functions to stay readable).

I agree that the nesting is bad. In this case, I'm moving the code into a separate function `scrape_post` to maintain a nesting of 3 levels.

cuihtlauac (approved):

PR #1145, opened 2023-05-01
Use Ocamlorg_static.Asset.url for all assets

### LINKED ISSUE #1049: Use `Ocamlorg_static.Asset.url` for all assets
When assets change, browsers may still have them cached. To ensure modified assets reach our users, there's the cache-busting immutable URLs (based on file digest) provided via `Ocamlorg_static.Asset.url`.

The assets of ocaml.org are found in the folder [/asset](https://github.com/ocaml/ocaml.org/tree/main/asset), and they are used throughout the templates in [/src/ocamlorg_frontend/](https://github.com/ocaml/ocaml.org/tree/main/src/ocamlorg_frontend).

Here's an example of how the changes look like: https://github.com/ocaml/ocaml.org/pull/1048/files.

This is a task suitable for an inexperienced contributor.

r17x:
Closes #1049 

Feel free to review this PRs @sabine

### DIFF
--- a/src/ocamlorg_frontend/components/fixed_footer.eml
+++ b/src/ocamlorg_frontend/components/fixed_footer.eml
@@ -4,7 +4,7 @@ let render () =
   <div class="container-fluid wide">
     <ul class="flex flex-col md:flex-row md:flex-wrap md:h-12 gap-x-6">
       <li>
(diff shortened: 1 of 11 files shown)


r17x:
I'm still have a question, could you confirm this file need to use `Ocamlorg_static.Asset.url`:

```bash
src/ocamlorg_frontend/pages/playground.eml
19:    <link rel="icon" type="image/x-icon" href="/favicon.ico">
```

@sabine

sabine:
We can use it for the favicon as well, indeed. Browsers will usually only fetch it once, then cache it for a long time no matter what we do. So either way is fine.

sabine:
@r17x Thank you!

PR #1146, opened 2023-05-02
Rename module Ood into Data

cuihtlauac:
Data is data, doesn't depend on a tool, let's name it accordingly.
This will make it understandable stuff may belong there, even when
it's not generated by ood-gen or yoshi.

P.S. This will make changes required by the latest set of comments in https://github.com/ocaml/ocaml.org/pull/977 look less bizarre


sabine (approved):
I read it as "ood = ocaml.org data". But it took me a while to come to this conclusion.

This makes the naming immediately obvious: now `Data` corresponds directly to the `data/` folder and when you see it in the templates, it's also clearer what this is.

cuihtlauac:
Should we add `open Data` at the beginning of `src/ocamlorg_web/lib/handler.ml`? I count 53 occurrences of `Data.`  in this file.

sabine:
> Should we add open Data at the beginning of src/ocamlorg_web/lib/handler.ml? I count 53 occurrences of Data. in this file.

I would recommend not to open anything unless it's unavoidable. Explicitness is helpful because it makes it obvious where things come from without needing a properly configured IDE. The problem isn't usually verbosity, but rather lack of clarity how things are structured.

cuihtlauac:
> I would recommend not to open anything unless it's unavoidable.

Fine, let's leave it as it is. 

I suggest we merge fast since it is changing so many things

PR #1158, opened 2023-05-04
Simplify the building process for the playground

gpetiot:
Now the process is just:
```
make playground
make start
```
and we don't have to manually switch between 2 opam switches.

### DIFF
--- a/playground/Makefile
+++ b/playground/Makefile
@@ -1,0 +1,11 @@
+.PHONY: create_switch
+create_switch: ## Create switch and pinned opam repo
+	opam switch create 5.0.0 5.0.0 --no-install
+
+.PHONY: deps
+deps: create_switch ## Install development dependencies
+	opam install . --deps-only
+
+.PHONY: build
+build:
+	opam exec --switch=5.0.0 -- dune build --root .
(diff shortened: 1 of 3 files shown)


cuihtlauac (commented):
Looks good to me. Explicitly saying `cd playground` isn't needed may help.

cuihtlauac (approved):

sabine:
This is excellent, thanks for being a great citizen who acts on the bad instructions and makes things better for the next person. :+1:

sabine (approved):

sabine:
```
$ make deps -C playground
make: Entering directory '/home/sabine-tarides/ocaml.org/playground'
opam switch create 5.0.0 5.0.0 --no-install
[ERROR] There already is an installed switch named 5.0.0
make: *** [Makefile:3: create_switch] Error 2
```

I think you can simplify the instructions to only call `make deps -C playground` :thinking:

playground/Makefile
```diff
@@ -1,11 +1,14 @@
+COMPILER=5.0.0
+SWITCH=5.0.0
```
cuihtlauac:
```suggestion
SWITCH=playground
```
I believe the compiler version is irrelevant there.

cuihtlauac (changes requested):

cuihtlauac (approved):

PR #1153, opened 2023-05-03
Documentation: First Hour Improvements including #553

sabine:
This expands on @christinerose's #553 by making more clarifications/rewrites on the First hour tutorial.

Notably:

* add a section about using the OCaml Playground
* mention what `#` means in the code examples

This tutorial is targeted mostly at beginners but sometimes makes references to C.

Patch also contracts the wording in the playground example and renames `OCaml playground` to `OCaml Playground` for consistent capitalisation.

ETA: I am putting this on staging.ocaml.org.

Issue #891, opened 2023-02-06
Add exercises tests


cuihtlauac:
Add a self-testing mechanism to the exercises. A mean for the user to check its implementation without reading the solution.

As a follow-up, those tests should be part of ocaml.org tests to check the validity of the proposed solutions.

sabine:
It would be really nice if every exercise had a test-suite. And then integrate that with the playground somehow:

1. there's a button "open in playground" which leads to a playground that contains the problem statement and some skeleton code to fill in.
2. problem's playground has a special button "check my solution" that runs the problem's test suite

Something like that, right?

cuihtlauac:
Definitely, that would be so cool

amarachigoodness74:
@sabine Can I try working on this?
@SaySayo @cuihtlauac

ghost:
@sabine Can I try working on this?
@SaySayo @cuihtlauac @tmattio

SaySayo:
You can go ahead @amarachigoodness74.

0xrotense:
Hello @SaySayo,  Can i also work on this together with @amarachigoodness74 ?

PR #1159, opened 2023-05-04
WIP: Evaluate exercises in the playground

gpetiot:
Related to #891 and https://github.com/ocaml/ocaml.org/pull/1153#discussion_r1183841042

This branch is rebased on top of #1158 for convenience.

I'm still discovering the codebase (and `dream`) so probably some redesign will be in order. For now, there is an issue with the paths pointing to the exercises, I guess the .md files are nowhere on the server (these paths are the parameters to the playground form). Also the `tests` list in the oracle is incorrectly typed, will fix last.

### DIFF
--- a/data/problems/oracles/001_tail.ml
+++ b/data/problems/oracles/001_tail.ml
@@ -1,0 +1,25 @@
+(* The user input is wrapped in a first [User_input] module.
+   This second module allows to check the type and gives a better error message
+   if the function under-test is missing. *)
+module User_input : sig
+  val last : 'a list -> 'a option
+end =
+  User_input
+;;
+
+let () =
+  let check (type i o) (f : i -> o) i o s n =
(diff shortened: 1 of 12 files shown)


gpetiot:
What works:
- added an `oracle` field to problems, read from a separate file
- added a link from the problems page to go to the playground template, parameterized with the id (number) of the problem
- added a hidden field to pass the oracle and statement to the playground
- router + handler look alright

What doesn't work:
- couldn't fill up the editor textarea with the problem statement (if I put it in the template it appears just before the area, and in the playground component it doesn;t work either
- I had to duplicate the url/handler/router code, I couldn't make it work with the optional parameter

tmattio:
Thanks for starting this work @gpetiot! I understand this lead to working on toplevel support for voodoo and odoc, and that we'll revisit the integration with the Playground once support landed upstream.

We might want to work on the designs with @Clairevanden when we pick this up. Closing the PR the meantime to close open loops/reduce noise.

PR #1166, opened 2023-05-09
Howto stage PR

cuihtlauac:
This proposes a couple of recipes to stage PR in staging as squashed merges: a single commit per PR. This removes the need to use `git rebase`


tmattio:
Would it make sense to put the FOR_MAINTAINERS.md content in HACKING.md or CONTRIBUTING.md? We already have a lot of files at the root of the project so I'm a bit wary of adding more

sabine:
I think it's mostly irrelevant for anyone who doesn't have push rights on the `staging` branch. HACKING or CONTRIBUTING target the public. That's why I moved the guide into the `doc/` folder, so that it is not at the root of the project.

tmattio:
> That's why I moved the guide into the doc/ folder, so that it is not at the root of the project.

Ah, hadn't seen that, thanks!

PR #1168, opened 2023-05-10
Bump ahrefs job as it's still relevant

sabine:
We might need a policy / system for checking whether job listings are still active. There is value in letting companies highlight to people that their listing is still relevant.

For now, I propose that manually bumping every 3-6 months is fine.

### DIFF
--- a/data/jobs.yml
+++ b/data/jobs.yml
@@ -34,6 +34,7 @@ jobs:
     locations:
       - Remote
       - Singapore
+    publication_date: 2023-05-10
     company: Ahrefs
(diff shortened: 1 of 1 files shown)


cuihtlauac (approved):

cuihtlauac:
It seems to be indeed: https://ahrefs.com/jobs/ocaml-developer
That's consistent with @Khady's comment to https://github.com/ocaml/ocaml.org/pull/909

sabine:
Yes, indeed, that's a bump request coming from discuss and I see no reason to deny it.

PR #1171, opened 2023-05-11
Setup the default example of the playground in the handler

gpetiot:
Suggested by @sabine 

It works, but we should probably pass the url of the file .ml instead of its contents to the playground template? It would probably make this feature less flexible (I think) when we inject exercise/tutorial code in the playground (see #1159), what do you think?

### DIFF
--- a/src/ocamlorg_web/lib/handler.ml
+++ b/src/ocamlorg_web/lib/handler.ml
@@ -227,7 +227,11 @@ let carbon_footprint = page Url.carbon_footprint
 let privacy_policy = page Url.privacy_policy
 let governance = page Url.governance
 let code_of_conduct = page Url.code_of_conduct
-let playground _req = Dream.html (Ocamlorg_frontend.playground ())
+
+let playground _req =
+  let default = Data.Code_example.get "default.ml" in
+  let default_code = default.body in
+  Dream.html (Ocamlorg_frontend.playground ~default_code)
 
 let papers req =
   let search_paper pattern t =
--- a/data/code_examples/default.ml
+++ b/data/code_examples/default.ml
@@ -1,7 +1,4 @@
-let print = {|let () = print_endline "Welcome to the OCaml Playground"|}
-
-let adts =
-  {|(* 
+(*
   Welcome to the official OCaml Playground!
 
   You don't need to install anything - just write your code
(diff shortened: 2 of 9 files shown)


sabine:
Injecting via the template is the more flexible option because it does not force a HTTP request. The handler decides where the data comes from. In general, having less HTTP requests to serve a page is desirable.

gpetiot:
Are there things that should be improved before merging this PR? (asking 'cause I want to rebase #1159 on it)

src/ocamlorg_web/lib/handler.ml
```diff
@@ -227,7 +227,11 @@ let carbon_footprint = page Url.carbon_footprint
 let privacy_policy = page Url.privacy_policy
 let governance = page Url.governance
 let code_of_conduct = page Url.code_of_conduct
-let playground _req = Dream.html (Ocamlorg_frontend.playground ())
+
+let playground _req =
+  let default = Data.Code_example.get "default.ml" in
+  let default_code = Omd.escape_html_entities default.body in
```
sabine: I was worried for a moment in the template when I saw `<%s!` (which means "render this unescaped"), but then I see it's being escaped here.

I would recommend moving the HTML escaping to the template and use `Dream.html_escape` instead of `omd`'s facilities. Then, the handler deals with data and the template deals with rendering and escaping.

Generally, this splitting of responsibilities is pretty powerful in terms of the flexibility it gives us regarding rendering - but it is currently being violated in some places because ood-gen is rendering HTML snippets in advance. In the case of the table of contents of tutorials https://github.com/ocaml/ocaml.org/pull/1106), I had to undo the HTML-snippet rendering in `ood-gen` in order to get at the underlying data.

sabine (commented):
Looks overall good to me, just one nitpick to make it easier to see that we are indeed correctly escaping.

src/ocamlorg_web/lib/handler.ml
gpetiot: I was wondering about the `<%s!` vs `<%s` tags.
It sounds better to do the escaping in the template indeed.

sabine: Oki great, this stuff isn't all self-documenting, so I am glad it was useful!

sabine (approved):

PR #1167, opened 2023-05-10
Serve dashboard assets from file system

sabine:
In https://github.com/ocaml/ocaml.org/pull/958, I mistakenly tried to serve the dashboard assets via `Ocamlorg_static.Asset.read`. However, the dashboard's assets are separate.

This patch serves the dashboard assets from the `_build` folder using the default loader from `Dream.static`.

PR #1172, opened 2023-05-12
No crunching (fixed)

gpetiot:
Build on top of #958 (reverts d1c193f) and #1167

The fix is in cf0bece, there was a wrong url in `src/ocamlorg_frontend/pages/package_documentation_not_found.eml`

Edit: Everything works fine locally, waiting for CI

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -32,6 +32,7 @@ COPY --from=build /home/opam/package.state /var/package.state
 COPY --from=build /home/opam/opam-repository /var/opam-repository
 COPY --from=build /home/opam/_build/default/src/ocamlorg_web/bin/main.exe /bin/server
 COPY --from=build /home/opam/_build/default/asset _build/default/asset
+COPY --from=build /home/opam/_build/default/src/dream_dashboard/asset _build/default/src/dream_dashboard/asset
 COPY --from=build /home/opam/_build/default/data/media _build/default/data/media
 
 COPY playground/asset playground/asset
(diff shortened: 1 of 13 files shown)


cuihtlauac:
Thanks, @gpetiot, for finding out what went wrong with #958.

By my tests, it is indeed fixed and faster, around 40%. Not a surprise; compilation is an IO-bound task.

I should have expressed my concerns about this patch during the first PR review. I'm sorry I did not.

How can we measure the impact on operation performance? It looks like a tradeoff; we win at compile time but lose on latency. I don't feel comfortable merging without some data on the latter impact.

Also, I'm cautious about changing a design *only* for compilation performance reasons. To me, compilation speed-up should primarily come from faster hardware, build set-up improvements or code cleaning.

The existing design is fast and simple, qualities which are hard to achieve simultaneously. This is done at the expanse of compilation time and memory consumption. I'm not sure tipping that balance is worth the workload.

sabine:
The next step here would be to serve the assets via a CDN to increase performance for people sitting in remote locations across the world.

ETA: but indeed, it may not yet be the right time for this and it all needs to be done at the same time in order to avoid a regression.

gpetiot:
Should I close this PR?

sabine:
Ok, so the situation with this is:

1. Its true usefulness will only unfold at the point where we either (a) run out of memory for the assets (this is many years in the future) or (b) serve assets from CDN in different geographic locations. As long as neither these happens, the benefit doesn't justify the complexity.
2. Only when we have another caching mechanism for serving the assets will this not cause a performance regression at runtime.

So yes, let's close it for now.

PR #1175, opened 2023-05-13
[RFC] Document the OCaml Platform governance

tmattio:
This commit extends the OCaml.org governance to include the OCaml Platform. We document the lifecycle and requirements for the OCaml Platform tools.

We also re-organise the structure of the governance to move the OCaml.org-specific text into a single section.

Outside of minor text changes to mention the OCaml Platform, the following sections remain untouched:

- Introduction
- Roles and responsibilities
- Processes

We also rename the document from "OCaml.org Governance" to "OCaml Governance" to indicate that it is no longer specific to the OCaml.org project and sub-domains.

Following the goal of the document to describe the _existing_ governance and not what we wish it was, the lifecycle and requirements for the Platform tools are taken from how the OCaml Platform is currently governed. Both the lifecycle and requirements have also been laid down in
[this talk](https://watch.ocaml.org/w/2KbfRNv2oLtkKXkbd5u9F1).

### DIFF
--- a/data/pages/governance.md
+++ b/data/pages/governance.md
@@ -1,107 +1,85 @@
 ---
-title: OCaml.org Governance
-description: The structure of the OCaml.org project, the roles involved and the responsibilities.
-meta_title: Governance of the OCaml.org domain
-meta_description: Read about the structure of the OCaml.org project, the roles involved and the responsibilities.
+title: OCaml Governance
+description: The structure of the OCaml projects, the roles involved and the responsibilities.
+meta_title: Governance of the OCaml projects and domain
+meta_description: Read about the structure of the OCaml projects, the roles involved and the responsibilities.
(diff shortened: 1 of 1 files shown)


avsm (approved):
This looks excellent and reflects my understanding of the consensus and current status quo built up over the past few years. The only change I suggest is to replace the reference to the email infrastructure list with one to https://github.com/ocaml/infrastructure.

tmattio:
Thanks for the review @avsm!

I'm going to merge this now - as you say, this is reflecting today's governance and not proposing any change.

From recent discussions on how the OCaml Platform team can better engage the community in the direction of the Platform, I'm anticipating experimentations on processes and further changes to the governance policy in the coming months/year, but for now, this should bring clarity on the status quo.

PR #1195, opened 2023-05-19
WIP: load packages in the playground

gpetiot:
Inspired by #1159, although it looks a bit more difficult. I laid out the base, but do you have ideas about the next step, should we modify voodoo? Should we wait for a release of odoc for https://github.com/ocaml/odoc/pull/909?

### DIFF
--- a/src/global/url.ml
+++ b/src/global/url.ml
@@ -14,6 +14,7 @@ module Package : sig
     ?hash:string -> ?version:string -> filepath:string -> string -> string
 
   val search_index : ?version:string -> digest:string -> string -> string
+  val playground : ?version:string -> string -> string
 end = struct
   let with_hash = Option.fold ~none:"/p" ~some:(( ^ ) "/u/")
   let with_version = Option.fold ~none:"/latest" ~some:(( ^ ) "/")
@@ -28,6 +29,7 @@ end = struct
 
   let file ?hash ?version ~filepath = base ?hash ?version ("/" ^ filepath)
   let search_index ?version ~digest = base ?version ("/search-index/" ^ digest)
+  let playground ?version name = "/play/package/" ^ name ^ with_version version
 end
 
 let sitemap = "/sitemap.xml"
--- a/src/ocamlorg_frontend/pages/package_overview.eml
+++ b/src/ocamlorg_frontend/pages/package_overview.eml
@@ -50,6 +50,17 @@ in
   </div>
 
   <h2 class="inline-flex items-center text-lg font-medium text-default">
+      <%s! Icons.command_line "mr-4 h-6 w-6 text-orange-600" %>
+      Try in the playground!
+  </h2>
+
+  <div class="mb-6">
+    <a href="<%s! Url.Package.playground ?version package.name %>">
+      Try in the playground
+    </a>
+  </div>
+
+  <h2 class="inline-flex items-center text-lg font-medium text-default">
       <%s! Icons.command_line "mr-4 h-6 w-6 text-primary-600" %>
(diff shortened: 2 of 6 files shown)


sabine:
This may be relevant: https://github.com/ocaml/ocaml.org/pull/908.

gpetiot:
I'm trying to organize my thoughts on this, basically we would need the .cma of every version of every opam package (or only display the link to the playground for the latest version, that's also possible). Even so that's a lot of data.

Let's suppose we have all the .cma's, do we upload them all on ocaml.org as assets so that the toplevel can programmatically load the ones it needs? Again, that's a lot of data.

Just loading the .cma of the package is not enough, we need its dependencies, is it enough to load all that is listed in the opam file? we have to assume the deps are exhaustively listed, which is not necessary as opam will get everything it needs.
Maybe there are even more difficulties that I missed.

Does it sound like something we want? I'm a bit lost but open to suggestions!

sabine:
> I'm trying to organize my thoughts on this, basically we would need the .cma of every version of every opam package (or only display the link to the playground for the latest version, that's also possible). Even so that's a lot of data.

That's true, and we would need to support all versions of a package in order to give a good experience, I think. 🤔 The point would be that people can link to this and existing links shouldn't become obsolete.

> Let's suppose we have all the .cma's, do we upload them all on ocaml.org as assets so that the toplevel can programmatically load the ones it needs? Again, that's a lot of data.

These would go to the same server as the other data generated by voodoo, as they're created during the package documentation build.

> Just loading the .cma of the package is not enough, we need its dependencies, is it enough to load all that is listed in the opam file? we have to assume the deps are exhaustively listed, which is not necessary as opam will get everything it needs. Maybe there are even more difficulties that I missed.

From what I understand, it is reasonable to load the dependencies specified by the package opam file, using the .cma files from the same universe (universe = an opam switch that is used to build multiple packages' documentation).

> Does it sound like something we want? I'm a bit lost but open to suggestions!

I think it would enable an interesting user experience as we could build top levels and interactive examples and tutorials for packages. This would allow people to interactively explore packages with no installation. However, it is unclear to me whether the performance of running this in the browser via jsoo is satisfying and how complex the whole thing will get. I would say it is worth investigating.

jonludlam:
I did a load of work on this last year. There were a few difficulties - I think the most annoying one that should be fixed is that it wasn't possible to load js stubs into the runtime, which rules out quite a lot of packages. I ended up building a different runtime worker for each set of stubs that a library required IIRC.

branches are here:
- https://github.com/ocaml-doc/voodoo/compare/main...jonludlam:voodoo:jsoo
- https://github.com/ocurrent/ocaml-docs-ci/compare/main...jonludlam:docs-ci:jsoo
- https://github.com/ocaml/ocaml.org/compare/main...jonludlam:ocaml.org:js_top_worker

... and possibly some more that I might have forgotten...

gpetiot:
Thanks for the pointers @jonludlam! Did you identify the list of steps to achieve this, what each branch does and what is missing? (or a brief summary would be so helpful, there are a lot of things happening in each branch!). Once I get the whole picture I will try to split them in smaller PRs if possible.

(I guess the voodoo branch can be considered first, in isolation of the rest.)

tmattio:
Similar to #1159 

> Thanks for starting this work @gpetiot! I understand this lead to working on toplevel support for voodoo and odoc, and that we'll revisit the integration with the Playground once support landed upstream.
> 
> We might want to work on the designs with @Clairevanden when we pick this up. Closing the PR the meantime to close open loops/reduce noise.

Issue #1140, opened 2023-04-27
Add blog labels


cuihtlauac:
It would be nice if blog entries would had labels

- [ ] Each blog entry would be labelled with something denoting its origin
- [ ] A filtering mechanism

kiyov09:
@cuihtlauac @sabine Is this issue open to be worked on or still missing some kind of specification?

PR #1197, opened 2023-05-22
Add Signal and Threads as an RSS source

cuihtlauac:


cuihtlauac:
Do not merge. 

Bug not showing up because scraping not part of CI tests. See: https://github.com/ocaml/ocaml.org/issues/1199

Merge fix https://github.com/ocaml/ocaml.org/pull/1200 before this one.

tmattio:
I'm wondering if we should add a separate data source and a section on the site for this. Podcasts and blog posts are quite different.

cuihtlauac:
What about adding tags and filtering? 

https://github.com/ocaml/ocaml.org/issues/1140

We'd have a podcast filter.

sabine (approved):

PR #1207, opened 2023-05-23
Add Caml Weekly News RSS feed

cuihtlauac:
Unrevert and fix https://github.com/ocaml/ocaml.org/pull/1201

Tests results:
- [X] `make scrape`
- [X] `make start`
- [X] Browsing

Notes:
- Items don't have an image. It would be nice to have a default image.


sabine (approved):

tmattio:
Can we do an initial scrape when adding an RSS source? Like this we can see the data that is added and make sure nothing's broken after pulling the data.

data/planet-sources.yml
```diff
@@ -7,3 +7,5 @@ sources:
     url: https://blog.janestreet.com/feed.xml
   - name: hannes
     url: https://hannes.robur.coop/atom
+  - name: camlweeklynews
```
tmattio: We can use the abbreviation:
```suggestion
  - name: cwn
```

tmattio (approved):

tmattio:
When we properly integrate CWN, a more appropriate place would be the news section of the blog, but let's come back to this then.

Thanks @cuihtlauac!

PR #1226, opened 2023-05-27
Add "Are we X yet?" pages

tmattio:
Inspired by Rust's [arewewebyet](https://www.arewewebyet.org/), I thought it useful to have similar pages for OCaml for some parts of the ecosystem.

The goal of these pages is three-fold:
- For newcomers, it offers an overview of the usability of OCaml for certain applications.
- For OCaml users, it can help the discovery of libraries and frameworks to perform certain tasks
- For community members, it can serve as a roadmap to focus our efforts on addressing specific pain points to make OCaml competitive with other languages for specific use cases

**TODO**

The PR isn't ready to be merged yet.

- Needs a description of the state of each category
- The UI is a bit rough and could use some improvements
- When complete, I'd like to get feedback from people who are working on improving OCaml web ecosystem and the community.

image


tmattio:
@bluddy this is similar to OCamlverse in many ways - are you interested in contributing? In terms, I'm thinking we could have pages like these for GUI, Machine Learning, Gaming, etc.

tmattio:
Also, cc @aantron and @anmonteiro as you probably know the state of the OCaml web ecosystem much better than I do.

I'll try to write a description for each category with my understanding of the state of things, but in case you see value in having this page on ocaml.org and have some spare time to help, don't hesitate to share some insights or even rough bullet points that could help me understand where we're at and what's missing.

data/are_we_yet/web.md
```diff
@@ -0,0 +1,252 @@
...
+    - cmarkit
+    - css
+    - csexp
+  - name: External Services
+    status: 🆘
+    description: |
+      n/a
+    packages:
+    - amqp-client
+    - redis
```
anmonteiro: https://github.com/didier-wenzek/ocaml-kafka
```suggestion
    - kafka
    - redis
```

tmattio: Thanks!

data/are_we_yet/web.md
```diff
@@ -0,0 +1,253 @@
...
+    - brotli
+    - zstd
+    - snappy
+    - lzo
+    - zstandard
+  - name: Crypto
+    status: 🟡
+    description: |
+      n/a
+    packages:
```
dinosaure: Please adds `digestif` on this topic

data/are_we_yet/web.md
```diff
@@ -0,0 +1,253 @@
...
+    - dream
+    - eliom
+    - opium
+    - sihl
+  - name: HTTP Clients
+    status: 🔴
+    description: |
+      n/a
+    packages:
```
dinosaure: Please adds `http-lwt-client` on this topic

data/are_we_yet/web.md
```diff
@@ -0,0 +1,253 @@
...
+    - dolog
+    - tracing
+    - catapult
+    - opentelemetry
+    - prometheus
+  - name: Lower Web-Stack
+    status: 🟡
+    description: |
+      n/a
+    packages:
```
dinosaure: Please adds `paf` and also `mimic` which is like `conduit` and remove `colombe` (which should be move into email stuffs)

data/are_we_yet/web.md
```diff
@@ -0,0 +1,253 @@
...
+    - yaml
+    - csv
+    - bencode
+    - toml
+    - xml-light
+    - graphql
+    - omd
+    - cmarkit
+    - css
+    - csexp
```
dinosaure: You probably should add `data-encoding` here

dinosaure (commented):
I added few things into your list

tmattio:
Thanks @dinosaure, I've pushed your suggestions. Do you maybe want to share a few thoughts on the status of the different categories? I still need to write a description for each and having your input would be very helpful

data/is_ocaml_yet/web.md
```diff
@@ -0,0 +1,258 @@
...
+  - name: Web Frameworks
+    status: 🟠
+    description: |
+      n/a
+    packages:
+    - dream
+    - eliom
+    - opium
+    - sihl
+  - name: HTTP Clients
```
edwintorok, line 141: See also https://discuss.ocaml.org/t/simple-modern-http-client-library/11239/14?u=edwin
The short version is that unfortunately if you want something that works with both HTTP/1.x and HTTP/2 and is easy to use, then your best bet is using a binding to a C library (ocurl).
If you're willing to use only HTTP/1.x or only HTTP/2.x or are fine with some of the other limitations of the pure OCaml implementations then there are many choices available (problem there being the many choices, you can spend a large amount of time just testing each until you find a combination that works for you, and inevitably there'll be bugs when all you wanted to do is GET a url or POST a URL).
Hopefully this can be improved, I think all the pieces are there, they are just not the most easy to use or come with a very large dependency cone.

The situation is much better when writing an HTTP *server* though.

aantron: [Hyper](https://github.com/aantron/hyper) will address this.

data/is_ocaml_yet/web.md
```diff
@@ -0,0 +1,258 @@
...
+    - postgresql
+    - aws-rds
+    - sequoia
+    - sqlgg
+    - petrol
+    - mariadb
+  - name: Email
+    status: 🟠
+    description: |
+      n/a
```
dinosaure: I wrote a little description which involves JaneStreet and I'm not sure that they want to communicate on such field:
> Industry and research have done a lot of work on email, where JaneStreet currently uses an OCaml implementation to manage email (https://signalsandthreads.com/building-a-functional-email-server/) and Cambridge University has funded further work on email analysis (https://www.youtube.com/watch?v=kQkRsNEo25k). As a result of this work, a number of projects exist which, among other things, allow emails to be sent and generated according to standards. Tools have been developed to verify and/or sign emails using mechanisms such as DKIM.

So we should probably ask them if they are fine with this little description. Also, you probably should add [`dkim`](https://github.com/mirage/ocaml-dkim) and [`uspf`](https://github.com/mirage/uspf).

data/is_ocaml_yet/web.md
```diff
@@ -0,0 +1,258 @@
...
+    - hacl_x25519
+    - x509
+    - fiat-p256
+    - sha
+    - blake2
+    - blake3
+    - bls12-381
+    - pbkdf
+    - dirsp-proscript
+    - digestif
```
dinosaure: It's worth to add [`spoke`](https://github.com/dinosaure/spoke) as an implementation of the SPAKE2+EE mechanism (or add it into _authentication_).

data/is_ocaml_yet/web.md
dinosaure: Also, if you can remove `facteur`, nobody really use it and I will probably archive it soon.

data/is_ocaml_yet/web.md
```diff
@@ -0,0 +1,258 @@
...
+    status: 🟡
+    description: |
+      n/a
+    packages:
+    - cryptokit
+    - mirage-crypto
+    - nocrypto
+    - pkcs11
+    - tls
+    - hacl_x25519
```
dinosaure: Probably better to delete this specific package when we archived it.

data/is_ocaml_yet/web.md
```diff
@@ -0,0 +1,258 @@
...
+    - decompress
+    - camlzip
+    - lz4
+    - ezgzip
+    - hpack
+    - brotli
+    - zstd
+    - snappy
+    - lzo
+    - zstandard
```
dinosaure: We probably can add [`carton`](https://github.com/mirage/ocaml-git) which does two levels of compressions (`zlib` and binary diff between objects).

data/is_ocaml_yet/gui.md
```diff
@@ -0,0 +1,31 @@
...
+  - name: Platform-Specific
+    status: 🆘
+    description: |
+      n/a
+    packages: []
+  - name: Web Technologies
+    status: 🆘
+    description: |
+      n/a
+    packages: []
```
xvw: I suggest to add : 
- [Bonsai](https://github.com/janestreet/bonsai)
- [OCaml-vdom](https://github.com/LexiFi/ocaml-vdom)
- [Fmlib_browser](https://github.com/hbr/fmlib)

data/is_ocaml_yet/web.md
```diff
@@ -0,0 +1,258 @@
+---
+id: web
```
xvw, line 2: As the question of the state of the art of static blog generators in OCaml is raised quite often on the forum, I think adding a section dedicated to them is a good thing.

data/is_ocaml_yet/web.md
```diff
@@ -0,0 +1,258 @@
...
+    - ethernet
+    - paf
+    - mimic
+  - name: Serializers
+    status: 🟡
+    description: |
+      n/a
+    packages:
+    - atd
+    - yojson
```
edwintorok: `jsonm` should be added too, but note that it doesn't produce standard conforming output if you serialize Float NaN/infinity values (the resulting json cannot be parsed), it require patching JSON parsers to accept the modified grammar.
`yojson` does at least produce standard conforming output, and the application can then handle NaN specifically (either as a null or string).However AFAICT yojson doesn't validate that its input is UTF-8 conforming.

My current stance is to use jsonm where validating input is important and use yojson where standard conforming output is important.

https://github.com/dbuenzli/jsonm/issues/12#issuecomment-1572025576

dbuenzli: I think this a wrong reading of `jsonm`'s api which [clearly states](https://erratique.ch/software/jsonm/doc/Jsonm/index.html#encoding_limitations) what's on the client to generate valid JSON. 

If fail to see how that's different from `yojson` which does not produces standard conforming output as it will happily output any byte string as an ill-coded json string; and I'm not even talking about the fact that it doesn't check the character encoding on input.

edwintorok: I'd like to have at least one json library that takes and checks spec conformant JSON on input and products spec conformant JSON as output. Whether that is by patching 'jsonm' to implement the extra checks that 'yojson' does or patch 'yojson' to do the extra checks that 'jsonm' does. I'm currently attempting to fix it by doing the former :)

edwintorok: I'd like to have at least one json library that takes and checks spec conformant JSON on input and products spec conformant JSON as output. Whether that is by patching 'jsonm' to implement the extra checks that 'yojson' does or patch 'yojson' to do the extra checks that 'jsonm' does. I'm currently attempting to fix it by doing the former :)

dbuenzli: So what is your spec conformant library going to do when it sees a NaN ? raise `Invalid_argument` ? That's no output which is not conformant. 

What you need to do on seing such a value depends on the JSON format you are outputing, this decision is left to the client in `jsonm` because that's the only reasonable course of action for a general purpose JSON library.

edwintorok: There are various choices: raising an exception (so we never output something non-conformant), or serializing as `null`, or as a string. Either of these choices would allow the resulting JSON to be parsed and converted back at the application level to float if needed, it'd at least be conformant with the JSON grammar (it is what some other JSON libraries are doing, so there would even be some limited interoperability).
You are right that the choice should be left to the caller, however outputting invalid JSON is not a good default choice, because not even jsonm will be able to read back the output it just produced.

Jsonm already transforms strings on output (which is much more expensive in terms of performance), so why not floating point values?

edwintorok: There are various choices: raising an exception (so we never output something non-conformant), or serializing as `null`, or as a string. Either of these choices would allow the resulting JSON to be parsed and converted back at the application level to float if needed, it'd at least be conformant with the JSON grammar (it is what some other JSON libraries are doing, so there would even be some limited interoperability).
You are right that the choice should be left to the caller, however outputting invalid JSON is not a good default choice, because not even jsonm will be able to read back the output it just produced.
OCaml has optional arguments so the library could implement one or more choices, and applications that really want  different semantic from the default can chose.

Jsonm already transforms strings on output (which is much more expensive in terms of performance), so why not floating point values?

dbuenzli: > Jsonm already transforms strings on output (which is much more expensive in terms of performance), so why not floating point values?

I don't think you are reading the documentation I'm pointing you to.

data/is_ocaml_yet/web.md
tmattio: Thanks for the suggestions and the description! Would you be able to give some insights on what's missing to get the ideal state for email support in OCaml? Could be gaps in supported protocols, missing libraries, or robustness/maturity. From a developer perspective, what's the story like at the moment if they want to send emails from their web applications?

data/is_ocaml_yet/web.md
tmattio: I'm adding `jsonm` to the list, thanks for the suggestion, it's a great library! I suggest continuing the discussion about the feature request on https://github.com/dbuenzli/jsonm/issues/22 though 😉

data/is_ocaml_yet/web.md
tmattio: Done

data/is_ocaml_yet/web.md
tmattio: Added in auth

data/is_ocaml_yet/web.md
tmattio: Great idea! I added a section with a few packages

data/is_ocaml_yet/gui.md
tmattio: I added these in the "Browser" category in "Is OCaml Web Yet?". I think the intention of Web Technologies for GUI should be more along the line of JavaScript engines, or HTML/CSS rendering engines

data/is_ocaml_yet/web.md
tmattio: Done

data/is_ocaml_yet/multicore.md
```diff
@@ -0,0 +1,26 @@
+---
+id: multicore
+question: Is OCaml multicore yet?
+answer: OCaml 5 is out, now we need to build an ecosystem!
+categories:
+  - name: Async IO
+    status: 🆘
+    description: |
+      n/a
+    packages: []
```
c-cube:
```suggestion
    packages:
    - [eio](https://github.com/ocaml-multicore/eio)
```

data/is_ocaml_yet/web.md
```diff
@@ -0,0 +1,275 @@
...
+    - avro
+    - capnp
+    - grpc
+    - protobuf
+    - rpc
+    - jsonrpc
+    - thrift
+    - bson
+    - cbor
+    - mqtt
```
c-cube:
```suggestion
    - mqtt
    - msgpck
```

data/is_ocaml_yet/web.md
```diff
@@ -0,0 +1,275 @@
...
+    - awa
+    - uri
+    - dns
+    - tcpip
+    - charrua
+    - mirage-nat
+    - arp
+    - ethernet
+    - paf
+    - mimic
```
c-cube:
```suggestion
    - mimic
    - tiny_httpd
```

data/is_ocaml_yet/web.md
tmattio: Thanks!

data/is_ocaml_yet/multicore.md
tmattio: Thanks, I added a few packages to the Multicore page, including eio

data/is_ocaml_yet/web.md
```diff
@@ -0,0 +1,277 @@
...
+    - sihl
+  - name: HTTP Clients
+    status: 🔴
+    description: |
+      n/a
+    packages:
+    - cohttp
+    - hyper
+    - ocurl
+    - http-lwt-client
```
c-cube:
```suggestion
    - http-lwt-client
    - ezcurl
```

listed in the discussion @edwintorok pointed to, and after all I think it _is_ more convenient than raw ocurl :)

data/is_ocaml_yet/web.md
c-cube: Should this also list ocurl? Curl supports SMTP(S): https://everything.curl.dev/usingcurl/smtp (edit: to be fair, I don't even know if ocurl exposes that)

data/is_ocaml_yet/web.md
```diff
@@ -0,0 +1,278 @@
...
+    - zmq
+  - name: Syndication/RSS
+    status: 🟡
+    description: |
+      n/a
+    packages:
+    - syndic
+    - river
+    - rss
+  - name: Templating
```
yawaramin, line 227: It should probably be mentioned here that templating is not the only option for generating HTML in OCaml and that libraries like Tyxml also exist to embed an HTML DSL directly into the language, which may be preferable for several reasons.

dbuenzli:
> It should probably be mentioned here that templating is not the only option for generating HTML in OCaml and that libraries like Tyxml also exist to embed an HTML DSL directly into the language, which may be preferable for several reasons.

That's the problem you get with this kind of page, it makes you think you need things you may well never end up using and misses things you may end up doing differently. If you ask me, "we" have been web for an extremely long time it just depends on what you need regardless of what other people try to make you think you absolutely need because of groupthink or whatever marketing and hype they have been enslaved by.

There's always a lot of enthusiasm when this kind of page gets created and then it slowly rots away (especially because people who really work with the language couldn't care less). That's not very good from a marketing viewpoint since it won't give an accurate view of the eco-system.

These pages are already starting to feel like a very broken index of the packages published on `opam`. I think it would be much better to invest time in taking time to formalize and maybe gently police `opam` tags a bit more – have a guide page for package publishers on which tags to use for which kind of service and perhaps opam-repository maintainer having an eye for them when packages are submitted.

You can then have "are we X yet" page mentioning and linking on the tags so that people can dynamically check for themselves the options they have at hand for that task.

data/is_ocaml_yet/web.md
dinosaure: > Would you be able to give some insights on what's missing to get the ideal state for email support in OCaml?

Currently, when it comes to sending emails, we can consider that all the tools are available to a developer. What remains experimental is the implementation of an SMTP service. We can nevertheless extend the existing system with new authentication mechanisms but, as far as I know, those currently available are sufficient.

> Could be gaps in supported protocols, missing libraries, or robustness/maturity.

Robustness & maturity are there. JaneStreet has been using its service for a long time and it's been 3-4 years since I've needed to make a major update to the email sending function - even though this part seems to be used by several web frameworks (like `dream` or `sihl`).

> From a developer perspective, what's the story like at the moment if they want to send emails from their web applications?

I think, still for sending emails, that the main thing is there (generation, protocols, authentication mechanism, verification and signature).

dbuenzli: > From a developer perspective, what's the story like at the moment if they want to send emails from their web applications?

It depends. 

If you are just sending emails to a reasonably sized group of users or administrators then you can just shell out to `sendmail` like basically any `php` application does (which on debian can be a proxy for something much simpler like [`msmtp`](https://wiki.debian.org/msmtp) if you install `msmtp-mta` that you connect to the mail server of your email provider). So it can be as simple [as that](https://github.com/dbuenzli/hyperbib/blob/a1def764f7f58b29bd732dacaa167e73497f6175/src/email.ml) (modulo the escaping :-).

If you need massive mailing then you likely better off with a mailing service like [mailgun] and then it's simply REST interaction with a webservice. You don't even need to bother with all the standards. The twist here, like with all webservices, is that if you use a popular language the REST interface is likely already implemented for you as a library so it's a bit more work. But are we ready ? Definitively.

[mailgun]: https://www.mailgun.com/

data/is_ocaml_yet/web.md
hannesm: FWIW, your table brought me to revising reverse dependencies in the mirage-crypto / ocaml-tls / dns-client environment, and I suspect the "OCaml Dependencies" and "System Dependencies" may need an update (also, I feel that a brief number is only about quantity about packages - but you can always bundle everything and get to a single dependency -- so maybe it's not the healthiest metrics (since people then strive for few large dependencies)). In any case, thanks again for the table which motivated me to disentangle some code.

sabine:
> formalize and maybe gently police opam tags a bit more – have a guide page for package publishers on which tags to use for which kind of service and perhaps opam-repository maintainer having an eye for them when packages are submitted.

I agree: we can (and should) use the categories that emerge here as basis for a set of opam tags "blessed" by the UI. With "blessed" I mean that it will be possible to browse packages by the blessed category tags and that search filters on the package search page prominently offer them.

So I think the major outcome here should be to establish a good taxonomy. :slightly_smiling_face:

Daniel, even if these pages rot away in 2 years time (which is not certain by any means!)... Not everything we build has to be for eternity. It has to serve a purpose now - and I think it does because
(1) it helps people see that there's so much going on in the OCaml community and ecosystem these days,
(2) it brings us together to look at how to make the ecosystem better, and
(3) it makes us build out a category tree for the opam packages to the point that we can rally to make one big effort of going over all the existing packages and assigning them to the proper categories.

When the search UI and package browsing UI for this blessed set of categories is live, motivation to go over all opam files will be at a high point. While at the moment, there just isn't a lot of momentum for this effort of tagging opam packages in a consistent fashion.

I know you don't really care so much about (1) and (2), but let's use the momentum to do (3).

When this effort here has settled, we take the categories that emerged and we look up the best existing matching tags for those. We build out the UI that effectively blesses these tags so it becomes very very obvious how this benefits the community and then it's time to rally people to help out with tagging the opam packages.

dbuenzli:
I think you are a bit overthinking all that. There's no need to bless anything nor any need to establish a taxonomy upfront. Tags will evolve anyways over time as technologies and hype cycles come and go. 

So you don't want a system that hardcodes anything, it should be easy to add any tag and get a widget for the tag – how many packages have the tag (if any), how many *stable* packages have the tag (self-assessed by the package owner via a `stable` or `unstable` tag), links to the lists of packages and maybe if that's your thing a fancy indicator based on the stats to indicate if we are blabla yet for the tech corresponding to the tag.

Just let these things organize themselves: watch what people are already using, filter and maybe clarify or improve them (e.g. unifying diverging terminology). As soon as people see there's a page where their stuff automatically advertises itself if they use the right tags, you absolutely won't have to rally people to help with tagging the packages, package owners will naturally use them (and at moment the community is sufficiently small to leave out the spam problem out of the equation).

chshersh:
I would love to see a separate category (or two) on CLI and TUI ❤️ 

For CLI, I can think of:

* [cmdliner](https://github.com/dbuenzli/cmdliner)
* [clap](https://github.com/rbardou/clap)
* [ANSITerminal](https://github.com/Chris00/ANSITerminal)

And for TUI:

* [notty](https://github.com/pqwy/notty)

data/is_ocaml_yet/gui.md
sabine: For these packages, a better category than "Browser" would be "Virtual DOM" as this is a keyword known in the frontend web dev space.

dbuenzli: Not really, not every browser ui framework works with a virtual DOM. In fact improvements in DOM rendering by browsers likely renders the whole idea obsolete.

data/is_ocaml_yet/web.md
```diff
@@ -0,0 +1,278 @@
...
+  - name: Static Site Generation
+    status: 🟡
+    description: |
+      n/a
+    packages:
+    - finch
+    - stone
+    - camyll
+    - dream-serve
+    - soupault
```
dinosaure: [`yocaml`](https://github.com/xhtmlboi/yocaml) is another solution used by many too.

xvw: Yes but ... not available on OPAM... :(

sabine: @xhtmlboi would you maybe be open to having yocaml published as an opam package?

xhtmlboi: I'm really sorry but I've never really used OPAM so I don't have no idea how to publish a package on OPAM.

sabine: @xhtmlboi would you be okay with it if someone else stepped up to help? I'm sure we'll find someone willing to put some effort to make it happen.

xhtmlboi: In order to be `<= 5.0.x`, I'm sure that I'll probably make some changes (for example, rename the module `Effect`). 
When it is done, I'll try to publish the repo on OPAM and if this branch is merged, I'll make a pull request. Thanks for the suggestion. Since @dinosaure is a power user of maintening OPAM packages, I guess that I can ask him for help if I struggle too much.

sabine: I added a link to the GitHub repo. :)

christinerose (approved):

sabine:
I'm merging this now.

If you (the reader) feel you want to contribute, it's very welcome to open a PR to add more stable packages or write something to entice people to add to the ecosystem, or add more pages for other topics you care about.

PR #1228, opened 2023-05-28
Fix Sys.getenv_opt type signature in tutorial

joshleeb:
This change is a minor fix to the _Data Types and Matching_ tutorial, to fix the type signature of `Sys.getenv_opt` which should be `string -> string option` and not `string -> string opt`.


christinerose (approved):

cuihtlauac (approved):

cuihtlauac:
Thanks, @joshleeb

Issue #1185, opened 2023-05-16
Make ocamlorg_package a standalone library


cuihtlauac:
Off-line discussions with @sabine and @tmattio have shown the library `ocamlorg_package` may have enough value on its own to be usable in other projects. In order to facilitate such a scenario, it should be decoupled from ocamlorg dependencies
 - [ ] `ocamlorg_data`
 - [ ] `global`

Initialization of the cache, which is currently done in the `init-cache` shell script should be turned into a internal mechanism (dupe of https://github.com/ocaml/ocaml.org/issues/1045)

 - [ ] Rewrite `init-cache` in OCaml, inside `ocamlorg_package`

It should become possible to copy the `ocamlorg_package` folder and use it as-is or turn it into a standalone package.

PR #1229, opened 2023-05-30
Make ocamlorg_package stand free

cuihtlauac:
- Remove dependency on ocamlorg_data
- Remove dependency on global

This addresses the two firsts issues of https://github.com/ocaml/ocaml.org/issues/1185


src/ocamlorg_package/import/string.ml
tmattio, line 1: We can have a module `import` in `ocamlorg_package` directly so we don't create a library for a single function

src/ocamlorg_package/lib/packages_stats.ml
```diff
@@ -1,4 +1,9 @@
-open Ocamlorg.Import
+module List = struct
```
tmattio: This can go in the import module mentioned above

tmattio:
This is great, thanks for doing this @cuihtlauac! I left a couple of comments

src/ocamlorg_web/lib/handler.ml
```diff
@@ -435,18 +440,27 @@ let packages state _req =
...
 let packages_search t req =
   match Dream.query req "q" with
   | Some search ->
+      let search' =
+        match String.split_on_char ':' search with
+        | [ "author"; default ] ->
+            let author = Data.Opam_user.github_name default in
+            "author:" ^ Option.value ~default author
+        | _ -> search
+      in
```
tmattio: Not sure that's correct. The search function in Ocamlorg_package will try to match the search with the author field so if you search by name and the author field contains the name, this introduces a regression.

Do we support multiple search? (e.g. `author:"Thibaut Mattio" author:"tmattio"`). If so, we can add the GitHub handle instead of replacing the search query.

src/ocamlorg_package/lib/ocamlorg_package.ml
```diff
@@ -453,12 +455,15 @@ end = struct
...
+  let to_request ~nick_resolve str =
     let str = String.lowercase_ascii str in
     let to_constraint = function
       | [ _; s ] -> Any s
       | [ _; "tag:"; s ] -> Tag s
-      | [ _; "author:"; s ] -> Author s
+      | [ _; "author:"; s ] ->
+          Author
+            Option.(
+              value ~default:s (map String.lowercase_ascii (nick_resolve s)))
```
tmattio: Isn't it doing the same thing that was done before?

c.f. previous comment:
> Not sure that's correct. The search function in Ocamlorg_package will try to match the search with the author field so if you search by name and the author field contains the name, this introduces a regression.
> 
> Do we support multiple search? (e.g. author:"Thibaut Mattio" author:"tmattio"). If so, we can add the GitHub handle instead of replacing the search query.

cuihtlauac: You can't just turn `author:tmattio` into 
`author:tmattio author:"thibaut mattio"` (if this is what you meant). We search for the conjunction of the filters and there's no disjunction operator.

tmattio: With the code above, my understanding is that if I type `author:"thibaut mattio"`, it will search for `author:tmattio`

cuihtlauac: I've changed the names to reflect they work the other way around.

tmattio: But then the problem remains, if I search for `author:tmattio`, it will search for `author:"thibaut mattio"`? Eitherway I am searching for something and I'm potentially not getting the exact matches in the result because the query that's executed is not the one I used. Again, it's fine to alter it if we're returning a superset (if we support querying the same field multiple time we can add to the query `author:"thibaut mattio" author:tmattio`), but it doesn't seem correct to return a different set.

cuihtlauac: The name/author/nick/email matching logic is now:
- Defined in `Data.Opam_user`
- Used in `Ocamlorg_package.search`
- Same as current online logic

Results with this PR should be the same as they used to be.

`Ocamlorg_package.Search` could be turned into a functor to avoid passing the `is_author_match` parameter from call to call.

tmattio: Cool! I suggest we move the matching logic out of Data though, we might change that to use fuzzy matching or something else in the future and how we search for authors shouldn't have any impact on data

tmattio:
@cuihtlauac This appears to break package authors:

image

cuihtlauac:
You're right; I've forgotten to `git add` the package state cache invalidation.

sabine (approved):

PR #1231, opened 2023-05-30
Implemented form input and form select tailwind components

SaySayo:
This pull request
- Implements form input and form select tailwind components and applied the components across pages where the form tags are used. 
- Cleans up the unused styles in forms, selects and button tags.

cc @sabine


sabine (approved):
I took the opportunity to create Dream components for `search_input` and `select` and tidy up the htmx indicator styles to not refer to explicit ids.

I think this is good to merge for now.

tmattio (approved):
Largely in favour of this PR and the way it's implemented! From a UI perspective though, I would find it better to have different designs for the navigation search bar and the ones throughout the site. Can come in another PR, let's merge this one. Thanks @SaySayo!

sabine:
Indeed, as the design system matures, this will be refined.

Issue #1178, opened 2023-05-15
Add an Error Template so that 500 Errors do not render an empty page


cuihtlauac:
Issue https://github.com/ocaml/ocaml.org/issues/1170 has shown when an internal error happens, nothing valuable is shown to the user. A page saying something like: “Sorry, an internal error took place. Not your fault. If you can, please let us know by filing a GitHub issue.”

Dream offers proper support for rendering an Error Template (see https://aantron.github.io/dream/#errors), but we are currently not using this yet.

The task here is to create an error template and to set up an error handler following [Dream's documentation](https://aantron.github.io/dream/#errors).

PR #1244, opened 2023-05-31
feat: handle internal error with template

r17x:
Hello @sabine & @cuihtlauac 

I've tried to use error_handler with template for resolved this issue #1178

Feel free for review or maybe you'll have any suggestion about my implementation

Thanks

image

### DIFF
--- a/src/ocamlorg_frontend/pages/internal_error.eml
+++ b/src/ocamlorg_frontend/pages/internal_error.eml
@@ -7,8 +7,8 @@ Layout.render
     <p class="text-4xl font-extrabold text-orange-600 sm:text-5xl">500</p>
     <div class="sm:ml-6">
       <div class="sm:border-l sm:border-gray-200 sm:pl-6">
-        <h1 class="text-4xl font-extrabold text-gray-900 tracking-tight sm:text-5xl">Page not found</h1>
-        <p class="mt-1 text-base text-gray-500">Sorry, an internal error took place. Not your fault. If you can, please let us know by filing a <a href="https://github.com/ocaml/ocaml.org/issues/new" rel="noopener noreferrer" target="_blank">GitHub issue</a>./p>
+        <h1 class="text-4xl font-extrabold text-gray-900 tracking-tight sm:text-5xl">Internal Error</h1>
+        <p class="mt-1 text-base text-gray-500">Sorry, an internal error took place. Not your fault. If you can, please let us know by filing a <a href="https://github.com/ocaml/ocaml.org/issues/new" rel="noopener noreferrer" target="_blank" class="font-extrabold text-orange-600 ">GitHub issue</a>.</p>
       </div>
       <div class="mt-10 flex space-x-3 sm:border-l sm:border-transparent sm:pl-6">
         <a href="/"
--- a/src/ocamlorg_frontend/pages/internal_error.eml
+++ b/src/ocamlorg_frontend/pages/internal_error.eml
@@ -1,0 +1,25 @@
+let render () =
+Layout.render
+~title:"Internal Error"
+~description:"This page does not exist." @@
+<div class="bg-white px-4 py-16 sm:px-6 sm:py-24 md:grid md:place-items-center lg:px-8">
+  <div class="sm:flex max-w-max mx-auto">
+    <p class="text-4xl font-extrabold text-orange-600 sm:text-5xl">500</p>
+    <div class="sm:ml-6">
+      <div class="sm:border-l sm:border-gray-200 sm:pl-6">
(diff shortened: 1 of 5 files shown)


tmattio:
Thanks @r17x! Can you add a screenshot of the new page in the description of the PR? It make it easier to review the UI alongside the code without having to run the PR locally 🙂

r17x:
@tmattio  I have been updated the PRs description. (included UI)

Cheers 🚀

src/ocamlorg_frontend/pages/internal_error.eml
```diff
@@ -0,0 +1,25 @@
+let render () =
+Layout.render
+~title:"Internal Error"
+~description:"This page does not exist." @@
+<div class="bg-white px-4 py-16 sm:px-6 sm:py-24 md:grid md:place-items-center lg:px-8">
+  <div class="sm:flex max-w-max mx-auto">
+    <p class="text-4xl font-extrabold text-orange-600 sm:text-5xl">500</p>
```
tmattio, line 7: The HTTP status isn't necessarily 500, Dream calls the error handler for errors other than 500. In fact, if I understand correctly, with this change, unfound pages (404) will now use this template?

r17x: I have been tested and you're corrected 400 will be use this templates. 

I think we need to changes, do you have any suggestions what I need to do?

tmattio: The error template receives `suggested_response ` as a last argument (https://aantron.github.io/dream/#val-error_template). You can use this to get the status (`Dream.status`) and with the status you can also get an explanation of the error using `Dream.status_to_reason`. You can pass these to the templates to display the status and explanation.

sabine: Currently, the 404 pages are being served from `static.ml`'s `loader` function which returns a 404 response if an asset is not found under the given URL. At the bottom of `router.ml`, we capture `/**` to try and load an asset - after all other routes fail to match.

Looking at the Dream documentation for `error_template`, it says

> In case of a 4xx or 5xx response from the application, that response itself is passed to the template.

Oh, that is interesting. If I understand that correctly, in case of a 4xx response the error handler of Dream will catch that and place the response into `suggested_response`. And indeed, when the handler returns a 4xx or 5xx response, we can be sure that this is using its own dedicated template. In that case, we should just return this suggested response coming from the handler.
It looks like we can see what the cause of the error is by inspecting the `error` parameter: https://aantron.github.io/dream/#type-error.

sabine: @tmattio passing all errors through the new template will not work because this would override the package documentation 404 template.

However, it still makes sense to have the "ordinary" 404 page returned by `http_or_404` use this new generic error template. (This doesn't have to be done in this PR, but could be a later refactor.)

(Instead of returning the regular `not_found` response, `http_or_404` could throw an exception that `error_template`'s error handler can consume to render the 404 error into the generic error response. :thinking:)

sabine:
Hey @r17x, cool that you're taking on this challenge! I hadn't looked into how Dream does this yet, so it's exciting to see how it plays out to make it work with the `Ocamlorg_frontend.package_documentation_not_found` template. :+1:

src/ocamlorg_frontend/pages/internal_error.eml
tmattio: > @tmattio passing all errors through the template will not work because this would override the package documentation 404 template.

Make sense. Tbh I don't think that would a problem to have a single error template but if we want to keep a separate template for the documentation what I suggested doesn't work indeed.

Perhaps a more general approach would be to check if suggested_response contains a body already and only use the template if it doesn't.

sabine: I haven't looked into how exactly Dream behaves wrt `suggested_response`. If we can check for presence of a body that makes sense and would be simpler than the exception approach.

sabine: To elaborate on why we need the package documentation 404 error page:

1. When you use the version switcher on a documentation page, you currently cannot see whether that page exists in a specific version. When you reach a 404 page this way, it's disorienting.
2. For packages that did not have their documentation build successfully: sometimes we have an error to display for the failed package documentation build, and sometimes we only have a 404 page.
3. Links into the package documentation of the latest version may, on a new release of the package, turn into 404 pages - it's much more useful to offer the user the search bar of the package documentation and the link to the package overview in this situation. A generic 404 page would be disappointing in this context.

tmattio:
Closing for now, but feel free to re-open if you continue to work on this @r17x.

PR #1248, opened 2023-06-01
Remove stray mention of domainslib from playground

jonludlam:


gpetiot (approved):

tmattio:
Thanks for the report @gpetiot and for the quick fix @jonludlam!

PR #1252, opened 2023-06-02
Search packages by module name

### LINKED ISSUE #1238: Search packages by module name
It does not seem to be possible to search packages by provided module names, for instance:

https://ocaml.org/packages/search?q=Mirage_crypto_rng_lwt

does not return anything

gpetiot:
Closes #1238 

`~name` is passed to `make` because calling `name opam` to try to get the package name would raise an exception if the `name:` field is missing from the opam file (which is often the case).

### DIFF
--- a/src/ocamlorg_package/lib/info.ml
+++ b/src/ocamlorg_package/lib/info.ml
@@ -2,7 +2,7 @@ type url = { uri : string; checksum : string list }
 
 (* This is used to invalidate the package state cache if the type [Info.t]
    changes. *)
-let version = "2"
+let version = "3"
 
 type t = {
   synopsis : string;
@@ -19,6 +19,7 @@ type t = {
   url : url option;
   publication : float;
   flags : OpamTypes.package_flag list;
+  module_name : string;
 }
 
 let relop_to_string = OpamPrinter.FullPos.relop_kind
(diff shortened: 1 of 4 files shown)


cuihtlauac:
Nice! Thanks

In order to have the complete list of modules inside a package, the data we get from docs-data.ocaml.or needs to processed more and stored in package state to be made available to search.

The good news is, the data is there, see for instance:

https://docs-data.ocaml.org/live/p/irmin/3.7.1/package.json

sabine:
1. Isn't searching for module names across packages something that Sherlodoc will do?
2. Indeed, indexing just the one module (if existing) listed in the opam file is unsatisfying. However, taking into account all module names in a package during search can produce conflicts with package names - so at the very least, the module name matches would need to score lower.
3. We can't see on the package search results page that a package was listed among the results because it had a matching module name.

I am not sure this feature is going to be worth the effort needed to make it useful and well-behaved. Instead, we could work towards integrating Sherlodoc across-package search to cater to the use case of searching for module names within packages.

If there are some good arguments against my objections or evidence that users really need to be able to search for module names in the package search, I'm happy to hear about it and change my opinion.

sabine:
I'm closing this because it looks like there's going to be a different, better place for this module-name-search feature.

PR #1256, opened 2023-06-05
Bring-up Dream to 1.0.0~alpha5

cuihtlauac:


dune-project
```diff
@@ -67,7 +67,7 @@
   (dream
    (>= 1.0.0~alpha3))
```
tmcgilchrist:
```suggestion
   (>= 1.0.0~alpha5))
```
Think about pinning to this version instead, to force the solver to find alpha5 or fail. Also this package constraint is repeated in the dune-project file.

sabine (approved):

PR #1263, opened 2023-06-05
Set package polling frequency to 1h

cuihtlauac:


ocamlorg.opam
```diff
@@ -25,7 +25,7 @@ depends: [
   "bos"
   "crunch"
   "mirage-kv-mem"
-  "dream" {>= "1.0.0~alpha5"}
+  "dream" {>= "1.0.0~alpha3"}
```
sabine: Why is dream listed twice - and with two different versions? :thinking:

cuihtlauac: It shouldn't be duplicated. I was in main yesterday. I will rebase this

sabine: Thanks!

sabine (approved):

PR #1283, opened 2023-06-07
Comment out credits on sequence tutorial

sabine:
Reason for doing this: Over time, these acknowledgements sections will grow very large and create a lot of visual noise in the tutorial right at the point where we want people to see the contribute and discuss links.

However, I think we can keep the credits/acknowledgements in a comment.


cuihtlauac:
What do you mean by a comment?

PR #1282, opened 2023-06-07
Enable mdx processing of data

cuihtlauac:
DO NOT MERGE - This nearly doubles the build time

- Let dune inspect data folder
- Make mdx skip purposedly hanging examples

Closes https://github.com/ocaml/ocaml.org/issues/1280

@gpetiot: can you help me sort out the `$MDX part` thing? I believe it is the last remaining error.

PR #1300, opened 2023-06-08
Fix mdx blocks

gpetiot:
We need to upgrade the version of the mdx extension to use the mdx sranza:
```dune
(mdx
 (files *.md))
```

Also, the `part` label is used to select parts of an .ml file to inject them in md blocks, and that's not what is done here, so I removed them. See https://github.com/realworldocaml/mdx/tree/main/test/bin/mdx-test/expect/parts-begin-end for an example.


christinerose (approved):

cuihtlauac:
Manually merged into #1282

PR #1318, opened 2023-06-14
Extensive editing of error-handling

tuohy:
- Major rewrite of section on using `Fun.protect`
- the errors thrown by `Option.get` `Result.(get_ok, get_error)` are `Invalid_argument` in 5.0, not `Failure` as was previously documented in "Convertions (sic) Between Errors"
- changed the example of unsafe use of marshal to use `ocaml -stdin` instead of compiling a file
- Many spelling and grammar fixes
- Some moving words around to stay within 80 characters/line
- I added myself as a contributor, there aren't guidelines for what level of contribution would be required for this kind of attribution

### DIFF
--- a/data/tutorials/lg_08_error_handling.md
+++ b/data/tutorials/lg_08_error_handling.md
@@ -11,10 +11,12 @@ date: 2021-05-27T21:07:30-00:00
 
 In OCaml, errors can be handled in several ways. This document presents most of
 the available means. However, handling errors using the effect handlers
-introduced in OCaml 5 isn't addressed yet. This topic is also addressed in th [_Error Handling_](https://dev.realworldocaml.org/error-handling.html) chapter
-of the _Real World OCaml_ book by Yaron Minsky and Anil Madhavapeddy (see references).
+introduced in OCaml 5 isn't addressed yet. This topic is also addressed in the
+[_Error Handling_](https://dev.realworldocaml.org/error-handling.html) chapter
+of the _Real World OCaml_ book by Yaron Minsky and Anil Madhavapeddy (see
+references).
 
-## Error as Special Values
+## Errors as Special Values
 
 Don't do that.
 
@@ -30,7 +32,7 @@ instance, in Unix systems, here what is contained in `man 2 read`:
 >
 > On error, -1 is returned, and `errno` is set to indicate the error.
 
-Great software was written using this style. However, since correct are errors
+Great software was written using this style. However, since correct and error
 values can't be distinguished, nothing but the programmer's discipline ensures
 errors aren't ignored. This has been the cause of many bugs, some with dire
 consequences. This is not the proper way to deal with errors in OCaml.
@@ -42,8 +44,8 @@ There are three major ways to make it impossible to ignore errors in OCaml:
 
 Use them. Do not encode errors inside data.
 
-Exceptions provide a mean to deal with errors at the control flow level while
-`option` and **`result`** provide means to turn errors into dedicated data.
+Exceptions provide a means to deal with errors at the control flow level, while
+`option` and **`result`** make errors distinct from normal return values.
 
 The rest of this document presents and compares approaches towards error
 handling.
@@ -61,7 +63,7 @@ However, exceptions have the great merit of being compiled into efficient
 machine code. When implementing trial and error approaches likely to back-track
 often, exceptions can be used to achieve good performance.
 
-Exceptions belong to the type `exn` which is an [extensible sum
+Exceptions belong to the type `exn`, which is an [extensible sum
 type](/manual/extensiblevariants.html).
 
 ```ocaml
@@ -128,7 +130,7 @@ val failwith : string -> 'a
 
 When implementing a software component which exposes functions raising
 exceptions, a design decision must be made:
-* Use the prexisting exceptions
+* Use the preexisting exceptions
 * Raise custom exceptions
 
 Both can make sense, and there isn't a general rule. If the standard library
@@ -139,39 +141,36 @@ desirable for errors that must be handled at the client level.
 
 ### Using `Fun.protect`
 
-The `Fun` module of the standard library contains the following defintion:
+Because handling exceptions interrupts normal control flow, using them can
+complicate some tasks requiring strictly ordered and coupled processes. For
+these scenarios, the `Fun` module of the standard library provides:
 ```ocaml
 val protect : finally:(unit -> unit) -> (unit -> 'a) -> 'a
 ```
 This function is meant to be used when something _always_ needs to be done
-_after_ a computation is complete, either it succeded or failed. Any computation
-can be postponed by wrapping it into a dummy function with only `()` as
- parameter. Here, the computation triggered by passing `x` to `f` (including its
- side effects) will not take place:
+_after_ a computation is complete, whether it succeeds or fails. The unlabeled
+function argument is called first, then the function passed as the 
+[labeled argument](/docs/labels) `finally` is called. `protect` then returns the
+same value as the unlabeled argument, or raises the same exception that that
+function raised.
+
+Note that functions passed to `protect` take only `()` as a parameter. This
+does not limit the cases where `protect` can be used; any computation `f x` can
+be wrapped in a function `fun () -> f x`. As always, the body of the function
+won't be evaluated until the function is called.
+
+The `finally` function is only expected to perform some side-effect, and should
+not raise any exception itself. If `finally` does throw some exception `e`,
+`protect` will raise `Finally_raised e`, wrapped to make it clear that the
+exception is not coming from the protected function.
+
+Let's demonstrate with a function that tries reading the first `n` lines of a
+text file (like the Unix command `head`). If the file has fewer than `n` lines,
+the function must throw `End_of_file`. In any case, the file descriptor must be
+closed afterwards. Here is a possible implementation using `Fun.protect`:
+
 ```ocaml
-let work () = f x
-```
-
-It would, on execution of `work ()`. This is what `protect` does, and such a
-`work` function is the kind of parameter `protect` expects. The `finally`
-function is called by `protect`, after the completion of `work ()`, in two
-possible way, depending on its outcome
-1. If it successed, the result produced is forwarded
-2. If it failed, exception raised is forwarded
-
-The `finally` function is only expected to perform some side-effect. In summary,
-`protect` performs two computations in order: `work` and then `finally`, and
-forwards the outcome of `work` either result or exception.
-
-The `finally` function shall not raise any exception. If it does, it will be
-raised again, but wrapped into `Finally_raised`.
-
-Here is an example of how it can be used. Let's imagine a function reading the
-`n` first line of a text file is needed (like the `head` Unix command). If the
-file hasn't enough lines, the function must throw `End_of_file`. Here is a
-possible implementation using `Fun.protect` to make sure the file is always closed:
-```ocaml
-# let rec head_channel chan =
+# let head_channel chan =
   let rec loop acc n = match input_line chan with
     | line when n > 0 -> loop (line :: acc) (n - 1)
     | _ -> List.rev acc in
@@ -180,26 +179,32 @@ val head_channel : in_channel -> int -> string list = <fun>
 # let head_file filename n =
   let ic = open_in filename in
   let finally () = close_in ic in
-  Fun.protect ~finally (fun () -> head_channel ic n);;
+  let work () = head_channel ic n in
+  Fun.protect ~finally work;;
 val head_file : string -> int -> string list = <fun>
 ```
+When `head_file` is called, it opens a file descriptor, defines `finally` and
+`work`, then `Fun.protect ~finally work` performs two computations in order:
+`work ()` and then `finally ()`, and has the same result as `work ()`, either
+returning a value or raising an exception. Either way, the file descriptor is
+closed after use.
 
 ### Asynchronous Exceptions
 
 Some exceptions don't arise because something attempted by the program failed,
-but rather because an external factor is impeding its execution. Those exeptions
-are called asynchronous. This is the case, for instance, of the following ones:
+but rather because an external factor is impeding its execution. Those 
+exceptions are called asynchronous. These include:
 
 * `Out_of_memory`
 * `Stack_overflow`
 * `Sys.Break`
 
 The latter is thrown when the user interrupts an interactive execution. Because
-they are losely or unrelated with the program logic, it mostly doesn't make
-sense to track the place where an asynchronous exceptions was thrown, could be
-anywhere. Defining if an application needs to catch those exceptions and how it
-should be done is beyond the scope of this tutorial. Interrested readers may
-refer to Guillaume Munch-Maccagnoni [A Guide to recover from
+they are loosely or not at all related with the program logic, it mostly doesn't
+make sense to track the place where an asynchronous exceptions was thrown, as it
+could be anywhere. Deciding if an application needs to catch those exceptions
+and how it should be done is beyond the scope of this tutorial. Interested
+readers may refer to Guillaume Munch-Maccagnoni [A Guide to recover from
 interrupts](https://guillaume.munch.name/software/ocaml/memprof-limits/recovering.html).
 
 ### Documentation
@@ -234,10 +239,10 @@ let () = Printexc.record_backtrace true
 
 ### Printing
 
-To print an exception, the module `Printexc` comes in handy. For instance, it
-allows the definition of a function, such as `notify_user : (unit -> 'a) -> 'a`
-to call a function and, if it fails, prints the exception on `stderr`. If
-stack traces are enabled, this function will also display it.
+To print an exception, the module `Printexc` comes in handy. For instance, the
+function `notify_user : (unit -> 'a) -> 'a` below calls a function, and if it
+fails, prints the exception on `stderr`. If stack traces are enabled, this
+function will also display it.
 
 ```ocaml
 let notify_user f =
@@ -275,7 +280,7 @@ errors at runtime.
 
 The compiler and runtime makes a best effort for raising meaningful exceptions.
 However, some error conditions may remain undetected, which can result in a
-segmentation fault. This is the specially the case for `Out_of_memory`, which
+segmentation fault. This is especially the case for `Out_of_memory`, which
 is not reliable. It used to be the case for `Stack_overflow`:
 
 > But catching stack overflows is tricky, both in Unix-like systems and under
@@ -288,14 +293,12 @@ is not reliable. It used to be the case for `Stack_overflow`:
 This has improved since. Only linked C code should be able to trigger an
 undetected stack overflow.
 
-### Genuinely Unsafe Functions
+### Inherently Unsafe Functions
 
-Some OCaml functions are genuinely unsafe. Use them with care; not like this:
+Some OCaml functions are inherently unsafe. Use them with care, not like this:
 
 ```shell
-> echo "fst Marshal.(from_string (to_string 0 []) 0)" > boom.ml
-> ocamlc boom.ml
-> ./a.out
+> echo "fst Marshal.(from_string (to_string 0 []) 0)" | ocaml -stdin
 Segmentation fault (core dumped)
 ```
 
@@ -303,9 +306,9 @@ Segmentation fault (core dumped)
 
 When a crash isn't coming from:
 * A limitation of the native code compiler
-* An genuinely unsafe function such as found in modules `Marshal` and `Obj`
+* An inherently unsafe function such as are found in modules `Marshal` and `Obj`
 
-It may be a language bug. It happens. Here is what to do when this is suspected:
+it may be a language bug. It happens. Here is what to do when this is suspected:
 
 1. Make sure the crash affects both compilers: bytecode and native
 1. Write a self-contained and minimal proof-of-concept code which does nothing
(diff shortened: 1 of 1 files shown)


cuihtlauac:
This is excellent, @tuohy; thank you very much. Everything is so much better the way you've turned it.

The only tiny place where I have something to say is this edit:

> Properly handling errors always makes the code harder to read

> Improperly handling errors always makes the code harder to read.

What I wanted to say was something like: code ignoring errors altogether is likely to be easiest to read, but wrong. Therefore, it is a matter of choosing the right/best/least evil/appropriate way to reduce readability to catch what needs to be caught. Obviously, I should I've written a whole sentence here.

cuihtlauac (approved):
This is good. Let's merge this

data/tutorials/lg_08_error_handling.md
```diff
@@ -11,10 +11,12 @@ date: 2021-05-27T21:07:30-00:00
 
 In OCaml, errors can be handled in several ways. This document presents most of
 the available means. However, handling errors using the effect handlers
-introduced in OCaml 5 isn't addressed yet. This topic is also addressed in th [_Error Handling_](https://dev.realworldocaml.org/error-handling.html) chapter
-of the _Real World OCaml_ book by Yaron Minsky and Anil Madhavapeddy (see references).
+introduced in OCaml 5 isn't addressed yet. This topic is also addressed in the
```
christinerose:
```suggestion
introduced in OCaml 5 hasn't been addressed yet. This topic is also addressed in the
```

data/tutorials/lg_08_error_handling.md
```diff
@@ -30,7 +32,7 @@ instance, in Unix systems, here what is contained in `man 2 read`:
 >
 > On error, -1 is returned, and `errno` is set to indicate the error.
 
-Great software was written using this style. However, since correct are errors
+Great software was written using this style. However, since correct and error
 values can't be distinguished, nothing but the programmer's discipline ensures
```
christinerose: "However, since correct and error values can't be distinguished"
This "correct" part I find a little confusing. Do you mean the style is correct but at the same time error values can't be distinguished?

If so, perhaps: "However, since this style is correct and error values can't be distinguished..."

data/tutorials/lg_08_error_handling.md
```diff
@@ -139,39 +141,36 @@ desirable for errors that must be handled at the client level.
...
-can be postponed by wrapping it into a dummy function with only `()` as
- parameter. Here, the computation triggered by passing `x` to `f` (including its
- side effects) will not take place:
-```ocaml
-let work () = f x
-```
+_after_ a computation is complete, whether it succeeds or fails. The unlabeled
+function argument is called first, then the function passed as the 
+[labeled argument](/docs/labels) `finally` is called. `protect` then returns the
+same value as the unlabeled argument, or raises the same exception that that
```
christinerose:
```suggestion
same value as the unlabeled argument or raises the same exception that
```
While using "that" twice in this sentence is also correct, I think it's not necessarily needed here. Author's choice, but the comma should be removed.

data/tutorials/lg_08_error_handling.md
```diff
@@ -139,39 +141,36 @@ desirable for errors that must be handled at the client level.
...
+function raised.
 
-It would, on execution of `work ()`. This is what `protect` does, and such a
-`work` function is the kind of parameter `protect` expects. The `finally`
-function is called by `protect`, after the completion of `work ()`, in two
-possible way, depending on its outcome
-1. If it successed, the result produced is forwarded
-2. If it failed, exception raised is forwarded
+Note that functions passed to `protect` take only `()` as a parameter. This
+does not limit the cases where `protect` can be used; any computation `f x` can
```
christinerose:
```suggestion
does not limit the cases where `protect` can be used. Any computation `f x` can
```

data/tutorials/lg_08_error_handling.md
```diff
@@ -139,39 +141,36 @@ desirable for errors that must be handled at the client level.
...
+be wrapped in a function `fun () -> f x`. As always, the body of the function
+won't be evaluated until the function is called.
 
-The `finally` function is only expected to perform some side-effect. In summary,
-`protect` performs two computations in order: `work` and then `finally`, and
-forwards the outcome of `work` either result or exception.
+The `finally` function is only expected to perform some side-effect, and should
+not raise any exception itself. If `finally` does throw some exception `e`,
```
christinerose:
```suggestion
not raise any exception itself. If `finally` does throw some exception `e`,
```
Should this be "throw the exception `e`" or "an exception `e`"? 
"Some" is usually used with a plural noun.

data/tutorials/lg_08_error_handling.md
```diff
@@ -180,26 +179,32 @@ val head_channel : in_channel -> int -> string list = <fun>
...
-they are losely or unrelated with the program logic, it mostly doesn't make
-sense to track the place where an asynchronous exceptions was thrown, could be
-anywhere. Defining if an application needs to catch those exceptions and how it
-should be done is beyond the scope of this tutorial. Interrested readers may
-refer to Guillaume Munch-Maccagnoni [A Guide to recover from
+they are loosely or not at all related with the program logic, it mostly doesn't
+make sense to track the place where an asynchronous exceptions was thrown, as it
+could be anywhere. Deciding if an application needs to catch those exceptions
+and how it should be done is beyond the scope of this tutorial. Interested
+readers may refer to Guillaume Munch-Maccagnoni [A Guide to recover from
```
christinerose:
```suggestion
readers may refer to Guillaume Munch-Maccagnoni's [A Guide to recover from
```

data/tutorials/lg_08_error_handling.md
```diff
@@ -473,18 +475,19 @@ Here is a questionable but straightforward implementation using exceptions:
...
+`String.index` fails, meaning no dot character was found, we may return the
+whole fully qualified domain name (FQDN) as a fallback, but only if it isn't
+the empty string.
 
-Note that `String.sub` may throw `Invalid_argument`. Fortunately, this can't
-happen. In the worst case, the `@` character is the last one, then `fqdn_pos` is
-off range by one but `fqdn_len` is null, and that combination of parameters
-doesn't count as an invalid substring.
+Note that generally `String.sub` may throw `Invalid_argument`. Fortunately, this
+can't happen when calculating `fqdn`: in the worst case, the `@` character is
```
christinerose:
```suggestion
can't happen when calculating `fqdn`. In the worst case, the `@` character is
```

data/tutorials/lg_08_error_handling.md
```diff
@@ -574,20 +577,20 @@ robustness. A couple of observations:
...
 
-When used to handle errors with catch statements, it requires some time to get
-used the latter style. The key idea is avoiding or deferring from directly
-looking into option values. Instead, pass them along using _ad-hoc_ pipes (such
-as `map` and `bind`). Erik Meijer call that style: “following the happy path”.
-Visually, it also resembles the “early return“ pattern often found in C.
+When used to handling errors with catch statements, it requires some time to get
+used the latter style. The key idea is avoiding or deferring looking directly
+into option values. Instead, pass them along using ad hoc pipes (such as `map`
+and `bind`). Erik Meijer calls that style “following the happy path”. Visually,
```
christinerose:
```suggestion
and `bind`). Erik Meijer calls that style “following the happy path.” Visually,
```

data/tutorials/lg_08_error_handling.md
```diff
@@ -574,20 +577,20 @@ robustness. A couple of observations:
...
+it also resembles the “early return“ pattern often found in C.
 
 One of the limitations of the option type is it doesn't record the reason which
```
christinerose:
```suggestion
One of the limitations of the option type is that it doesn't record the reason that
```

data/tutorials/lg_08_error_handling.md
```diff
@@ -574,20 +577,20 @@ robustness. A couple of observations:
...
 what went wrong. For this reason, functions returning option values should
-document the circumstances under which it may return `None`. Such a
-documentation is likely to ressemble to the one required for exceptions using
-`@raise`. The **`result`** type is intended to fill this gap: manage error in data,
-like option values, but also provide information on errors, like exceptions. It
-is the topic of the next section.
+document the circumstances under which they may return `None`. Such
+documentation is likely to resemble that required for exceptions using `@raise`.
+The **`result`** type, described in the next section, is intended to fill this
+gap: manage errors in data like option values, but also provide information on
```
christinerose:
```suggestion
gap: manage errors in data like option values and provide information on
```
Keep "but also" if the second part wouldn't normally be expected to follow

data/tutorials/lg_08_error_handling.md
```diff
@@ -660,17 +663,17 @@ val Option.to_result : none:'e -> 'a option -> ('a, 'e) result
...
-- `Yaml.find` recursively searches a key in a Yaml tree, if found, it returns
+- `Yaml.of_string` parses a string and turns it into an ad hoc OCaml type
+- `Yaml.find` recursively searches for a key in a Yaml tree. If found, it returns
   the corresponding data, wrapped in an option
 - `Option.to_result` performs conversion of an **`option`** into a **`result`**.
 - Finally, `let*` stands for `Result.bind`.
 
-Since functions from the `Yaml` module both returns **`result`** data, it is easier
-to write a pipe which processes that type all along. That's why
+Since both functions from the `Yaml` module return **`result`** data, it is
```
christinerose, line 671:
```suggestion
Since both functions from the `Yaml` module return **`result`** data, it is
```
Unless the actual command is capitalized like `Yaml`, I suggest keeping it all lower case `yaml` (or exactly how it appears in the command). If it's not a command, then it needs to be all capitalized like YAML. Basically match the formatting they use here: https://yaml.org/

data/tutorials/lg_08_error_handling.md
```diff
@@ -660,17 +663,17 @@ val Option.to_result : none:'e -> 'a option -> ('a, 'e) result
...
+easier to write a pipe which processes that type all along. That's why
 `Option.to_result` needs to be used. Stages which produce **`result`** must be
 chained using `bind`, stages which do not must be chained using some map
```
christinerose:
```suggestion
chained using `bind`; stages which do not must be chained using some map
```

data/tutorials/lg_08_error_handling.md
```diff
@@ -758,10 +763,10 @@ c
 a >>= fun x -> b >>= fun y -> c
 ```
 Variables `x` and `y` may appear in `c` in the three cases. The first form isn't
-very convenient, it uses a lot of parenthesis. The second one is often the
-prefered one due to its ressemblance with regular local definitions. The third
+very convenient, it uses a lot of parentheses. The second one is often the
```
christinerose:
```suggestion
very convenient, as it uses a lot of parentheses. The second one is often the
```
...or "and it uses..." ?

data/tutorials/lg_08_error_handling.md
```diff
@@ -758,10 +763,10 @@ c
...
+preferred one due to its resemblance with regular local definitions. The third
```
christinerose:
```suggestion
preferred due to its resemblance with regular local definitions. The third
```

data/tutorials/lg_08_error_handling.md
```diff
@@ -774,8 +779,8 @@ let* y = f x in
 g y
 ```
 Writing `x >>= f` is very close to what is found in functionally tainted
-programming languages, which have methods and receivers such as Kotlin, Scala,
-Go, Rust, Swift or even modern Java, where it would be looking like:
+programming languages which have methods and receivers such as Kotlin, Scala,
+Go, Rust, Swift or even modern Java, where it would look like:
```
christinerose:
```suggestion
Go, Rust, Swift, or even modern Java, where it would look like:
```

data/tutorials/lg_08_error_handling.md
```diff
@@ -789,35 +794,37 @@ File.read_opt path
 |> Result.map_error (Printf.sprintf "%s, error: %s: " path)
 ```
 
-By the way, this style is called [Tacit
-Programming](https://en.wikipedia.org/wiki/Tacit_programming). Thanks to the
-associativity priorities of the `>>=` and `|>` operators, no parenthesis extends beyond a single line.
+By the way, this style is called [tacit
+programming](https://en.wikipedia.org/wiki/Tacit_programming). Thanks to the
+associativity priorities of the `>>=` and `|>` operators, no parenthesized
```
christinerose:
```suggestion
associativity priorities of the `>>=` and `|>` operators, no parenthesised
```

data/tutorials/lg_08_error_handling.md
```diff
@@ -789,35 +794,37 @@ File.read_opt path
...
+expression extends beyond a single line.
 
 OCaml has a strict typing discipline, not a strict styling discipline;
 therefore, picking the right style is left to the author's decision. That
 applies error handling, pick a style knowingly. See the [OCaml Programming
```
christinerose:
```suggestion
applies error handling, so pick a style knowingly. See the [OCaml Programming
```
Or if not "so," make it a separate sentence: Pick a style knowingly.

data/tutorials/lg_08_error_handling.md
```diff
@@ -868,16 +875,16 @@ Running this on Win32, of course, won't throw an error.
 
 Writing `assert false` would just stop your program. This idiom is sometimes
 used to indicate [dead code](https://en.wikipedia.org/wiki/Dead_code), parts of
-the program that must be writen (often for type-checking or pattern matching
+the program that must be written (often for type-checking or pattern matching
```
christinerose:
```suggestion
the program that must be written (often for type-checking or pattern-matching
```

data/tutorials/lg_08_error_handling.md
```diff
@@ -868,16 +875,16 @@ Running this on Win32, of course, won't throw an error.
...
 completeness) but are unreachable at run time.
 
-Asserts should be understood as executable comments. There aren't supposed to
+Asserts should be understood as executable comments. They aren't supposed to
 fail, unless during debugging or truly extraordinary circumstances that
 absolutely prevent the execution from making any kind of progress.
 
 When the execution reaches conditions which can't be handled, the right thing to
 do is to throw a `Failure`, using `failwith "error message"`. Assertions aren't
```
christinerose:
```suggestion
do is to throw a `Failure` by using the `failwith "error message"`. Assertions aren't
```

data/tutorials/lg_08_error_handling.md
```diff
@@ -907,9 +914,10 @@ several other mainstream languages, OCaml provides several mechanisms to handle
 exceptional events, all with good runtime performance and code
 understandability. Using them properly requires some initial learning and
 practice. Later, it always requires some thinking, which is good since proper
```
christinerose:
```suggestion
practice. Later, it always requires some thinking, which is beneficial because proper
```

data/tutorials/lg_08_error_handling.md
```diff
@@ -907,9 +914,10 @@ several other mainstream languages, OCaml provides several mechanisms to handle
...
-error management shouldn't ever be overlooked. No error handling is better
-than the others, and is should be matter of adequacy to the context rather than
-of taste. But opinionated OCaml code is also fine, so it's a balance.
+error management shouldn't ever be overlooked. No error handling mechanism is
+always better than the others, and choosing one to use should be matter of
+fitting the context rather than of taste. But opinionated OCaml code is also
+fine, so it's a balance.
```
christinerose:
```suggestion
error management shouldn't ever be overlooked. No error handling mechanism is
always better than the others, and choosing one to use should be a matter of
fitting the context rather than that of taste. But opinionated OCaml code is also
fine, so it's a balance.
```

christinerose (changes requested):
Nicely done! Only some grammar/syntax/formatting suggestions and some questions about clarity in a few places.

data/tutorials/lg_08_error_handling.md
cuihtlauac: Good point. Actually, it's above that we need to write YAML. Here, it's a module name, it has to stay like that.

data/tutorials/lg_08_error_handling.md
```diff
@@ -30,7 +32,7 @@ instance, in Unix systems, here what is contained in `man 2 read`:
 >
 > On error, -1 is returned, and `errno` is set to indicate the error.
 
-Great software was written using this style. However, since correct are errors
+Great software was written using this style. However, since correct and error
```
cuihtlauac:
```suggestion
Great software was written using this style. However, since the expected and error
```

data/tutorials/lg_08_error_handling.md
```diff
@@ -139,39 +141,36 @@ desirable for errors that must be handled at the client level.
...
+Note that functions passed to `protect` take only `()` as a parameter. This
+does not limit the cases where `protect` can be used. Any computation `f x` can
+be wrapped in a function `fun () -> f x`. As always, the body of the function
+won't be evaluated until the function is called.
 
-The `finally` function is only expected to perform some side-effect. In summary,
-`protect` performs two computations in order: `work` and then `finally`, and
-forwards the outcome of `work` either result or exception.
+The `finally` function is only expected to perform some side-effect, and should
+not raise any exception itself. If `finally` does throw some exception `e`,
```
cuihtlauac:
```suggestion
not raise any exception itself. If `finally` does throw an exception `e`,
```

data/tutorials/lg_08_error_handling.md
```diff
@@ -639,7 +642,7 @@ let map_error f = function
 
 The same reasoning applies to `Result.bind`, except there's no `bind_error`.
 Using those functions, here is an hypothetical example of code using [Anil
-Madhavapeddy OCaml Yaml library](https://github.com/avsm/ocaml-yaml):
+Madhavapeddy's OCaml Yaml library](https://github.com/avsm/ocaml-yaml):
```
cuihtlauac:
```suggestion
Madhavapeddy's OCaml YAML library](https://github.com/avsm/ocaml-yaml):
```

data/tutorials/lg_08_error_handling.md
```diff
@@ -868,16 +875,16 @@ Running this on Win32, of course, won't throw an error.
...
+the program that must be written (often for type-checking or pattern-matching
 completeness) but are unreachable at run time.
 
-Asserts should be understood as executable comments. There aren't supposed to
+Asserts should be understood as executable comments. They aren't supposed to
 fail, unless during debugging or truly extraordinary circumstances that
 absolutely prevent the execution from making any kind of progress.
 
 When the execution reaches conditions which can't be handled, the right thing to
 do is to throw a `Failure`, using `failwith "error message"`. Assertions aren't
```
cuihtlauac:
```suggestion
do is to throw a `Failure`, by calling `failwith "error message"`. Assertions aren't
```

data/tutorials/lg_08_error_handling.md
```diff
@@ -139,39 +141,36 @@ desirable for errors that must be handled at the client level.
...
 This function is meant to be used when something _always_ needs to be done
-_after_ a computation is complete, either it succeded or failed. Any computation
-can be postponed by wrapping it into a dummy function with only `()` as
- parameter. Here, the computation triggered by passing `x` to `f` (including its
- side effects) will not take place:
-```ocaml
-let work () = f x
-```
+_after_ a computation is complete, whether it succeeds or fails. The unlabeled
+function argument is called first, then the function passed as the
```
cuihtlauac:
```suggestion
function argument is called first, then the function passed as the
```

cuihtlauac (changes requested):

data/tutorials/lg_08_error_handling.md
christinerose: Okay, cool. I figured it was a different format since it was a module, but should it then be `yaml` or `Yaml`?

christinerose (approved):

data/tutorials/lg_08_error_handling.md
cuihtlauac: That's super tricky. 
- Technology Name: YAML
- OCaml module name : Yaml
- Opam package name: yaml

data/tutorials/lg_08_error_handling.md
christinerose: Is it "...since the expected error messages can't be distinguished" ?

Or do you mean ... the expected messages including error messages?

cuihtlauac: Maybe something like: “The values which are expected can't be distinguished from the values which represent errors”

data/tutorials/lg_08_error_handling.md
cuihtlauac:
```suggestion
Great software was written using this style. However, as expected return values can't be distinguished from values representing errors, nothing but the programmer's discipline ensures
```

cuihtlauac (changes requested):

data/tutorials/lg_08_error_handling.md
tuohy: "some one" and "some thing" are so commonplace they have turned into words. "an exception" loses the connotation that this is contrary to expectation. I would stand by "some exception" but "an exception" means almost exactly the same.

data/tutorials/lg_08_error_handling.md
tuohy: Need to delete "the" at the end of previous line

data/tutorials/lg_08_error_handling.md
```diff
@@ -758,10 +763,10 @@ c
 a >>= fun x -> b >>= fun y -> c
 ```
 Variables `x` and `y` may appear in `c` in the three cases. The first form isn't
-very convenient, it uses a lot of parenthesis. The second one is often the
-prefered one due to its ressemblance with regular local definitions. The third
+very convenient, as it uses a lot of parentheses. The second one is often the
```
tuohy:
```suggestion
very convenient, as it uses a lot of parentheses. The second one is often
```
The change to the next line makes the final "the" incorrect

christinerose (approved):

cuihtlauac (approved):

christinerose (approved):

sabine:
Thanks @tuohy, @cuihtlauac, @christinerose!

PR #1345, opened 2023-06-28
add workflow to annotate broken links in markdown

Shreyas-21:


sabine (approved):
Thanks @Shreyas-21, this is so useful!

PR #1240, opened 2023-05-31
Add changelog and blog post for opam 2.1.5 release

tmattio:
cc @hannesm @rjbou

PR #1351, opened 2023-06-29
Add attribution to blog posts

reynir:
This is low effort and doesn't consider what css *should* be used. I have not tested it. It should hopefully not be difficult for someone with insight into the website to remedy this PR.

Motivated by https://github.com/ocaml/ocaml.org/pull/1240


tmattio (approved):
Thanks a lot @reynir!

I pushed a compilation fix and a minor stylistic change.

Issue #1322, opened 2023-06-17
Cannot open website without intenet connection


AshineFoster:
I am not sure if this is the intended behaviour but I cannot open the website if I do not have an internet connection. It gives me the following when I run `make start` :
```sh
opam exec -- dune build --root .
opam exec -- dune exec src/ocamlorg_web/bin/main.exe
Saving build trace db...main.exe: [INFO] State cache file: /home/ashine/.cache/ocamlorg/package.state
main.exe: [INFO] Package state loaded (4323 packages, opam commit 01f6a0905760e389a567ee124839f7b2e4c3864e)
17.06.23 18:26:34.050                       Running on 0.0.0.0:8080 (http://localhost:8080)
17.06.23 18:26:34.050                       Type Ctrl+C to stop
fatal: unable to access 'https://github.com/ocaml/opam-repository.git/': Could not resolve host: github.com
Fatal error: exception Failure("(\"git\", \"git\" \"-C\" \"/home/ashine/.cache/ocamlorg/opam-repository\" \"pull\" \n\"-q\" \"--ff-only\" \"origin\" \"master\") exited with status 1")
Raised at Stdlib.failwith in file "stdlib.ml", line 29, characters 17-33
Called from Lwt.Sequential_composition.map.create_result_promise_and_callback_if_deferred.callback in file "src/core/lwt.ml", line 1966, characters 41-46
make: *** [Makefile:37: start] Error 2
```
I would like to be able to work on the website even without internet. Is there some option I can use to enable this?

cuihtlauac:
Hi @AshineFoster; as the server needs to connect to several backends, it assumes a connection to the Internet. It should be possible to patch to enable running in a degraded mode, without access to the Internet, for debugging. Everything related to packages would be broken, but the rest could be served over localhost. I believe the errors you are seeing are coming from `src/ocamlorg_package/lib/opam_repository.ml`, this is where attempts to connect to github takes place.

Hope it helps.

AshineFoster:
Thank you. I will try to edit that file to add an option to work offline.

Issue #1208, opened 2023-05-23
Default image for RSS items that haven't got an image


cuihtlauac:
Blog entries that don't have an image look a bit dull compared to those having one. It would be nice to have a default image and display that for blog posts that don't have one.

We could use/modify the ocaml logo or an image of camels. Open to other ideas.

This is a simple issue suitable for an OCaml beginner as there's only two things to do:
1. Add a suitable image with no legal conflicts to the `asset/` folder, and
2. Modify the template https://github.com/ocaml/ocaml.org/blob/main/src/ocamlorg_frontend/pages/blog.eml#L30 to render the image when a blog post doesn't have one.

PR #1366, opened 2023-07-03
Add basic offline command `make offline`

AshineFoster:
This adds basic support for running the website without internet access from issue #1322.


sabine:
Thanks @AshineFoster, this is one way to make offline development of the site possible, and it seems to work!

I thought a bit about this, and I think the better way to achieve this is to not add a command line option.

Instead, it would be sensible to make `Opam_repository.pull` fail gracefully around https://github.com/ocaml/ocaml.org/blob/main/src/ocamlorg_package/lib/ocamlorg_package.ml#L161 in `Ocamlorg_package` and log a message that pulling from opam repository was not possible and that we're continuing to build the package state using the existing local state of opam-repository.

When `Opam_repository.clone` fails, there's no chance to run offline, so this would need to still fail.

Are you up for modifying your patch in that direction?

AshineFoster:
Sure. I will give it another look and implement your suggestion.

christinerose (approved):

christinerose (approved):

sabine:
Hey @AshineFoster, I think something went a little wrong when merging the two branches because one of them was going off a more recent version of the `main` branch.

One way to fix it is to `git rebase main` after making sure that your main branch is up to date. :slightly_smiling_face:

Another way would be `git cherry-pick`ing the commits onto a fresh branch and then replacing this one when satisfied.

AshineFoster:
I did a rebase to fix the error. Thank you for the pointers.

Makefile
```diff
@@ -36,6 +36,10 @@ install: all ## Install the packages on the system
 start: all ## Run the produced executable
 	opam exec -- dune exec src/ocamlorg_web/bin/main.exe
 
+.PHONY: offline
+offline: all ## Run the produced executable without an internet connection. Some functionality will not be avaliable.
+	opam exec -- dune exec src/ocamlorg_web/bin/main.exe -- -offline
+
```
sabine: I think this can be removed.

src/ocamlorg_package/lib/opam_repository.ml
```diff
@@ -16,6 +16,15 @@ module Process = struct
...
     | Unix.WEXITED 0 -> ()
+    | Unix.WEXITED 1 ->
+        if Pull.is_pull () then
+          prerr_endline
+            "\n\
+             Pulling from opam repository was not possible. \n\
+             We will continue by building the package state using the existing \
+             local state of opam-repository.\n\
+             Some functionality might not work properly.\n"
+        else Fmt.failwith "%a %a" pp_cmd cmd pp_status (Unix.WEXITED 1)
```
sabine: I am conflicted by this.

The module `Process` seems to try to encapsulate calling arbitrary processes. So we can't really introduce anything specific to the pull request here.

A global reference (from a perspective of code health) should be a last resort strategy when there's no other, more local solution.

My impression is that `check_status` deciding that we `Fmt.failwith` is wrong in the first place.

The sensible thing for `check_status` to do would be to return a `Result` value. So it would either return `Ok` with the value we get in the successful case, or it would return `Error` with the formatted error message. Then the call sites of the `Process.exec`, `Process.pread`, `Process.pread_lines` decide, if an error is worth throwing an exception or not.

sabine (commented):
It seems to be a bit trickier than I thought. The direction is good, but I think the existing `Process` as a module is overstepping its responsibilities when it decides to throw an exception in case of a non-zero status code.

src/ocamlorg_package/lib/opam_repository.ml
AshineFoster: Sorry for the late reply. I removed the global reference which was indeed bad design. I agree with your opinion about the `Process ` module but feel it is out of scope for this pull request.  I can open an issue about it and try to implement your suggestions if that is ok with you.

sabine: https://github.com/ocaml/ocaml.org/pull/1366/commits/dbf21f049a82f5dbdef2ab67eb7b6914391e12e7 looks reasonable and mergeable to me.

I'm super happy when someone tidies up - it's very helpful and a great learning opportunity! There's also other things you could do, like https://github.com/ocaml/ocaml.org/issues/1208 (displaying a default image for RSS feed entries that don't have one). Pretty sure we could find something a bit more challenging than that as well.

sabine (approved):

sabine:
Thanks @AshineFoster, now ocaml.org can run offline, to enable people in environments with flaky or non-existant internet connection to work on and test the codebase!

PR #1369, opened 2023-07-04
Promote the changelog to non-experimental

tmattio:


sabine:
You need to add another way than the banner to get to the changelog. :grin:

cuihtlauac:
Up to my understanding, the ways to access the changelog section are as follows:

* Direct URL
* Footer link

Should we have some dedicated section/text in the "/community" page?

sabine:
I think a good place would be the main landing page, but we probably also need a banner maybe above the "News" section on the blog page.

We will realize the final positions when we do the redesign of landing page and community section - probably Q4.

tmattio:
Now that we can link to changelog entries directly with #1394, I suggest we move ahead with this PR nonetheless, and revisit how we can highlight the Changelog later.

sabine (approved):

PR #1350, opened 2023-06-29
Tidy up tailwind.config.js color names in preparation for an eventual dark mode 

SaySayo:
This PR adds a work-in-progress dark mode to Ocaml.org. cc @sabine

PR #1375, opened 2023-07-04
Fix copy-to-clipboard styles

sabine:
Reducing the available text colors in https://github.com/ocaml/ocaml.org/pull/1350 to a smaller set broke the copy-to-clipboard widgets.

This patch allows all Tailwind default colors as text color in tailwind.config.js - for some reason, Tailwind doesn't check that all the colors that are used as text color are actually available. I would have expected Tailwind to fail on this.

To identify exactly which default Tailwind colors we're using, I added a simple bash script to the repository under `scripts/scan-tailwind-colors.sh`. This helps with aligning the colors we use in the templates with the colors used in Figma.

### DIFF
--- a/scripts/scan-tailwind-colors.sh
+++ b/scripts/scan-tailwind-colors.sh
@@ -1,0 +1,14 @@
+#!/bin/bash
+
+DIR="src/ocamlorg_frontend"
+
+TMP_FILE="tmp.txt"
+
+# Write all matches to temporary file
+grep -r -o -h -E "text-\w+-[0-9]{2,3}|bg-\w+-[0-9]{2,3}" $DIR > $TMP_FILE
+
+# Filter unique matches and print them
+sort $TMP_FILE | uniq
+
+# Remove temporary file
+rm $TMP_FILE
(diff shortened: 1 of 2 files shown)


scripts/scan-tailwind-colors.sh
```diff
@@ -0,0 +1,14 @@
...
+
+DIR="src/ocamlorg_frontend"
+
+TMP_FILE="tmp.txt"
+
+# Write all matches to temporary file
+grep -r -o -h -E "text-\w+-[0-9]{2,3}|bg-\w+-[0-9]{2,3}" $DIR > $TMP_FILE
+
+# Filter unique matches and print them
+sort $TMP_FILE | uniq
```
cuihtlauac, line 11:
```suggestion
sort -u $TMP_FILE
```

cuihtlauac (approved):

scripts/scan-tailwind-colors.sh
cuihtlauac:
```suggestion
sort -u $TMP_FILE
```

cuihtlauac (changes requested):

cuihtlauac (approved):

Issue #1346, opened 2023-06-28
Proofreading the sequence tutorial


gmevel:
Hi, I just read [the sequences tutorial](https://ocaml.org/docs/sequences) whose source is [here](data/tutorials/ds_05_seq.md) (commit 0f4aa1ec), here is some feedback (my apologies for this not being a PR; writing is too much of a rabbit hole for me).

 1. First, a high-level opinion: the tutorial starts with:

    > Sequences are very much like lists. However from a pragmatic perspective, one should imagine they may be infinite. That's the key intuition to understanding and using sequences.

    IMHO, the key idea about sequences is that **their elements are computed on-demand, and not stored in memory**. I believe *this* is what should be highlighted in the very first (well, second) sentence of a tutorial. The fact that a sequence might be infinite is a corollary of this. Another corollary, perhaps more frequently used, is that sequences often allow to reduce memory consumption from linear to constant space.

 1.   Still in the intro: for people familiar with Python, I believe it would be very useful to mention Python’s generators: OCaml sequences are similar to Python generators. The main difference is that each element of a Python generator is consumed only once and never seen again, while an element in an OCaml sequence can be queried several times, but is re-computed each time (more expressive but opportunity for bugs). Also, OCaml does not have all the convenient syntax that Python has (there is no `yield` in OCaml [at least, until algebraic effects land in OCaml 5!]). The contrast between `list` and `Seq.t` in OCaml is the same as between `range` and `xrange` in old Python 2 (or `[i for i in range(100)]` versus `range(100)` in Python 3).

 1. > with a twist when it's not empty: a frozen tail

    The wording seems weird to me. Reword as `with a twist when it's not empty: its tail is frozen` ?

 1. > All have unforeseeable termination, and it is easier to consider them infinite.

    Well, *potentially* infinite: as the consumer of a sequence, you must not forget to deal with `Seq.Nil`, i.e. what to do if the stream ends.

 1. > writing `fun _ -> a`, which has type `'a -> t`. When writing `fun () -> a` instead, we get a function of type `unit -> t`.

    Not sure whether this detour about polymorphism is useful.

 1. > In reality, since there isn't an infinite amount of distinct values of type `int`, those sequences aren't indefinitely increasing. When reaching `max_int`, the values will circle down to `min_int`. They are ultimately periodic.

    Overly long digression w.r.t. the main topic which is `Seq.t`. I would cut this remark, or at least shorten it and put it in parentheses.

 1. > Writing this:
    > 
[...]

PR #1376, opened 2023-07-05
Include @gmevel proof-reading of Seq tutorial

cuihtlauac:
Items from issue #1346 where either:
- Immediately included, when obvious
- Inlined as comment, when further discussion/rewriting is needed

After a couple of commits here are the remaining open items:

1. Comparison with Pythons generators
2. The real Eratosthenes sieve as an exercice
3. Example in section “Sequences Are Functions”

### DIFF
--- a/data/tutorials/language/3ds_05_seq.md
+++ b/data/tutorials/language/3ds_05_seq.md
@@ -18,8 +18,16 @@ Sequences are very much like lists. However from a pragmatic perspective, one
 should imagine they may be infinite. That's the key intuition to understanding
 and using sequences.
 
+<!--
+IMHO, the key idea about sequences is that their elements are computed on-demand, and not stored in memory. I believe this is what should be highlighted in the very first (well, second) sentence of a tutorial. The fact that a sequence might be infinite is a corollary of this. Another corollary, perhaps more frequently used, is that sequences often allow to reduce memory consumption from linear to constant space.
+-->
+
+<!--
+Still in the intro: for people familiar with Python, I believe it would be very useful to mention Python’s generators: OCaml sequences are similar to Python generators. The main difference is that each element of a Python generator is consumed only once and never seen again, while an element in an OCaml sequence can be queried several times, but is re-computed each time (more expressive but opportunity for bugs). Also, OCaml does not have all the convenient syntax that Python has (there is no yield in OCaml [at least, until algebraic effects land in OCaml 5!]). The contrast between list and Seq.t in OCaml is the same as between range and xrange in old Python 2 (or [i for i in range(100)] versus range(100) in Python 3).
+-->
+
 One way to look at a value of type `'a Seq.t` is to consider it as a list, with
-a twist when it's not empty: a frozen tail. To understand this analogy,
+a twist when it's not empty: its tail is frozen. To understand this analogy,
 consider how sequences are defined in the standard library:
 ```ocaml
 type 'a node =
@@ -54,11 +62,10 @@ Frozen-by-function tails explain why sequences may be considered potentially
 infinite. Until a `Seq.Nil` value has been found in the sequence, one can't say
 for sure if some will ever appear. The sequence could be a stream of incoming
 requests in a server, readings from an embedded sensor, or system logs. All have
-unforeseeable termination, and it is easier to consider them infinite.
+unforeseeable termination, and it is easier to consider them potentially infinite.
 
 In OCaml, any value `a` of type `t` can be turned into a constant function by
-writing `fun _ -> a`, which has type `'a -> t`. When writing `fun () -> a`
-instead, we get a function of type `unit -> t`. Such a function is called a
+writing `fun _ -> a` or `fun () -> a`. The latter function is called a
 [_thunk_](https://en.wikipedia.org/wiki/Thunk). Using this terminology, `Seq.t`
 values are thunks. With the analogy used earlier, `a` is frozen in its thunk.
 
@@ -68,10 +75,9 @@ Here is how to build seemingly infinite sequences of integers:
 val ints : int -> int Seq.t = <fun>
 ```
 The function `ints n` looks as if building the infinite sequence `(n; n + 1; n +
-2; n + 3;...)`. In reality, since there isn't an infinite amount of distinct
-values of type `int`, those sequences aren't indefinitely increasing. When
-reaching `max_int`, the values will circle down to `min_int`. They are
-ultimately periodic.
+2; n + 3;...)`. In reality, since machine integers have bounds, the sequence
+isn't indefinitely increasing. When reaching `max_int`, it will circle
+down to `min_int`.
 
 The OCaml standard library contains a module on sequences called
 [`Seq`](/releases/5.0/api/Seq.html). It contains a `Seq.iter` function, which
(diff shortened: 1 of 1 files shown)


data/tutorials/ds_05_seq.md
```diff
@@ -17,10 +17,16 @@ You should be comfortable with writing functions over lists and options.
 
 Sequences are very much like lists. However from a pragmatic perspective, one
```
christinerose:
```suggestion
Sequences are very much like lists. However, from a pragmatic perspective, one
```

data/tutorials/ds_05_seq.md
```diff
@@ -17,10 +17,16 @@ You should be comfortable with writing functions over lists and options.
...
 should imagine they may be infinite. That's the key intuition to understanding
-and using sequences.
+and using sequences. To achieve this, sequence elements are computed on demand,
```
christinerose:
```suggestion
and using sequences. To achieve this, sequence elements are computed on demand
```

data/tutorials/ds_05_seq.md
```diff
@@ -17,10 +17,16 @@ You should be comfortable with writing functions over lists and options.
...
+and not stored in memory. Perhaps more frequently, sequences also allow for
+reducing memory consumption from linear to constant space
+
+<!--
+Still in the intro: for people familiar with Python, I believe it would be very useful to mention Python’s generators: OCaml sequences are similar to Python generators. The main difference is that each element of a Python generator is consumed only once and never seen again, while an element in an OCaml sequence can be queried several times, but is re-computed each time (more expressive but opportunity for bugs). Also, OCaml does not have all the convenient syntax that Python has (there is no yield in OCaml [at least, until algebraic effects land in OCaml 5!]). The contrast between list and Seq.t in OCaml is the same as between range and xrange in old Python 2 (or [i for i in range(100)] versus range(100) in Python 3).
+-->
 
 One way to look at a value of type `'a Seq.t` is to consider it as a list, with
```
christinerose:
```suggestion
One way to look at a value of type `'a Seq.t` is to consider it as a list, but it contains
```

data/tutorials/ds_05_seq.md
```diff
@@ -69,34 +74,43 @@ Here is how to build seemingly infinite sequences of integers:
...
-code is also an infinite loop:
+in an OCaml toplevel means “print integers forever,” and you have to press
+`Ctrl-C` to interrupt the execution. The following code is the same infinite
+loop without any output:
 ```ocaml
 # Seq.iter ignore (ints 0);;
 ```
-But the key point is: it doesn't leak memory.
+The key point is: it doesn't leak memory. This example is running in constant
+space, it is effectively nothing more than an infinite loop, which can be
```
christinerose:
```suggestion
space. It is effectively nothing more than an infinite loop, which can be
```

data/tutorials/ds_05_seq.md
```diff
@@ -69,34 +74,43 @@ Here is how to build seemingly infinite sequences of integers:
...
+confirmed by monitoring the space consumption of the program, and by noticing
```
christinerose:
```suggestion
confirmed by monitoring the space consumption of the program and by noticing
```

data/tutorials/ds_05_seq.md
```diff
@@ -204,7 +224,7 @@ Here is a quick check:
 - : int list = [0; 1; 4; 9; 16; 25; 36; 49; 64; 81]
 ```
 The function `Seq.uncons` returns the head and tail of a sequence if it is not
-empty, or it otherwise returns `None`.
+empty otherwise, it returns `None`.
```
christinerose:
```suggestion
empty. Otherwise, it returns `None`.
```

data/tutorials/ds_05_seq.md
```diff
@@ -225,20 +245,67 @@ taking care of open files. While the code above is fine, this one no longer is:
...
+        Skip (prime_list, n+1)
+    end
+    ([], 2)
+```
+-->
+
 ## Sequences Are Functions
 
+<!--
+No need to introduce another mathematical sequence, we can re-use earlier examples (better for pedagogy):
```
christinerose:
```suggestion
No need to introduce another mathematical sequence, we can reuse earlier examples (better for pedagogy):
```

data/tutorials/ds_05_seq.md
```diff
@@ -225,20 +245,67 @@ taking care of open files. While the code above is fine, this one no longer is:
...
+let rec ints n = Seq.cons n (ints (n+1))
+```
+-->
 Although this looks like a possible way to define the [Fibonacci
 sequence](https://en.wikipedia.org/wiki/Fibonacci_number):
 ```ocaml
 # let rec fibs m n = Seq.cons m (fibs n (n + m));;
 val fibs : int -> int -> int Seq.t = <fun>
 ```
 It actually isn't. It's a non-ending recursion which blows away the stack.
```
christinerose:
```suggestion
It actually isn't. It's an unending recursion which blows away the stack.
```

christinerose (approved):
Great work! Just found a few minor grammatical things to fix.

christinerose (approved):

data/tutorials/ds_05_seq.md
```diff
@@ -331,4 +391,5 @@ OCaml 4.14. Beware books and documentation written before may still mention it.
   * Guillaume Petiot [@gpetiot](https://github.com/gpetiot)
   * Xavier Van de Woestyne [@xvw](https://github.com/xvw)
   * Simon Cruanes [@c-cube](https://github.com/c-cube)
+  * Glen Mével [gmevel](https://github.com/c-cube)
```
gmevel:
```suggestion
  * Glen Mével [gmevel](https://github.com/gmevel)
```

c-cube is someone else. ;-) But I don’t care being credited.

gmevel:
Thanks! I obviously have nothing against it. :-) Just a remark on your list of TODOs:

> 2. The real Eratosthenes sieve as an exercice

I think there has been a misunderstanding: the sieve of Eratosthenes *needs* indexed accesses, you cannot implement it (meaningfully) with `Seq.t`. For me it is enough to rebrand your existing example as “a variant of the trial-division algorithm” rather than as “the Eratosthenes sieve”.

cuihtlauac:
> > 2. The real Eratosthenes sieve as an exercice
> 
> I think there has been a misunderstanding: the sieve of Eratosthenes _needs_ indexed accesses, you cannot implement it (meaningfully) with `Seq.t`. For me it is enough to rebrand your existing example as “a variant of the trial-division algorithm” rather than as “the Eratosthenes sieve”.

I was thinking of turning your versions of `primes` into an exercise or using one of the versions from Melissa O’Neill paper. She calls hers “An Incremental Functional Sieve”, how would you call yours?

gmevel:
> > > 2. The real Eratosthenes sieve as an exercice
> > 
> > 
> > I think there has been a misunderstanding: the sieve of Eratosthenes _needs_ indexed accesses, you cannot implement it (meaningfully) with `Seq.t`. For me it is enough to rebrand your existing example as “a variant of the trial-division algorithm” rather than as “the Eratosthenes sieve”.
> 
> I was thinking of turning your versions of `primes` into an exercise or using one of the versions from Melissa O’Neill paper. She calls hers “An Incremental Functional Sieve”, how would you call yours?

IMHO these examples involve mathematical knowledge and clever algorithmic that is entirely independent from the topic of `Seq.t`, so I’m not sure of the relevance of such an exercise in the context of this tutorial.

-----

With that said, let me answer your question, keeping in mind that this is nothing but a large digression.

Thanks for mentioning [O’Neill’s paper][oneill]! I wasn’t aware of that. This functional version is a pearl and I’d like to try it.

My algorithm (I assume you are referring to [my library `euler`][euler]) is a sieve of Eratosthenes with a mutable array of booleans. It is “segmented” (i.e. it loops over an array of a fixed size rather than allocating a large array of a length equal to the maximum integer). Essentially, this is made possible by keeping track of the prime numbers we have found so far and, for each of them, what is the latest multiple of the prime that we have run over.

O’Neill pushes this idea further, all the way to getting a purely functional algorithm where there is no mutable array anymore. I hadn’t thought of that! To be tested in practice, but this seems better to me for a variety of reasons:

  1. the usual advantages of purely functional programming versus mutation;
  2. it might save a (constant but) large amount of space, by not having an array of booleans at all (what remains to be seen is the overhead of her priority queue over my simpler list/array of primes);
  3. most importantly (and relevant to this discussion): it naturally produces a `Seq.t` (traversal is controlled by the consumer, who chooses if and when to compute the next prime); whereas my version is an iterator (traversal is controlled by the producer, in other words, it runs wholly in a single row) and it is not easy to turn it into a `Seq.t` (which would definitively be transient, because of mutation).

Both O’Neill’s code and mine feature the constant-time optimization which she calls a “wheel”, and which in my code I call pre-culling, i.e skipping the multiples of small prime numbers.

So to answer your question, “how to call my sieve”, I would say “an (imperative) segmented sieve of Eratosthenes with a wheel”… but that’s of no use to you since there are no `Seq.t` involved. :-)

[euler]: https://github.com/gmevel/euler-lib
[oneill]: https://www.cs.hmc.edu/~oneill/papers/Sieve-JFP.pdf

christinerose (approved):

sabine (approved):

PR #1388, opened 2023-07-11
Install autoconf in the Docker container

sabine:
Build was failing at installing `conf-autoconf.0.1`. This installs `autoconf` in the docker container.


cuihtlauac:
This is strange. Why is this coming?

sabine:
I agree that this is strange, since we're pinning `opam-repository`, I would have expected no new dependencies popping up like this.

cuihtlauac (approved):

PR #1389, opened 2023-07-12
Tab Component For Learn Area Redesign

SaySayo:
Created a Tab Component To Learn Area Redesign.
Learn area tab component
cc @sabine

### DIFF
--- a/src/ocamlorg_frontend/layouts/learn_layout.eml
+++ b/src/ocamlorg_frontend/layouts/learn_layout.eml
@@ -1,45 +1,41 @@
 type active_icon_link =
-  | Learn
+  | Overview
+  | GetStarted
+  | Platform
+  | BestPractices
   | Exercises
+  | Books
+  
 
-let sidebar_icon_link_block
+let tabs
 ~current =
-  <div class="space-y-4">
-    <a href="<%s Url.learn %>" class="flex justify-start items-center">
-      <div class="bg-primary-600 text-white rounded w-8 h-8 flex items-center justify-center">
-        <%s! Icons.book "h-5 w-5" %>
-      </div>
-      <div class="ml-3 font-semibold <%s if current != Learn then "opacity-70" else "" %>">Learn OCaml</div>
+  let link ~href ~title ~current =
+    <a href="<%s href %>" class="justify-start px-4 py-2 items-center font-semibold border-2 border-b-4 border-transparent rounded <%s if current then "no-underline text-primary-700 border-b-primary-700" else "opacity-80" %>">
+      <%s title %>
     </a>
+  in
+    <div class="bg-tab-overview text-default">
+      <nav class="container-fluid wide flex flex-wrap justify-around">
+        <%s! link ~href:Url.learn ~title:"Overview" ~current:(current = Overview) %>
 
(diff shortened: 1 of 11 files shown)


tailwind.config.js
```diff
@@ -54,6 +54,7 @@ module.exports = {
 
       "search-keyboard-cursor": "#0C3B8C", // background for cursor highlighting in keyboard navigable areas (e.g. package search dropdown)
       "search-term-highlight": "rgb(221, 232, 251)",
+      "tab-overview": "rgba(211, 213, 249, 0.60)",
```
sabine: This color seems to occur in the designs in more places than just the Learn Area tabs. Here's a potentially better interim name until we choose one in the Figma documents:

```suggestion
      "learn-area-light-blue": "rgba(211, 213, 249, 0.60)",
```

sabine (commented):
Here's two observations:

1.
Screenshot from 2023-07-13 12-21-44
There used to be a gap between the section titles in the sidebar and the previous block (e.g. between "OCaml on Windows" and "LANGUAGE"). While removing the margin at the very top is indeed correct, we still need the spacing between those link blocks. The simple solution is to add a parent div in `learn_sidebar.eml` with a flex column layout with a gap matching the old margins.

2.
Screenshot from 2023-07-13 12-23-25
Screenshot from 2023-07-13 12-23-31
The alignment of the links with Icons and the ones without is slightly off. The simple solution is to give the links with Icons the same transparent borders as the other links.

src/ocamlorg_frontend/layouts/learn_layout.eml
```diff
@@ -1,45 +1,41 @@
...
-      <div class="bg-primary-600 text-white rounded w-8 h-8 flex items-center justify-center">
-        <%s! Icons.book "h-5 w-5" %>
-      </div>
-      <div class="ml-3 font-semibold <%s if current != Learn then "opacity-70" else "" %>">Learn OCaml</div>
+  let link ~href ~title ~current =
+    <a href="<%s href %>" class="justify-start px-4 py-2 items-center font-semibold border-2 border-b-4 border-transparent rounded <%s if current then "no-underline text-primary-700 border-b-primary-700" else "opacity-80" %>">
+      <%s title %>
     </a>
+  in
+    <div class="bg-tab-overview text-default">
```
sabine: Is `text-default` necessary here? It looks like we set the text color on the links, and that's usually the better place to do it.

SaySayo: do you mean I move the text-default color to the nav? The `text-default` class has to be set other the text color would be white

sabine: You explicitly set `text-primary-700` on the `a` elements when current is true. This would be a better, more precise place to set `text-default` in the case it's not current. Relying on inheritance makes the CSS harder to visualize when reading it.

sabine:
Thank you @SaySayo, I rebased on main and left-aligned the tabs. I also saw that the background image from the books section would overlap the tabs due to its negative margin, so I removed that negative margin. IMO good to go.

sabine (approved):

PR #1395, opened 2023-07-14
Add changelog entries for OCaml LSP releases

tmattio:
I noticed that we were missing changelog entries for the recent releases of OCaml LSP.

cc @rgrinberg @voodoos - would be great to have these for future releases if possible, with https://github.com/ocaml/ocaml.org/pull/1394, we'll now link to the OCaml Changelog entries from the newsletter.

### DIFF
--- a/data/changelog/ocaml-lsp/2023-06-21-ocaml-lsp-1.16.1.md
+++ b/data/changelog/ocaml-lsp/2023-06-21-ocaml-lsp-1.16.1.md
@@ -1,0 +1,69 @@
+---
+title: Ocaml-lsp 1.16.1
+date: "2023-06-21"
+tags: [ocaml-lsp, platform, release]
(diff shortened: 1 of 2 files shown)


rgrinberg:
Which entries were missing?

tmattio:
For the releases of OCaml LSP 1.16.1 and 1.16.2

rgrinberg:
I see them in CHANGES.md:

```
# 1.16.2

## Fixes

- Fix file permissions used when specifying output files of pp and ppx. (#1153)

# 1.16.1

## Fixes

- Support building with OCaml 5.0 and 5.1 (#1150)

```

tmattio:
I meant entries on the [OCaml Changelog](https://ocaml.org/changelog), the ones that are being added in this PR.

PR #1427, opened 2023-07-20
Skill Tag Component [Learn Area Redesign]

SaySayo:
Created a skill tag component for learn area redesign.  cc @sabine 
advanced tag

intermediate tag
beginner tag


src/ocamlorg_frontend/components/learn_overview.eml
```diff
@@ -0,0 +1,13 @@
+let skill_tag 
+difficulty =
+  match difficulty with 
+  | Some "beginner" ->
+    <div class="w-28 h-8 p-1 inline-block bg-white rounded-3xl border-2 border-teal-700 font-normal text-center text-teal-700"> Beginner  </div>
```
sabine: The text is not centered because of the explicit height. It's better to do it with adequate padding to achieve the desired height.

src/ocamlorg_frontend/components/learn_overview.eml
```diff
@@ -0,0 +1,13 @@
+let skill_tag 
+difficulty =
+  match difficulty with 
+  | Some "beginner" ->
+    <div class="w-28 py-1 inline-block bg-white rounded-3xl border-2 border-teal-700 font-normal text-center text-teal-700"> Beginner  </div>
```
sabine: usually you would not set an explicit width that depends on the width of the text inside. It's usually more reasonable to define padding. Then, the width can be controlled/affected by the styles of the surrounding elements. (E.g. in a flex-col, this component would take up the full width, whereas in a flex-row it will shrink.)

sabine:
```suggestion
    <div class="py-1 px-2 inline-block bg-white rounded-3xl border-2 border-teal-700 font-normal text-center text-teal-700"> Beginner  </div>
```

sabine: looking at Figma, it looks like a transparent background to me

sabine (approved):
Thanks @SaySayo, let's merge this!

PR #1434, opened 2023-07-21
Playground: Fix And Extend Code-in-URL Handling

### LINKED ISSUE #1433: Playground: loading of editor content from `#code=`  broken
Background: The playground has the feature to share code via a URL using the "share" button, and load code that is included in the URL upon loading the page. The code is encoded in base64 and placed in the fragment identifier of the URL in the format `#code=<base64 data>`.

The problem: the playground writes the base64 string directly to the fragment identifier without URL encoding it. That would not be a problem, since the special characters in base64 (`+`, `/` and `=`) ARE safe in the fragment identifier—unlike in the query parameters, where `+` represents a space. The problem does not occur until loading the code back again: unfortunately, on line 114 the fragment identifier is interpreted as query parameters, so the `+` is parsed to a space!

https://github.com/ocaml/ocaml.org/blob/dff235aa3c0d86acbff071bfef262ee4d214764f/playground/src/main.ml#L112-L115

I would suggest not parsing the fragment identifier as key-value pairs at all—after all, these are not query parameters. Instead, I would suggest hardcoding the `#code=` prefix. This remains backwards compatible (unless people were making their own URLs with additional "query" parameters in the fragment identifier, which I don't think is very likely).

*Incidentally, try.ocamlpro.com has a similar bug: for example, the link https://try.ocamlpro.com/#rawcode/let%20x%20=%2012%20in%203%20+%20x turns the `+` into a space.*

<details>

<summary>Example/Reproduction:</summary>

Load this "code" into the playground editor:

<pre><br>| _ :: _ ->
| ->
</pre>

Note that this code snippet contains exactly one leading empty line (3 lines total + 1 trailing newline)!

Then, click the share button. The following URL is written to the browser's location/address bar: https://ocaml.org/play#code=CnwgXyA6OiBfIC0+CnwgLT4=

Now, try loading that URL in a new tab. Instead of loading the code snippet again, the default code is loaded into the editor.

For this code, trying the same loads all the code except the last `>`:

<pre><br>| _ :: _ ->
</pre>

(once again, leading newline)

</details>

just-max:
Fixes #1433. See that issue for the background.

This PR fixes the handling of `#code=`, and also adds a `#rawcode=` option (shamelessly inspired by https://try.ocaml.pro/).

~~As mentioned in the issue, the fragment is no longer parsed as a full key-value dictionary.~~

~~Changes:~~
~~+ correctly handle '+' in base64 encoded string~~
~~+ add #rawcode= as alternative to #code= (takes plain non-base64 text)~~
~~+ fragment identifer no longer parsed as full key-value dictionary~~

### DIFF
--- a/playground/src/main.ml
+++ b/playground/src/main.ml
@@ -109,22 +109,38 @@ let handle_output (o : Toplevel_api.exec_result) =
 module Codec = struct
   let ( let+ ) = Result.bind
 
+  let replace ~find ~replace s =
+    Jstr.cuts ~sep:find s |> Jstr.concat ~sep:replace
+
   let from_window () =
-    let uri = Window.location G.window |> Uri.fragment in
-    match Uri.Params.find (Jstr.v "code") (Uri.Params.of_jstr uri) with
-    | Some jstr ->
-        Result.to_option
-        @@ let+ dec = Base64.decode jstr in
-           let+ code = Base64.data_utf_8_to_jstr dec in
-           Ok (Jstr.to_string code)
-    | None -> None
+    let uri = Window.location G.window in
+    (* Bypass double encoding (dbuenzli/brr/issues/50) *)
+    let fragment = Jv.get (Uri.to_jv uri) "hash" |> Jv.to_jstr in
+    let params = Uri.Params.of_jstr (fragment |> Jstr.slice ~start:1) in
+    let from_code jstr =
+      (* previously, '+' was not URL-encoded and thus became ' ' *)
+      let unspace = replace ~find:(Jstr.v " ") ~replace:(Jstr.v "+") jstr in
+      Result.to_option
+      @@ let+ dec = Base64.decode unspace in
+         let+ code = Base64.data_utf_8_to_jstr dec in
+         Ok (Jstr.to_string code)
+    in
+    let from_rawcode jstr = Some (Jstr.to_string jstr) in
+    List.find_map
+      (fun (k, f) -> Option.bind (Uri.Params.find (Jstr.v k) params) f)
+      [ ("code", from_code); ("rawcode", from_rawcode) ]
 
   let to_window s =
     let data = Base64.data_utf_8_of_jstr s in
     let+ bin = Base64.encode data in
+    let query = Uri.Params.of_assoc [ (Jstr.v "code", bin) ] in
     let uri = Window.location G.window in
-    let+ s = Uri.with_uri ~fragment:(Jstr.concat [ Jstr.v "code="; bin ]) uri in
-    Ok (Window.set_location G.window s)
+    (* Bypass double encoding (dbuenzli/brr/issues/50); consider that without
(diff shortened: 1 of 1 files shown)


playground/src/main.ml
```diff
@@ -106,18 +106,30 @@ let handle_output (o : Toplevel_api.exec_result) =
   let out = El.div output_elements in
   El.append_children output [ out ]
 
+let strip_prefix ~prefix s =
+  let open Jstr in
+  if starts_with ~prefix s then
+    let l = length prefix in
+    Some (sub ~start:l ~len:(length s - l) s)
+  else None
```
just-max: Maybe there is a better place where this could live?

sabine:
@just-max thanks for going above and beyond by also opening a PR with a proposed fix to the reported issue!

I'll look at this next week when I had a moment to think if this solution is compatible with what is planned for the playground in the long term. :+1:

just-max:
Thanks for taking a look! Sorry for even more text on this, but this is subtler than I thought and my original fix might have been a bit rushed.

After thinking about it a bit more, the original approach of using the fragment identifier as query parameters is not as unidiomatic as I initially thought, and the [`Brr.Uri` docs](https://ocaml.org/p/brr/0.0.5/doc/Brr/Uri/Params/index.html) even suggest it directly.

So from what I can see there are two options:

1. continue as suggested in this PR, accepting that the 'syntax' of the fragment identifier differs from the standard parsing of query parameters (`+` is not a space).
2. adapt the encoder to percent-encode its output, turning `+` into `%2B`.

The first option smells of a bad time, and I would instead lean towards the second. It has the disadvantage of still being broken for existing share links (if they contain a `+`), but otherwise requires no changes to the decoder, so existing non-broken share links would still be non-broken. In fact, the output for non-broken share links would still be exactly the same.

The implementation of `#rawcode` is also not quite right, since it should percent-decode the code. This would also be fixed (modulo https://github.com/dbuenzli/brr/issues/50) by option 2.

Of course, option 1 would be possible but since you point to long-term planning it would make it more involved to add more parameters.

sabine:
I think you gave a really good assessment of the situation. :+1: 

Indeed, option 2 seems to be the most backwards- and forwards-compatible way to go.

just-max:
Ok, I've implemented the suggested fix. In the future, one might want to use URI-safe base 64 instead, but that could still be done in a backwards compatible way since there is no ambiguity between that and the current regular base 64.

sabine (approved):

sabine:
Thank you @just-max! Indeed, it's possible to improve on this later, but for now the bug is fixed. Maybe a future release of `brr` will provide the raw access that will simplify this code.

just-max:
Sounds good! Daniel has already published a new release of brr to address the issue, but as that might affect other uses of the library, it's probably best to make the switch in a separate PR anyway. I should have some time later this week :)

sabine:
Oh I see, that is an amazingly fast release! And sure, that will be super helpful!

PR #1450, opened 2023-07-25
Search Banner Component According to Learn Area Redesign

SaySayo:
Created the search banner according to the learn area redesign

Figma Design
Screenshot 2023-07-25 at 11 13 44

Current outcome
Screenshot 2023-07-24 at 19 11 45

There's a slight color difference but that will be fixed when there's a definite color scheme for the redesign


sabine:
We'll merge this at the point when we also implemented the search feature.

PR #1469, opened 2023-08-02
Make playground/gen_stdlib.sh mac compatible

gpetiot:
Got some issues on mac, let's see if the CI is okay with this change.

### DIFF
--- a/playground/Makefile
+++ b/playground/Makefile
@@ -1,17 +1,19 @@
 COMPILER=5.0.0
-SWITCH=playground
 
 .PHONY: create_switch
-create_switch: ## Create switch and pinned opam repo
-	opam switch create $(SWITCH) $(COMPILER) --no-install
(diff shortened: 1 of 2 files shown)


sabine (approved):
Works on my machine. However, with this patch, we can no longer use a local switch for the playground.

I was a bit confused at some point before about why the playground Makefile creates a named global switch, while the ocaml.org Makefile creates a local switch. This sounds like a recipe for potential trouble to me: When the playground uses a global switch, and ocaml.org uses a local one, can we be sure that the correct playground switch will be used and not accidentally the local switch of ocaml.org?

sabine:
After having working with a global switch for the playground and a local switch for ocaml.org some more, I think the sane solution here is to change the playground Makefile to use a local switch. Then things should naturally fall in place.

PR #1476, opened 2023-08-04
dune exec should point to the built executable

djavorszky:
Minor fix to the getting started docs; It corrects the `dune exec` command, as the executable gets built into `./_build/default/bin` instead of `./bin` folder.

### DIFF
--- a/data/tutorials/getting-started/gs_00_up_and_running.md
+++ b/data/tutorials/getting-started/gs_00_up_and_running.md
@@ -419,7 +419,7 @@ When we change our program, we can type `dune build` again to make a new
 executable. To run the program, we can use:
 
 ```
-$ dune exec ./bin/main.exe
+$ dune exec ./_build/default/bin/main.exe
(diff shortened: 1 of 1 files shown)


just-max:
Works fine for me:

```text
$ dune init project helloworld
Entering directory '/home/max/tmp/helloworld'
Success: initialized project component named helloworld
$ cd helloworld/
$ dune exec ./bin/main.exe
Hello, World!
```

In general `dune exec`-ing the path to the executable in the source folder does work.

djavorszky:
Hmm, you're right @just-max, I think what confused me is that there's no `./bin/main.exe` file at all after `dune build`, so tab-completion didn't work.

Thanks for checking it!

djavorszky:
Hmm, you're right @just-max, I think what confused me is that there's no `./bin/main.exe` file at all after `dune build`, so tab-completion didn't work.

Thanks for checking it!

PR #1477, opened 2023-08-04
Playground: upgrade brr to 0.0.6 to fix URI param hack

just-max:
Follow-up to https://github.com/ocaml/ocaml.org/pull/1434#issuecomment-1660357789 as promised :)

I didn't see any other uses of brr's URI module, so no further changes should be required.


just-max:
CC @sabine: There's no rush, but I wouldn't want it to go under either :>

sabine (approved):
Thanks @just-max, sorry, we were all mostly on holiday and definitely won't let this one slip. This is really nice!

PR #1490, opened 2023-08-12
Fix link to heading around UTop recommendation

bitsapien:
The link here does not navigate to the right place.

https://github.com/ocaml/ocaml.org/assets/3462960/bdd9edd3-44e1-43a5-b656-15f39d62a946

This happened because the title was changed at https://github.com/ocaml/ocaml.org/commit/6e62d71f601f35119f05722ffc3934bd5170fd45 and this link was not updated to reflect that.

This commit fixes it.


sabine:
Thanks @bitsapien, that's very helpful.

These links to sections within other documents aren't checked, and it's a bit of a pain currently to keep them in a good shape.

PR #1441, opened 2023-07-24
Learn overview api cards

SaySayo:
WIP: Created Component for Learn Overview download and standard library card.
Screenshot 2023-07-24 at 12 51 53

Screenshot 2023-07-24 at 12 53 17

PR #1491, opened 2023-08-12
Update new Install and Stdlib API cards to have consistent rounded borders

dmmulroy:
The new Install and Stdlib API cards that were introduced yesterday in #1441 at commit 5bdc2a8 look(ed) a bit off and "sharp" relative to the rest of the page. I updated their styles to include `rounded-xl` for visual consistency.

Before:
image

After:
image

### DIFF
--- a/src/ocamlorg_frontend/components/learn_components.eml
+++ b/src/ocamlorg_frontend/components/learn_components.eml
@@ -27,7 +27,7 @@ let command_line_icon class_ =
 let install_card 
 latest_version
 =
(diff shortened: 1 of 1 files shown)


sabine:
Hey @dmmulroy, we're working towards the design at https://www.figma.com/file/bL31M0WWTTdu2z9jAOX77J/UX%2FUI-OCaml-Learn-Area?type=design&node-id=442-1340&mode=design&t=ByhGOHlUdsTLmswo-0.

My impression is that the card background color is still darker than in Figma, however, I also agree that it looks friendlier with rounded borders.

PR #1516, opened 2023-09-04
Update about.eml

ShalokShalom:
Really love this page overall. 

And I thought it is currently underselling OCaml in significant ways.  This PR changes a few minor things:

I rephrased the type inference part, since it did sound before like only the variables would be inferred. 

One thing that I removed, is the mention of the Java compiler.  That one has received no commit for over 1 1/2 years, and it is unlikely, that it is picked up by anybody, so I changed it to the “actually in active development” ocaml-wasm project.

I also changed a few other things, and like to discuss some potential changes to this site, here in the comments. 

Specifically, I think the GC section forgets about the generational nature of the GC, and what impact that has on performance. Garbage collection generally is seens as a performance hog, and I like to make clear, that this isnt the case here. 

You dont need to go to Rust, to get proper system level performance ;) 
I would like to point out the dual nature of our GC, and how efficient it is.

### DIFF
--- a/src/ocamlorg_frontend/pages/about.eml
+++ b/src/ocamlorg_frontend/pages/about.eml
@@ -60,7 +60,7 @@ Layout.render
           </p>
           <ul>
             <li>
-              <strong><a href="https://dev.realworldocaml.org/garbage-collector.html">Generational garbage collection</a></strong> for automatic memory management, now a feature of almost every modern, high-level language.
+              <strong><a href="https://dev.realworldocaml.org/garbage-collector.html">Generational garbage collection</a></strong> for automatic memory management.
             </li>
             <li>
               <strong>First-class functions</strong> that can be passed around like ordinary values, as seen in
--- a/src/ocamlorg_frontend/pages/about.eml
+++ b/src/ocamlorg_frontend/pages/about.eml
@@ -81,13 +81,13 @@ Layout.render
               found in distributed, big-data frameworks like Hadoop.
             </li>
             <li>
-              <strong>Type inference</strong>, so you don’t need to annotate every single variable in your program with
-              its type. Instead, types are inferred based on how a value is used. Available in a limited form in C# with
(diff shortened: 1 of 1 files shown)


src/ocamlorg_frontend/pages/about.eml
```diff
@@ -81,13 +81,13 @@ Layout.render
               found in distributed, big-data frameworks like Hadoop.
             </li>
             <li>
-              <strong>Type inference</strong>, so you don’t need to annotate every single variable in your program with
-              its type. Instead, types are inferred based on how a value is used. Available in a limited form in C# with
+              <strong>Type inference</strong>, so you don’t need to annotate function parameters, return types, and variables.
```
sabine:
```suggestion
              <strong>Type inference</strong>, so you don’t need to annotate every function parameter, return type, and variable. 
```

sabine:
Thanks @ShalokShalom!

I think you're right that there should be proof and explanation added about the GC being efficient and reliable.

ShalokShalom:
Wonderful. 

How would that proof look like?

sabine:
I don't know, maybe there's benchmarks somewhere worth linking to. :thinking:

ShalokShalom:
What do you think of this page?
https://ocamlpro.com/blog/2020_03_23_in_depth_look_at_best_fit_gc/

ShalokShalom:
I did add that.

sabine:
I looked around a bit and the best I found is [this explanation of the Garbage Collector](https://dev.realworldocaml.org/garbage-collector.html), however, it doesn't go into details on performance.

There are some benchmarks for the compiler at https://sandmark.ocamllabs.io/, including some stats on the GC. I think that might be a useful resource here.

ShalokShalom:
So, I am fine how it is now. 

I added a second link, and I think it describes it splendid now. 
Thanks for working with you!

ShalokShalom:
Go ahead 👍🏻

sabine:
@ShalokShalom Thank you!

PR #998, opened 2023-03-21
Add section on debugging data races with TSan

jmid:
This PR adds an fourth section to the current debugging guide on using the new Thread Sanitizer mode to detect data races in multi-domain code.

We are currently experiencing issues with TSan back traces missing file names and line numbers, when compiled with older versions of GCC (and `clang` too) as illustrated by the included `<null>`s.
For this reason I am therefore marking the PR as 'draft'.
Once this issue has been settled, I'll remove the label and update the expected output to be `<null>`-free :grin: 

CC: @OlivierNicole @fabbing

PR #1524, opened 2023-09-13
Add 'Transitioning to multicore' guide

jmid:
This PR adds a guide on how to utilize TSan to make OCaml code multicore ready.
It uses a running bank example, adopted from @art-w.

The idea is to share a helpful workflow (or recipe) for making such a transition, that was succesfully used in the context of an experimental Irmin multicore port.

TO DO: 
- [x] Write a shitty first draft
- [x] Include a TSan backtrace with line numbers 
- [x] Input + proof reading from @art-w and @OlivierNicole
- [x] Elaborate on the motivation in the introduction
- [x] Adjust the text with a (hopefully forthcoming) `5.1.0+tsan` switch

### DIFF
--- a/data/tutorials/guides/1wf_04_multicore_ready.md
+++ b/data/tutorials/guides/1wf_04_multicore_ready.md
@@ -297,8 +297,35 @@ $ dune runtest
 This works well and TSan no longer complains, so our little library is
 ready for OCaml 5.x parallelism, hurrah!
 
+
+### Final remarks
+
+The programming pattern of 'always-having-to-do-something-at-the-end'
+that we encountered with the missing `Mutex.unlock` is a recurring
+one for which OCaml offers a dedicate function:
+```ocaml
+ Fun.protect : finally:(unit -> unit) -> (unit -> 'a) -> 'a
+```
+Using `Fun.protect`, we could have written our final fix as follows:
+```ocaml
+let transfer t ~src_acc ~dst_acc ~amount =
+  begin
+    if amount <= 0 then raise (Invalid_argument "Amount has to be positive");
+    if src_acc = dst_acc then raise (Invalid_argument "Cannot transfer to yourself");
+    Mutex.lock lock; (* addition *)
+    Fun.protect ~finally:(fun () -> Mutex.unlock lock) (* addition *)
+      (fun () ->
+         begin
+           if t.(src_acc) < amount
+           then raise (Invalid_argument "Not enough money on account");
+           t.(src_acc) <- t.(src_acc) - amount;
+           t.(dst_acc) <- t.(dst_acc) + amount;
+         end)
+  end
+```
+
 Admittedly, using a `Mutex` to ensure exclusive access may be a bit
 heavy if performance is a concern.  If this is the case, one option is
 to replace the underlying `array` with a lock-free data structure,
 such as the [`Hashtbl`
-from`Kcas_data`](https://ocaml-multicore.github.io/kcas/doc/kcas_data/Kcas_data/Hashtbl/index.html). 
+from`Kcas_data`](https://ocaml-multicore.github.io/kcas/doc/kcas_data/Kcas_data/Hashtbl/index.html).
--- a/data/tutorials/guides/1wf_04_multicore_ready.md
+++ b/data/tutorials/guides/1wf_04_multicore_ready.md
@@ -9,23 +9,23 @@ category: "Guides"
 
 # Transitioning to Multicore with ThreadSanitizer
 
-The 5.0 release brought multicore `Domain`-based parallelism to the
+The 5.0 release brought Multicore, `Domain`-based parallelism to the
 OCaml language. Parallel `Domain`s performing uncoordinated operations
 on shared mutable memory locations may however cause data races. Such
 issues will unfortunately not
 [(yet)](https://blog.janestreet.com/oxidizing-ocaml-parallelism/) be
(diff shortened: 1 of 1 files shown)


data/tutorials/guides/1wf_04_multicore_ready.md
```diff
@@ -0,0 +1,303 @@
...
+category: "Guides"
+---
+
+
+# Transitioning to Multicore with ThreadSanitizer
+
+The 5.0 release brought multicore `Domain`-based parallelism to the
+OCaml language, and with it new opportunities for data races. In this
+guide, we will study a step-wise workflow that utilizes the
+ThreadSanitizer (TSan) tool to help make your OCaml code 5.x ready.
```
sabine: I'm not sure the intro given here is specific enough. Maybe there's the possibility to elaborate what ThreadSanitizer is and why one would use it? It's probably not clear for a first-time reader what "5.x ready" means  and what ThreadSanitizer is at this point. There's just this little hint about data races.

sabine (commented):
Title ideas: "How to Write Thread-Safe Libraries", or "How to Test for Race Freedom". Reasoning: I don't think anyone wants to "transition to multicore", yet I am pretty sure people want to write libraries or code that can be used in a multicore context or test that their code does not have race conditions.

jmid:
The review from @art-w and @OlivierNicole was taken off line. Here are the two changes made to address it:
- f2d2a410 mentions `Fun.protect` as a cleaner way to ensure `Mutex.unlock` is always run
- 77c544f9 adds a word of warning and shares the trick of coordinating on an `Atomic.t`

I also included a few minor adjustments I spotted while proof-reading it myself.

jmid:
> Title ideas: "How to Write Thread-Safe Libraries", or "How to Test for Race Freedom". Reasoning: I don't think anyone wants to "transition to multicore", yet I am pretty sure people want to write libraries or code that can be used in a multicore context or test that their code does not have race conditions.

First of all, I completely agree that the title is not great! :100: :sweat_smile: 
However the guide is not intended to be about writing *new* thread-safe code and 
https://github.com/ocaml/ocaml.org/blob/main/data/tutorials/guides/1wf_01_debugging.md#detecting-a-data-race-with-thread-sanitizer from #998 already documents how to run TSan.

After many years of OCaml being single-core, a lot of well-testing single-`Domain` code exists.
This guide shares with the greater OCaml community a tried-and-true tool workflow that has already been used succesfully to prototype parallelism in Irmin.

data/tutorials/guides/1wf_04_multicore_ready.md
jmid: Point well taken. I'll elaborate some more :+1:

jmid: There's now a revised attempt in b3f8f25

jmid:
I've now addressed the previous comments, so this should be ready for review.

christinerose (approved):

data/tutorials/guides/1wf_04_multicore_ready.md
```diff
@@ -211,16 +211,16 @@ WARNING: ThreadSanitizer: data race (pid=26148)
 ```
 
 Notice we obtain a back trace of the two racing accesses, with
-- a write in one `Domain` coming from the array assignment in
-  `Bank.transfer` and
-- a read in another `Domain` coming from a call to
+- A write-in one `Domain`, coming from the array assignment in
```
jmid: This change alters the meaning AFAICS. The intension was to explain that there was a race between one domain performing a 'write operation' and another domain performing a 'read operation', hence
> - "A write in one `Domain`

jmid:
Thanks for the review and the fixes @christinerose. I spotted a minor nit, commented above.

data/tutorials/guides/1wf_04_multicore_ready.md
christinerose: Thanks for the clarification! I'll change it back.

christinerose:
> Thanks for the review and the fixes @christinerose. I spotted a minor nit, commented above.

My pleasure! 🙂

data/tutorials/guides/1wf_04_multicore_ready.md
christinerose: I removed the comma from both lines.

jmid: Eek, I meant the inserted dash in the `write in` to `write-in` change, but I realize now that that wasn't clear :sweat_smile:

christinerose: OH! So the commas are correct before "coming"?

christinerose: Actually, I like it without the commas there better. 
IRT the write in vs. write-in, I see how I misread it. Thanks for the catch!

christinerose (approved):

data/tutorials/guides/1wf_04_multicore_ready.md
jmid: I'm fine with the latest version, but also admit that I'm probably not the right one to make the call about the comma in the light of all the other ones I missed... :smile:

tmattio:
Thanks @jmid and @christinerose, new documentation content is always greatly appreciated and this one seems particularly important!

I agree with @sabine that we could try to find a better title, however, I wouldn't want this to block the merge, the current title is more than acceptable. A follow-up PR that changes it if we find anything better is welcome!

Merging :)

PR #1525, opened 2023-09-14
Remove Is OCaml learning yet page

tmattio:
Reverts part of #1226.

I'm largely in favour of pushing early versions of pages to iterate on them, but the learning page is essentially a skeleton at this stage, I don't think it provides any value and I'd be embarrassed to have someone land on it to compare it with the fantastic https://www.arewewebyet.org/.

I would have preferred to have other pairs of eyes on the "Is OCaml web yet?" page before merging it, but if @sabine you consider it ready to at least provide some value and accurate information, I'm fine with having it merged.


sabine:
If the point is to highlight where there is need for people to contribute, I think the learning page is pretty okay while the web page gives a more comprehensive image. It shows that there are some foundations and bindings (which is honestly great - I was surprised there is so much) but that one cannot expect to build entire pipelines in OCaml yet.

The problem with these pages in general is that we don't really have capacity to do them well. Creating better filtering for opam packages by tags would be more impactful.

I'm fine with removing the learning page, though, if you feel it's not giving people something useful.

PR #1529, opened 2023-09-18
Tutorial on Mutability and Imperative Programming

### LINKED ISSUE #1462: Mutable State / Imperative Programming
Goal: Reader understands the different capabilities of OCaml that allow them to work with mutable state
- references
- mutable record fields
- isolating mutability / pure functions to reason about, mutable state where it's helpful

Isolate mutability in tutorials as it should be in well-architectured project. 

PR: https://github.com/ocaml/ocaml.org/pull/1529

cuihtlauac:
Mutable Data: references, records and arrays/bytes
Imperative
Exceptions
Howto mix styles without getting into trouble
Refunctionalization techniques


Tchou:
Hi,
some low-level remarks. 
```
# #show ref;;
```
why not `#show_type ref` ? This would only show the type definition (which is what is interesting) and would allow you to remove the awkward reference to binding with C (also I haven't checked but I think makemutable is not a C function but a compiler primitive).


"Note: This example will not run in the REPL, since the function Array.truncate is not defined."

Why not use `Array.sub` then ? it only adds a 0 parameter and the code can actually be tested (and replace the paragraph about the fake truncate by on explaining in passing what `Array.sub` does).

Not too sure about the "Analytics" example. I would avoid depending on code that is left to the reader to imagine. I think you could achieve the same effect by logging something in a or on stderr. I would maybe start this part by a more general remark saying that in OCaml, side-effects do not occur in types (whether they are modification of mutable state or exceptions). That is a strength (allows you to use local side effects, or functions that behaves as-if they were pure but more efficient) but also a problem since, when the side effect is actually part of the spec, one has to rely on documentation to be aware of that (e.g. a function that mutates its argument).

As for side effects in arguments I would generalize a bit saying that this also apply to constructor argument and that code such as:
```
let r = ref 0 in
((incr r; !r), (incr r; !r))
```
may return different values depending on the back-end, optimization level, version of the compiler etc…

cuihtlauac:
> why not #show_type ref ? This would only show the type definition (which is what is interesting) and would allow you to remove the awkward reference to binding with C (also I haven't checked but I think makemutable is not a C function but a compiler primitive).

Fixed

> "Note: This example will not run in the REPL, since the function Array.truncate is not defined."

If we show code which is bad but somehow working, it will be copied and used. truncate is way to work around that (analytics undefined too).

> Not too sure about the "Analytics" example. I would avoid depending on code that is left to the reader to imagine. I think you could achieve the same effect by logging something in a or on stderr. I would maybe start this part by a more general remark saying that in OCaml, side-effects do not occur in types (whether they are modification of mutable state or exceptions). That is a strength (allows you to use local side effects, or functions that behaves as-if they were pure but more efficient) but also a problem since, when the side effect is actually part of the spec, one has to rely on documentation to be aware of that (e.g. a function that mutates its argument).

I like the idea you suggest about adding text. What I like about analytics is the discrepancy between copying an array and opening a network connection. But you are right, we should not assume everybody knows about analytics and provide some details

> let r = ref 0 in
((incr r; !r), (incr r; !r))

Excellent example!

data/tutorials/language/0it_06_imperative.md
```diff
@@ -0,0 +1,654 @@
...
+**`let … in`**
+
+```ocaml
+# let () = print_string "This is" in print_endline " really Disco!";;
+This is really Disco!
+- : unit = ()
+```
+
+Using the `let … in` construct means two things:
+* Names may be bound (in the example, no name is bound since the )
```
christinerose: This sentence is incomplete

data/tutorials/language/0it_06_imperative.md
```diff
@@ -0,0 +1,654 @@
...
+  print_endline "Hello,";
+  print_endline "world!";
+  42;;
+Hello,
+world!
+- : int = 42
+```
+In this example, the first two expressions are `print_endline` function calls, which produce side effects (printing to the console), and the last expression is simply the integer `42`, which becomes the value of the entire sequence. The `;` operator is used to separate these expressions.
+
+**Remark** Even thought it's called the sequence operator, the semicolon is not truly an operator because it is not a function of type `unit -> 'a -> 'a`. It is rather a _construct_ of the language. That allows terminating sequences with a semicolon, which is ignored.
```
christinerose: Please check my revision here to ensure it's still accurate. 🙂

data/tutorials/language/0it_06_imperative.md
```diff
@@ -0,0 +1,654 @@
...
+
+When you use the `downto` keyword (instead of the `to` keyword), the counter decreases on every iteration of the loop.
+
+For example, `for` loops are convenient for iterating over and modifying arrays:
+```ocaml
+# let sum = ref 0 in
+  for i = 0 to Array.length j - 1 do sum := !sum + j.(i) done;
+  !sum;;
+- : int = 42
+```
```
christinerose: Please check this rearrangement. I did it per @cuihtlauac's instruction in Slack DM, but I might have misunderstood.

Also, I get the result of `35` here, not `42`.

data/tutorials/language/0it_06_imperative.md
```diff
@@ -0,0 +1,654 @@
...
+
+**Note:** Here is how to do the same thing using an iterator function:
+```ocaml
+# let sum = ref 0 in Array.iter (fun i -> sum := !sum + i) a; !sum;;
+- : int = 42
```
christinerose: Also get `35` here instead of `42` because I do above. 🙂

data/tutorials/language/0it_06_imperative.md
```diff
@@ -0,0 +1,654 @@
...
+    while true do
+      let c = get_char () in
+      if c = '\027' then raise Exit;
+      print_char c;
+      flush stdout
+    done
+  with Exit -> ();;
+```
+
+This `while` loop echoes characters typed on the keyboard. When the ASCII `Escape` character is read, the `Exit` exception is thrown, returning `- : unit = ()`.
```
christinerose: Please review my revisions here, @cuihtlauac, to ensure it's still accurate.

data/tutorials/language/0it_06_imperative.md
```diff
@@ -0,0 +1,654 @@
...
+      | c when c >= ' ' -> record_char !state c
+      | _ -> ("", !state) in
+    print_string s;
+    state := new_state;
+    flush stdout;
+    loop state;;
+val loop : char list ref -> 'a = <fun>
+
+# let state = ref [] in try loop state with Exit -> state_to_string !state;;
+```
```
christinerose, line 483: Since this doesn't return anything but just leaves a space for more input, should we mention that or clarify it's necessary to return twice to get back to the prompt?

data/tutorials/language/0it_06_imperative.md
```diff
@@ -0,0 +1,654 @@
...
+
+This example illustrates the following:
+- The functions `record_char` and `remove_char` neither update the state nor produce side effects. Instead, they each return a pair of values consisting of a string to print and the next state, `new_state`.
+- I/O and state-update side effects happen inside the `loop` function.
```
christinerose, line 489: Is my revision of "state-update side effects" correct? Before it said "state update side effects happen..." 

I stumbled on the sentence wondering if it was:
- states update side effects
- `state` updates side effects
- `state updates` side effects
Can we clarify this? ... same in li.488

data/tutorials/language/0it_06_imperative.md
```diff
@@ -0,0 +1,654 @@
...
+- The state is passed as a parameter to the `loop` function.
+
+This is a possible way to handle an application-wide state. As in the [Function-Encapsulated Mutability](#good-function-encapsulated-mutability) example, state-aware code is contained in a narrow scope; the rest of the code is purely functional.
```
christinerose, line 492: Same question as li.483.

data/tutorials/language/0it_06_imperative.md
```diff
@@ -0,0 +1,654 @@
...
+
+<!-- FIXME: reference CS3110 memoization documented on ocaml.org after it is merged -->
+
+### Good: Functional by Default
+
+By default, OCaml programs should be written in a mostly functional style. This constitutes trying to avoid side effects where possible and relying on immutable data instead of mutable state.
+
+It is possible to use an imperative programming style without losing the benefits of type and memory safety. However, it doesn't usually make sense to only program in an imperative style. Not using functional programming idioms at all would result in non-idiomatic OCaml code.
+
+Most existing modules provide an interface meant to be used in functional way. Some would require the development and maintenance of [wrapper libraries](https://en.wikipedia.org/wiki/Wrapper_library) to be used in an imperative setting and such use would in many cases be inefficient.
```
christinerose: If it's inefficient, should it be mentioned here?

data/tutorials/language/0it_06_imperative.md
```diff
@@ -0,0 +1,654 @@
...
+    (Array.truncate k_len k, Array.truncate m_len m);;
+Error: Unbound value Array.truncate
+```
+
+**Note:** This example will not run in the REPL, since the function `Array.truncate` is not defined.
+
+To understand why this is bad code, assume that the function `Array.truncate` has type `int -> 'a array -> 'a array`. It behaves such that `Array.truncate 3 [5; 6; 7; 8; 9]` returns `[5; 6; 7]`, and the returned array physically corresponds to the 3 first cells of the input array.
+
+The type of `partition` would be `('a -> bool) -> 'a array -> 'a array * 'a array`, and it could be documented as:
+> `partition p k` returns a pair of arrays `(m, n)` where `m` is an array containing all the elements of `k` that satisfy the predicate `p`, and `n` is an array containing the elements of `k` that do not satisfy `p`. The order of the elements from the input array is preserved.
```
christinerose, line 584: `n` (formerly `c`) isn't in the above example. Is this intended? Will the reader understand?

christinerose (commented):
Questions found in my review

data/tutorials/language/0it_06_imperative.md
cuihtlauac: I need to fix offline

christinerose (approved):

data/tutorials/language/0it_06_imperative.md
```diff
@@ -399,14 +394,21 @@ The type of a `while` loop expression is `unit`.
 2
 3
 4
+5
 - : unit = ()
 ```
 
+Here:
+- `!i < 5` is the condition
```
christinerose:
```suggestion
- `!i <= 5` is the condition.
```
Should this have the `=` too?

data/tutorials/language/0it_06_imperative.md
```diff
@@ -399,14 +394,21 @@ The type of a `while` loop expression is `unit`.
...
+- The expression ` Printf.printf "%i\n" !i; i := !i + 1;` is the body of the loop
+
+The iteration executes the body as long as the condition remains true.
```
christinerose: executes the expression, perhaps? Have we used "body" in this context before?

data/tutorials/language/0it_06_imperative.md
```diff
@@ -421,7 +423,7 @@ The following example uses the `get_char` function we defined earlier (in the se
   with Exit -> ();;
 ```
 
-This `while` loop echoes characters typed on the keyboard. When the ASCII `Escape` character is read, the `Exit` exception is thrown, returning `- : unit = ()`.
+This `while` loop echoes characters typed on the keyboard. When the ASCII `Escape` character is read, the `Exit` exception is thrown, which terminates the iteration and triggers the display of the reply from the REPL: `- : unit = ()`.
```
christinerose:
```suggestion
This `while` loop echoes characters typed on the keyboard. When the ASCII `Escape` character is read, the `Exit` exception is thrown, which terminates the iteration and displays the REPL reply: `- : unit = ()`.
```

data/tutorials/language/0it_06_imperative.md
```diff
@@ -606,24 +608,28 @@ Consider this code:
 Error: Unbound module Analytics
 ```
 
-**Note:** This code will not run because there is no module called `Analytics`.
+**Note:** This code will not run because there is no module called `Analytics`. [Analytics](https://en.wikipedia.org/wiki/Web_analytics) are remote monitoring libraries.
 
 A module called `Array` is defined; it shadows and includes the [`Stdlib.Array`](/api/Array.html) module. See the [Module Inclusion](docs/modules#module-inclusion) part of the [Modules](docs/modules) tutorial for details about this pattern.
 
-To understand why this code is bad, assume that `Analytics.collect` is a function that makes a network connection to transmit data to another server.
+To understand why this code is bad, figure out that `Analytics.`collect` is a function that makes a network connection to transmit data to a remote server.
```
christinerose: Is `Analytics.collect` not correct? Should it be `Analytics` `connect` or Analytics `collect`?

christinerose (commented):
A few questions

data/tutorials/language/0it_06_imperative.md
cuihtlauac: body expression

data/tutorials/language/0it_06_imperative.md
cuihtlauac: No back tick in the middle

data/tutorials/language/0it_06_imperative.md
cuihtlauac: We need to say something like:
> Type and edit any single line text, and then press return to get back to the repl

christinerose (approved):

sabine (approved):

sabine:
Anything left to do before me merge this?

PR #1531, opened 2023-09-20
Tutorial on Polymorphic Variants

### LINKED ISSUE #1533: Polymorphic Variants
ChatGPT hallucinates so much on that topic, it means it never processed anything about it. Which means a tutorial is highly needed

PR: https://github.com/ocaml/ocaml.org/pull/1531

cuihtlauac:
Polymorphic variants deserve a tutorial supplementing documentation from the manual and gathering information spread in the mailing list, discuss and blog posts. 

Here is the outline:

1. Introduction and historical notes
2. A First Example
3. Exact, Closed and Open Variants
4. Subtyping
5. Named Polymorphic Variants
6. Advanced: Alised, Parametrized and Recursive Polymorphic Variabnts
7. Advanced: Combining Polymorphic Variants and Simple Variants
8. Uses-Cases
9. Drawbacks
10. When to Use Polymorphic Variants
11. Conclusions & References


christinerose:
Apologies, this commit ^^ is through li. 65.

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+
+However, they are type-checked using different algorithms, which results in a different programming experience. The relationship between value and type (written with the colon symbol `:`) is changed with polymorphic variants. Usually, values are thought of as inhabitants of the type, which is regarded as a set-like thing. Rather, polymorphic variant values should be considered as pieces of data that several functions can accept. Polymorphic variants types are a way to express compatibility relationships between those functions. The approach in this tutorial is to build sense from experience using features of polymorphic variants.
+
+**Prerequisites**: This is an intermediate-level tutorial. It is required to have completed tutorials on [Functions and Values](/docs/functions-and-values), [Basic Data Types](/docs/basic-data-types), and [Lists](/docs/lists) to begin this one.
+
+### Origin and Context
```
leostera, line 21: Love this bits of trivia <3 but I'd put them all the way at the _end_ of the tutorial. Almost an appendix for those that are interested in deepening the context in which polyvars where created.

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+- Maintain existing code that is using polymorphic variants
+- Sort out polymorphic variant types and errors
+  - Add type annotation needed to resolve polymorphic variant type-checking issues
+  - Add or remove catch-all patterns in polymorphic variant pattern matching
+  - Insert or remove type casts
+- Refactor between simple and polymorphic variants
+- Choose to use polymorphic variants when really needed
+-->
+
+### A Note on Simple Variants and Polymorphism
```
leostera, line 50: Looking at the learning goals for this page, I'm not sure if this clarification helps me write/use/assess polyvars at all.

Feels like a "nice to know" but not a "must know" 🤔 – maybe it would fit better near `## When to Use Polymorphic Variant`?

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+
+However, they are type-checked using different algorithms, which results in a different programming experience. The relationship between value and type (written with the colon symbol `:`) is changed with polymorphic variants. Usually, values are thought of as inhabitants of the type, which is regarded as a set-like thing. Rather, polymorphic variant values should be considered as pieces of data that several functions can accept. Polymorphic variants types are a way to express compatibility relationships between those functions. The approach in this tutorial is to build sense from experience using features of polymorphic variants.
+
+**Prerequisites**: This is an intermediate-level tutorial. It is required to have completed tutorials on [Functions and Values](/docs/functions-and-values), [Basic Data Types](/docs/basic-data-types), and [Lists](/docs/lists) to begin this one.
```
leostera, line 19: Love prerequisites! ✨ I think I'd phrase it around skills/concepts rather than completed tutorials, and it could be placed right below the main title or in a dedicated section so the very first thing I notice is whether I need to learn something _before_ digging into this tutorial.

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+category: "Language"
+---
+
+# Polymorphic Variants
+
+## Introduction
+
+This tutorial teaches you how to use polymorphic variants. This includes starting to use them, maintaining a project already using them, deciding when to use them or not, and balancing their unique benefits against their drawbacks.
+
+Product types and data types such as `option` and `list` are variants and polymorphic. In this tutorial, they are called _simple variants_ to distinguish them from the _polymorphic variants_ presented here. Simple variants and polymorphic variants are close siblings. Their values are both introduced using labels that may carry data. Both can be recursive and have type parameters. By the way, don't trust a [LLM](https://en.wikipedia.org/wiki/Large_language_model) chatbot if it tells you polymorphic variants are dynamically typed; it is a hallucination. Like simple variants, polymorphic variants are type-checked statically.
```
leostera, line 15: I think rather than preambling around LLMs or whether other kinds of types are also polymorphic (but meaning something different), I'd just immediately start building an intuition:
1. _what is_ a polymorphic variant,
2. _why_ would I want to use them,
3. and _how_ to identify them.

```markdown
A polymorphic variant is a more flexible kind of variant type that does not need
to be defined upfront, and is useful for working with groups of constructors
(we call them _tags_) that can grow or shrink.

When they are used to create values, you'll see them prefixed with a backtick:
`` `hello_world ``.

When they are used in type-signatures, you'll see them wrapped in square brackets:
 `` [ `hello_world ] ``.
```

The rest of the tutorial can then dig into these facets and make sure we go from an intuition to a proper understanding, including things like its origins or why the relationship between value and type may be different.

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+
+The identifiers `list`, `option`, and others are _type operators_. Just like functions, they take parameters. Although these parameters are not values, they are types. Their results aren't values but types, too.
+
+Simple variants are polymorphic, but not in the same sense as polymorphic variants.
+- Simple variants have [parametric polymorphism](https://en.wikipedia.org/wiki/Parametric_polymorphism)
+- Polymorphic variants have a form of [structural polymorphism](https://en.wikipedia.org/wiki/Structural_type_system)
+
+## A First Example
+
+### Creating Polymorphic Variant Types From Pattern Matching
```
leostera, line 62: An even smaller example we could use here is a list. I think it could cement the intuition that polyvars are closer to the data itself than to defined types:

```ocaml
# let my_favorite_fruits = [ `banana ];;
val my_favorite_fruits : [> `banana ] list

# let my_favorite_fruits = [ `banana; `apple ];;
val my_favorite_fruits : [> `apple | `banana ] list

# let my_favorite_fruits = [ `banana; `apple; `kiwi ];;
val my_favorite_fruits : [> `apple | `banana | `kiwi ] list
```

You could easily go from there to using polyvars that carry data:

```ocaml
# let my_favorite_fruits = [ `banana; `apple `green; `orange `blood; ];;
val my_favorite_fruits :
  [> `apple of [> `green ] | `banana | `orange of [> `blood ] ] list
```

leostera: This is the path I'd take when explaining them, starting from the smallest possible example and growing one step at a time:

1. `` `banana `` is of type `` [> `banana ] ``

2. if you put it together with other things that are their own type (`` `apple ``) then they form a larger type (`` [> `apple | `banana ] ``)

3. if you use an argument like a polyvar in a function, then the input type gets formed by its usage:

```ocaml
# let f `hello = `world ;;
val f : [< `hello ] -> [> `world ] = <fun>
```

4. then we can explain `>` and `<`, the _exact_ polyvar type, and what combinations like `` [> `name of string < `age of int ] `` mean

6. then we can introduce sharing

```ocaml
# let say_hi (`name name) = "hello, " ^ name;;
val say_hi : [< `name of string ] -> string

# let say_bye (`name name) = "goodbye, " ^ name;;
val say_bye : [< `name of string ] -> string
```

7. and upfront definitions to make the reuse cleaner

```ocaml
type name = [ `name of string ]
```

8. and now that you have a polyvar defined upfront, you can introduce composition

```ocaml
type name = [ `name of string ]
type age = [ `age of int ]
type attributes = [ name | age | `other of (string, string) ]
```

9. profit!

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+
+The expression ``[< `Broccoli | `Fruit of string ]`` plays the role of a type. However, it does not represent a single type, it represents three different types.
+- The type that only has `` `Broccoli`` as an inhabitant, its translation into a simple variant is `type t0 = Broccoli`
+- The type that only has `` `Fruit`` as an inhabitant, its translation into a simple variant is `type t1 = Fruit of string`
+- Type type that has both `` `Broccoli`` and `` `Fruit`` inhabitants, its translation into a simple variant is `type t2 = Broccoli | Fruit of string`
+
+Note each of the above translations into simple variants is correct. However, entering them as-is into the environment would lead to constructor shadowing (unless type annotation is used at pattern or expression level, see section on [Shared Constructors](#Shared-Constructors)).
+
+This also illustrates the other striking feature of polymorphic variants: values can be attached to several types. For instance, the tag `` `Broccoli`` inhabits ``[ `Broccoli ]`` and``[ `Broccoli | `Fruit of String ]``, but it also inhabits any type that contains it.
```
leostera, line 79: > "but it also inhabits any type that contains it."

This is going to raise eyebrows. Inhabits alone is a term that confused the hell out of me for a long time.

I think if we build the intuition that `` `Broccoli `` is a value that belongs to the type `` [ > `Broccoli ] `` then it would make more sense that anywhere that we see `` `Broccoli `` in the type, then `` `Broccoli `` is a valid value.

Of course there's the case where `` `Broccoli `` belongs to `` [ > `Apple ] `` because of `>` but I'd leave `>` until we talk about polyvars as function inputs/outputs.

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+
+What is displayed by the type-checker, for instance ``[< `Broccoli | `Fruit of string ]``, isn't a single type. It is a type expression that designates a constrained set of types. For instance, all the types that are defined by a group of tags containing either`` `Broccoli`` or `` `Fruit of string`` and nothing more. This is the meaning of the `<` sign in this type expression. This is a bit similar to what happens with `'a list`, which isn't a single type either, but a type expression that designates the set of `list` types of something, i.e. the set of types where `'a` has been replaced by some other type. It is also meant to indicate that the exact types are subsets of the indicated ones (this will be explained in the section on [Subtyping](#Subtyping-of-Polymorphic-Variants)).
```
leostera, line 81: > "isn't a single type"

Not entirely sure if the idea that "in OCaml every value has a single type" (principality) has been hammered in by now in the tutorials, but if it has then this would be weird to read.

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+- polymorphic variant: type expressions displayed by the OCaml type-checker such as ``[< `Broccoli | `Fruit of string ]``
+
+### Tag Sharing
+
+Here is another function using pattern matching on a polymorphic variant.
+```ocaml
+# let g = function `Edible name -> name | `Broccoli -> "Brassica oleracea";;
+val g : [< `Broccoli | `Edible of string ] -> string = <fun>
+
+# f `Broccoli;;
```
leostera, line 96: Consider that if I jump into this section, I don't know where `f` came from or what it is. Keeping it self-contained should be okay since these examples are super small.

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+
+# g `Broccoli;;
+- : string = "Brassica oleracea"
+```
+
+Both `f` and `g` accept the`` `Broccoli`` tag as input because they both have code for it. They do not have the same domain because `f` also accepts `` `Fruit of string`` whilst `g` also accepts `` `Edible of string``. The domains of `f` and `g` express this. The tag `` `Broccoli`` satisfies both the constraints of the domain of `f`: ``[< `Broccoli | `Fruit of string ]`` and the constraints of the domain of `g`: ``[< `Broccoli | `Fruit of string ]``. That type is ``[ `Broccoli ]``. The value defined by a tag belongs to several types, similarly, a tag accepting functions belongs to several types.
+
+Polymorphic variant tags are meant to be used as stand-alone values, wherever it makes sense. As long as used consistently, with a single implicit type per tag, the type checker will accept any combination of them in pattern-matching expressions.
+
+### Static Type-Checking
```
leostera, line 107: This section seems unnecessary if the remarks of the LLM hallucinating are removed.

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+
+However, they are type-checked using different algorithms, which results in a different programming experience. The relationship between value and type (written with the colon symbol `:`) is changed with polymorphic variants. Usually, values are thought of as inhabitants of the type, which is regarded as a set-like thing. Rather, polymorphic variant values should be considered as pieces of data that several functions can accept. Polymorphic variants types are a way to express compatibility relationships between those functions. The approach in this tutorial is to build sense from experience using features of polymorphic variants.
+
```
Octachron, line 18: I don't understand what this paragraph is trying to explain.

First, since typecheckers are not the subject on the article, bringing type-checker algorithms only encourage the idea that you can only understand OCaml language if you are a type system expert .

Second, polymorphic variants are using the same type-checking algorithm that the rest of the language.

Third, `None` already belongs to multiple type, there is nothing special at this level with respect to the row type variable of objects or polymorphic variable.

I advocate to completely remove this paragraph.

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+**Prerequisites**: This is an intermediate-level tutorial. It is required to have completed tutorials on [Functions and Values](/docs/functions-and-values), [Basic Data Types](/docs/basic-data-types), and [Lists](/docs/lists) to begin this one.
+
+### Origin and Context
+
+Polymorphic variants originate from Jacques Garrigue's work on Objective Label, which was [first published in 1996](https://caml.inria.fr/pub/old_caml_site/caml-list-ar/0533.html). It became part of standard Objective Caml with [release 3.0](https://caml.inria.fr/distrib/ocaml-3.00/) in 2000, along with labelled and optional function arguments. They were introduced to give more precise types in [LablTk](https://garrigue.github.io/labltk/).
```
Octachron, line 23: This paragraph should probably be a footnote except if the expected audience of the tutorial are student in a Programming Language class.

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+
+The core type system of OCaml follows a [_nominal_](https://en.wikipedia.org/wiki/Nominal_type_system) discipline. Variants must be explicitly declared before being used. The typing discipline used for polymorphic variants and classes is different, as it is [_structural_](https://en.wikipedia.org/wiki/Structural_type_system).
+
+In the nominal approach of typing, types are first defined; later, when the type of an expression is inferred, three outcomes are possible:
+1. If a matching type is found, it becomes the inferred type.
+1. If any type can be applied, a type parameter is created.
+1. If typing inconsistencies are found, an error is raised.
+
+This is very similar to solving an equation in mathematics. Equations accept either zero, exactly one, several, or infinitely many numbers as solutions. Nominal type-checking finds that either zero, exactly one, or any type can be used in an expression.
+
```
Octachron, line 33: This is misleading, consider `None`.

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+In the structural approach of typing, type definitions are optional, so they can be omitted. Type-checking an expression constructs a data structure that represents the types that are compatible with it. These data structures are displayed as type expressions sharing a resemblance with simple variants.
+
```
Octachron, line 35: The simplest value with a structural in OCaml is probably `let id x = x` which doesn't fit this paragraph.

I am also still confused why we are speaking of a ` data structure that represent types`? Why are the type system implementation details the focus of this paragraph?

data/tutorials/language/0lg_10_polymorphic_variants.md
Octachron: I agree, if this section belongs to this tutorial, it is not at its place.

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+TODO: Expand this section
+
+> There is one more downside: the runtime cost. A value `Pair (x, y)` occupies 3 words in memory, while a value `` `Pair (x, y)`` occupies 6 words.
+
+Guillaume Melquiond
+
+## When to Use Polymorphic Variant
+
+Only using polymorphic variants is over-engineering, it should be avoided. Polymorphic variants aren't an extension of simple variants. From the data perspective, polymorphic variants and simple variants are equivalent. They both are sum types in the algebraic data types sense, both with parametric polymorphism and recursion. Back quote isn't a difference that matters. The only meaningful difference is the algorithm. Therefore, deciding when to use polymorphic variants boils down to another question:
```
leostera, line 673: > "Only using polymorphic variants is over-engineering,"

I also don't recommend people to _only use polyvars_ but I think we can do better and explain _why_.

For example, the type errors can get VERY out of hand.

Also not sure why reiterating the whole "not an extension, but equivalent" is relevant.

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+
+# (List.hd u) (`Fruit "Pitahaya");;
+Error: This expression has type [> `Fruit of string ]
+       but an expression was expected of type [< `Broccoli ]
+       The second variant type does not allow tag(s) `Fruit
+```
+
+Function `f` accepts tags `` `Broccoli`` and `` `Fruit`` whilst `g` accepts `` `Broccoli`` and `` `Edible``. But if `f` and `g` are stored in a list, they must have the same type. That forces their domain to be restricted to a common subtype. Although `f` can handle the tag `` `Fruit``, it no longer accepts that parameter when `f` is extracted from the list.
+
+### Weaken Type-Checking and Harder Debugging
```
leostera, line 615: When you write "weaken type-checking" are you referring to having both a `Gray` and `Grey` tags? I assumed this was meant to be an example of an accident/typo, but I don't think its fair to call it "weakened type-checking".

Like the compiler isn't going to mistake `Grey` for `Gray`.

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+* Many variant declarations look alike
+* Equivalent constructor duplicated in several variants
+* Variant declared for a single-purpose
+* Variant declaration feeling made up or artificial
+
+Having difficulties finding names for simple variants or constructors may be smell for one of the above. When such discomfort is felt, polymorphic variants may not be the solution, but they can be considered. Keep in mind they may ease some parts but at the expense of understandability, precision and performance. Hopefully, very soon, LLM bots will be able to refactor variants of some sort into the other. This will ease experimentation.
+
+To conclude, remember a simple variant that naturally encodes some data should remain a simple variant.
+
+## Conclusion
```
leostera, line 688: Instead of a Conclusion i'd have a section that summarizes what was learned in this tutorial:

* We learned how to create new polymorphic variant tags
* We learned how to to constraint them 
* We learned how to reuse tags
* We learned how to use polymorphic variants for flexible error handling
* etc

data/tutorials/language/0lg_10_polymorphic_variants.md
cuihtlauac: Agreed. Your feedback on this is consistent with the discuss thread; it will go at the end of the document.

data/tutorials/language/0lg_10_polymorphic_variants.md
cuihtlauac: Your suggestion makes sense too. I'll do that.

data/tutorials/language/0lg_10_polymorphic_variants.md
cuihtlauac: It is hard to reach a consensus on this. Some want to have prerequisites as the very first thing. Others prefer to start with the learning goals. Ideally, both could be turned into metadata. That would allow more rendering options.

data/tutorials/language/0lg_10_polymorphic_variants.md
cuihtlauac: That seems consistent with the feedback in the discuss thread. A differently toned introduction is demanded. Let's write it.

data/tutorials/language/0lg_10_polymorphic_variants.md
cuihtlauac: Agreed. The terminology I've used here is clumsy. This is needs to be rewritten.

data/tutorials/language/0lg_10_polymorphic_variants.md
cuihtlauac: You're right again. This needs to be said differently.

data/tutorials/language/0lg_10_polymorphic_variants.md
cuihtlauac: You are right. The document does not state explicitly examples are meant to be played in order and jumping into a section does not work. This needs to be clarified.

data/tutorials/language/0lg_10_polymorphic_variants.md
cuihtlauac: Fair enough, it is just a remark.

data/tutorials/language/0lg_10_polymorphic_variants.md
cuihtlauac: I like the idea of an introduction that would list the practical features of polymorphic variants, from a developer's perspective. I agree it would be friendlier. What do you think of this ?
- Optional declaration
- Tag sharing
- Hash patterns

That creates a different, and probably better, learning track. It implies a large refactoring of the rest of the text, but I agree it's worth the effort.

I'm concerned about the relationship with simple variants. Some newcomers are puzzled by the resemblance-difference with polymorphic variants. Plenty of questions arise in relation to that.

data/tutorials/language/0lg_10_polymorphic_variants.md
cuihtlauac: Agree. This is consistent with feedback in the discuss thread.

data/tutorials/language/0lg_10_polymorphic_variants.md
cuihtlauac: You're right. If we downplay the type-checking algorithm story, this will likely disappear.

data/tutorials/language/0lg_10_polymorphic_variants.md
cuihtlauac: Agreed.

data/tutorials/language/0lg_10_polymorphic_variants.md
cuihtlauac: Do you have an example code triggering such type errors? I may be useful. 

"not an extension, but equivalent": Yes, and that discussion will by downsized and what's left will move to the end of the document.

data/tutorials/language/0lg_10_polymorphic_variants.md
cuihtlauac: > Like the compiler isn't going to mistake Grey for Gray.

But it would if it had been simple variants. That's the point. But I didn't say it explicitly, so that's on me.

data/tutorials/language/0lg_10_polymorphic_variants.md
cuihtlauac: Yes, that's simpler.

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+* ``[ `Night ] list`` is a subtype of ``[ `Night | `Day ] list``
+
+Note that these types are covariant because of the way their type parameters appear in the type of their constructors. This is detailed in [Chapter 11, Section 8.1](https://v2.ocaml.org/manual/typedecl.html#ss:typedefs) of the OCaml Manual. This has nothing to do with being predefined types.
+
+### Function Codomains are Covariant
+
+The function type is covariant on codomains. Casting a function is allowed if the target codomain is larger.
+```ocaml
+# let upcast_dom (f : int -> [ `Night ]) = (x :> int -> [ `Night | `Day ]);;
+val f : (int -> [ `Night ]) -> int -> [ `Night | `Day ] = <fun>
```
reynir:
```suggestion
# let upcast_dom (f : int -> [ `Night ]) = (f :> int -> [ `Night | `Day ]);;
val upcast_dom : (int -> [ `Night ]) -> int -> [ `Night | `Day ] = <fun>
```

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+* On function type: ``int -> [ `Night ]`` is a subtype of ``int -> [ `Night | `Day ]``
+
+Adding tags to a polymorphic variant codomain of a function is harmless. Extending a function's codomain is pretending it can return something that is never returned. It is a false promise, and the precision of the type is reduced, but it is safe, no unexpected data will ever be returned by the function.
+
+### Function Domains are Contravariant
+
+The function type is _contravariant_ on domains. Casting a function is allowed if the target domain is smaller.
+```ocaml
+# let upcast_cod (f : [ `Night | `Day ] -> int) = (x :> [ `Night ] -> int);;
+val f : ([ `Night | `Day ] -> int) -> [ `Night ] -> int = <fun>
```
reynir:
```suggestion
# let upcast_cod (f : [ `Night | `Day ] -> int) = (f :> [ `Night ] -> int);;
val upcast_cod : ([ `Night | `Day ] -> int) -> [ `Night ] -> int = <fun>
```
:up: :fish:

reynir (changes requested):
I found some typos in the OCaml code.

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+### Error Handling Using the `result` Type
+
+The [Error Handling](/docs/error-handling) guide details possible ways to handle errors in OCaml. Among various mechanisms, the `result` type, when used as a monad, provides a powerful means to handle errors. Refer to the guide and documentation on this type to learn how to use it. In this section, we discuss why using polymorphic variants to carry error values can be beneficial.
+
+Let's consider this exception-raising code:
+```ocaml
+# let f_exn m n =
+    let open List in
+    let u = init m Fun.id |> map (fun n -> n * n) in
+    nth u n;;
```
reynir: I don't think the brevity is worth losing explicitness.
```suggestion
    let u = List.init m Fun.id |> List.map (fun n -> n * n) in
    List.nth u n;;
```

data/tutorials/language/0lg_10_polymorphic_variants.md
```diff
@@ -0,0 +1,713 @@
...
+# type nth_error = Too_short of int * int | Negative_index of int;;
+type nth_error = Too_short of int * int | Negative_index of int
+
+# let nth u i = try Ok (List.nth u i) with
+    | Invalid_argument _ -> Error (Negative_index i)
+    | Failure _ -> Error (Too_short (List.length u, i))
+val nth : 'a list -> int -> ('a, nth_error) result = <fun>
+
+# let f_res m n =
+    let* u = init m Fun.id in
```
reynir, line 534: I think `let*` is not defined in this snippet, and when defining it in the snippet below it may be worth linking to some text about the `let*` syntax(?)

cuihtlauac: You are right, the definition of let should move up. Links are also needed you are right

data/tutorials/language/0lg_10_polymorphic_variants.md
reynir: The variable `x` still needs to be renamed to `f` I believe.

tmattio:
I'm closing this as stalled. I've heard that a rewrite was in progress somewhere, but in the meantime, I'll close this. Feel free to re-open or open a new PR if you resume work on this.

PR #1534, opened 2023-09-21
Refactor manual redirections and add 5.1 redirections

### LINKED ISSUE #1528: Documentation PDF Format unaccessicle (page 404)

Octachron:
This PR adds the redirection for the binary archives of the manual to the list of the redirections to `v2.ocaml.org`.

I might have refactored a bit the redirection list along the way.

close #1528

### DIFF
--- a/src/ocamlorg_web/lib/redirection.ml
+++ b/src/ocamlorg_web/lib/redirection.ml
@@ -2,255 +2,126 @@ open Ocamlorg
 
 let fwd_v2 target = (target, Url.v2 ^ target)
 
+
+let pp_ocaml_version ppf (mj,mn)=
+  if mj >= 5 then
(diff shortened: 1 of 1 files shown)


src/ocamlorg_web/lib/redirection.ml
cuihtlauac, line 1: This is great. I wish there were a tool to check/test those kind of changes. It's just two lists of strings. We should be able to check equality between them.

cuihtlauac (approved):

Octachron:
There was a mistake on the introduction of ".adoc", but I have now tested the change with https://gist.github.com/Octachron/63bcff0d9c7511183c965d15d1aff092 .

tmattio:
Thanks @Octachron!

PR #1541, opened 2023-09-23
Tab breadcrumbs for mobile view on learn area

SaySayo:
Adding a breadcrumb as replacement for the navigation tab on mobile screens 
Screenshot 2023-09-22 at 21 01 10


sabine (commented):
1. Click area needs to be bigger. Instead of putting the vertical padding on the `nav` element, it should be on the `a` element:

Screenshot from 2023-10-17 14-03-03



2. Font in the design is `font-bold` instead of `font-semibold` on the sections.
Screenshot from 2023-10-17 14-05-15
Screenshot 2023-10-17 at 14-12-27 Learn OCaml

3. Separator between "Learn" and section seems to have a different stroke width in the design and seems smaller in the design.

4. I would aim for having the same height in total as the tabs, for consistency and so it is simple to click or tap.

src/ocamlorg_frontend/css/styles.css
```diff
@@ -256,3 +256,12 @@ summary::-webkit-details-marker {
   display: none;
 }
 
+select {
+  -webkit-appearance: none;
+  -moz-appearance: none;
+  appearance: none;
+}
```
sabine: use Tailwind class instead

src/ocamlorg_frontend/layouts/learn_layout.eml
```diff
@@ -7,14 +7,58 @@ type section =
   | Exercises
   | Books
 
+
```
sabine:
```suggestion
```

src/ocamlorg_frontend/css/partials/package_breadcrumbs.css
```diff
@@ -5,4 +5,13 @@
 
 .package-breadcrumbs .breadcrumbs-tag {
   @apply text-sm px-3 py-0.5 rounded font-bold text-white inline-block;
+}
+
+.learn-breadcrumbs li:not(:last-child)::after {
+  @apply text-lighter;
+  content: " >";
+}
```
sabine: is that necessary? I thought the separator was an svg icon.

SaySayo: My bad. Thats an error, forgot it in there while still using the `>` as seperator.

sabine (approved):

PR #1548, opened 2023-09-28
Update exercise 23 solution and problem statement

### LINKED ISSUE #1492: Update List exercise 23: Extract a Given Number of Randomly Selected Elements From a List
## Overview

There seems to be an inconsistency in the recommended solution and the example answer in the exercise.
The recommended solution returns `["a"; "b"; "c"]` while the expected answer is `["g"; "d"; "a"]`.

I was wondering if it is due to the `Random` module since the problem description states that using `Random` without `Random.self_init` should allow for reproducible results. However, the closest I got to the expected answer was by applying the following steps:
1. use `Random.init 0` to explicity set the seed to 0
2. Add two `List.rev` in `extract` and on the result of `aux`.

After these steps i was able to get  `["g"; "d"; "b"]` instead of  `["g"; "d"; "a"]`, so I am not too sure if there is a bug in the solution/answer or if it an issue with my setup with Ocaml + opam that is causing this.

My setup
- ocaml version: 5.0.0
- opam version: 2.1.5
- dune version: 3.10.0

** Current exercise description, solution and modified solution below.

### Current Exercise Description/Solution

image
image

### Modified solution

This modified version of the solution returns close to the expected answer;  `["g"; "d"; "b"]` instead of  `["g"; "d"; "a"]`

```ocaml
let rand_select list n =
  Random.init 0; (* explicitly set seed to zero *)
  let rec extract acc n = function
    | [] -> raise Not_found
    | h :: t -> if n = 0 then h, List.rev acc @ t else extract (h :: acc) (n - 1) t (* reverse accumulator *)
  in
  let extract_rand list len = extract [] (Random.int len) list in
  let rec aux n acc list len =
    if n = 0
    then acc
    else (
      let picked, rest = extract_rand list len in
      aux (n - 1) (picked :: acc) rest (len - 1))
  in
  let len = List.length list in
  aux (min n len) [] list len |> List.rev (* reverse accumulated result from tail recursive helper function *)
;;
```

calaquende:
Update exercise description and solution to use `Random.init 0` at the start of the function for reproducibility.

Closes #1492

### DIFF
--- a/data/problems/023_extract_random.md
+++ b/data/problems/023_extract_random.md
@@ -10,9 +10,10 @@ tags: [ "list" ]
 
 ```ocaml
 # let rand_select list n =
+    Random.init 0;
     let rec extract acc n = function
       | [] -> raise Not_found
-      | h :: t -> if n = 0 then (h, acc @ t) else extract (h :: acc) (n - 1) t
+      | h :: t -> if n = 0 then (h, List.rev acc @ t) else extract (h :: acc) (n - 1) t
     in
     let extract_rand list len =
       extract [] (Random.int len) list
@@ -23,17 +24,17 @@ tags: [ "list" ]
         aux (n - 1) (picked :: acc) rest (len - 1)
     in
     let len = List.length list in
-      aux (min n len) [] list len;;
+      aux (min n len) [] list len |> List.rev;;
 val rand_select : 'a list -> int -> 'a list = <fun>
 ```
 
 # Statement
 
 The selected items shall be returned in a list. We use the `Random`
-module but do not initialize it with `Random.self_init` for
-reproducibility.
+module but and initialize it with `Random.init 0` at the start of 
+the function for reproducibility.
 
 ```ocaml
 # rand_select ["a"; "b"; "c"; "d"; "e"; "f"; "g"; "h"] 3;;
-- : string list = ["g"; "d"; "a"]
+- : string list = ["g"; "d"; "b"]
 ```
--- a/data/problems/023_extract_random.md
+++ b/data/problems/023_extract_random.md
@@ -31,7 +31,7 @@ val rand_select : 'a list -> int -> 'a list = <fun>
 # Statement
 
 The selected items shall be returned in a list. We use the `Random`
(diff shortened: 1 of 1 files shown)


christinerose (approved):

data/problems/023_extract_random.md
```diff
@@ -23,17 +24,17 @@ tags: [ "list" ]
...
 val rand_select : 'a list -> int -> 'a list = <fun>
 ```
 
 # Statement
 
 The selected items shall be returned in a list. We use the `Random`
-module but do not initialize it with `Random.self_init` for
-reproducibility.
+module but and initialise it with `Random.init 0` at the start of 
+the function for reproducibility.
```
tmattio:
```suggestion
the function for reproducibility and validate the solution. To make the function truly random, however,
one should remove the call to `Random.init 0`
```

data/problems/023_extract_random.md
```diff
@@ -10,9 +10,10 @@ tags: [ "list" ]
 
 ```ocaml
 # let rand_select list n =
+    Random.init 0;
     let rec extract acc n = function
       | [] -> raise Not_found
-      | h :: t -> if n = 0 then (h, acc @ t) else extract (h :: acc) (n - 1) t
+      | h :: t -> if n = 0 then (h, List.rev acc @ t) else extract (h :: acc) (n - 1) t
```
tmattio: Since the elements are extracted randomly, it doesn't make much sense to re-order the list (it's an additional operation, but we don't care about the order).
If the solution doesn't match the returned value of the function with `Random.init 0`, let's change the solution instead :)

tmattio (commented):
Thanks for the PR, @marcfyk! I've left a couple of comments

data/problems/023_extract_random.md
the-ides-of-marc: I removed the unnecessary reverse operations :)

tmattio:
Thanks @marcfyk, much appreciated! Don't hesitate to open other issues if you notice anything else with the documentation/exercises

Issue #1566, opened 2023-10-04
Highlight Selected Tag in Changelog


sabine:
Note: this has been selected as an Outreachy issue. To be assigned, you must have completed the instructions in https://github.com/ocaml/ocaml.org/issues/1553.

Please only request to be assigned to this issue if you are applying to the Dark Mode Outreachy project.

# Task

On the OCaml Changelog https://ocaml.org/changelog it is possible to filter the changelog by tag. However, there is no visual highlighting of the tag on the filtered page.

Here's what it should look like:

![Screenshot from 2023-10-04 10-02-00](https://github.com/ocaml/ocaml.org/assets/6594573/a8428127-8331-41e0-86ce-a5bdf304ce48)

What you need to do:

In the template `changelog.eml`, modify the Tailwind styles of the tag. If the tag being rendered is the same as the currently selected one, use `bg-primary-700` and `text-white`.

Additionally, you need to make sure that the hover and focus styles are reasonable by adding appropriate Tailwind CSS classes.

kate-bagenzo:
Hi - Can I be assigned to this? (My Ocaml.org setup + 1st dream example are posted in #1553)

sabine:
Sure, you're now assigned to this issue @kate-bagenzo

SaySayo:
For clarification purposes, are you applying for the dark mode or the GUI project? @kate-bagenzo

kate-bagenzo:
The dark mode project. @SaySayo

sabine:
Completed by #1567

PR #1567, opened 2023-10-04
Highlight selected changelog tag

kate-bagenzo:
Solution for #1566 to add orange highlight & hover style to the currently selected tag in the changelog.


sabine (approved):
Looks good to me. Nice and subtle hover effect.

PR #1580, opened 2023-10-05
New nav layout according to Figma design

### LINKED ISSUE #1569: Implement New Navigation Structure on the Exercises Page
This is an Outreachy Medium difficulty issue. You must have completed a simple issue (marked with label "outreachy") first, to be eligible for being assigned to this one.

Please only request to be assigned this issue if you're applying to the Dark Mode Outreachy project.

# Task

The Navigation in the left sidebar of the exercises at https://ocaml.org/problems needs to be improved:

|Current Exercises Page|Planned Design|
|-|-|
|Screenshot 2023-10-04 at 16-46-24 Exercises|Screenshot from 2023-10-04 16-47-33|

Here is [the Figma document for reference regarding the styles](https://www.figma.com/file/bL31M0WWTTdu2z9jAOX77J/UX%2FUI-OCaml-Learn-Area?type=design&node-id=1539%3A3356&mode=design&t=lTPcjIhaySpBMtWd-1).

How to solve this issue:

1. Modify the template `problems.eml` (may be renamed to `exercises.eml` by https://github.com/ocaml/ocaml.org/issues/1558): Instead of rendering a list of links in the sidebar, use `summary` and `details` HTML elements to group the exercises into different sections.
2. Do not render empty sections. (When the page is filtered by difficulty, some of the sections may be empty.)
3. Assign Tailwind CSS styles to match the Figma design. Try to make use of existing variables in `tailwind.config.js` / match what you see in other parts of the site.

SwizzCoder:
closes #1569 
cc @sabine @SaySayo @gpetiot 
This PR implements the new Figma design for navigation layout on the exercises page.

## Screenshots
When page is loaded
image

When a category is opened
image

### DIFF
--- a/src/ocamlorg_frontend/components/learn_components.eml
+++ b/src/ocamlorg_frontend/components/learn_components.eml
@@ -192,4 +192,5 @@ let lang_manual_banner =
       <a href="<%s Url.manual %>" class="w-full md:w-auto mt-5 py-2 px-24 bg-default text-sm rounded-sm items-center text-center text-primary-700 sm:px-20 sm:text-md font-medium leading-7 md:text-lg">Take Me There</a>
     </div>
     <div class="min-w-[200px] min-h-[200px] overflow-hidden rounded-full bg-cover bg-center bg-no-repeat md:min-w-[240px] md:min-h-[240px]" style="background-image: url('../img/home/ocaml_camel.png')">
+  </div>
 
--- a/src/ocamlorg_frontend/pages/learn.eml
+++ b/src/ocamlorg_frontend/pages/learn.eml
@@ -74,9 +74,9 @@ Learn_layout.single_column_layout
     <%s! Learn_components.left_books_content_block books ~title:"Recommended For Beginners" ~link:{href = Url.books; title = "See More Books"} %>
     <%s! Learn_components.right_books_content_block intermediate_books ~title:"For Intermediate Users" ~link:{href = Url.books; title = "See More Books"} %>
   </div>
-    
+  <div>
   <%s! Learn_components.lang_manual_banner %>
-
+  </div>
   <div class="container-fluid">
     <div class="mt-10 lg:mt-20 border-b border-gray-200 pb-10">
       <h3 class="font-bold">Papers</h3>


SwizzCoder:
@sabine @SaySayo @gpetiot 
I have a few questions. 
- From the Figma design, we have a hover effect when the cursor is crossing over any of the exercise topics. I believe the background color applied is `bg-primary-300`. However, applying this color to the element class has no effect. I think the `talwind.config.js` file is not properly reading these `RGBA` colors. Other colors with this problem too will be the `primary-100` and `primary-200` colors. Given that this task is about the nav structure is it okay to create a separate issue to handle the colors, given that the cause of it not working may not be related to this task.
- The Figma design features a drop caret on each category. I wish to ask if the platform already uses a library for icons. I tried to look in the code but couldn;t find any icon library like fontawesome etc. I didn't want to include a whole icon library just for the caret so what are your thoughts? Is it okay to just use unicodes? Which may be problematic as unicodes are not well parsed on some systems and may leave the website appearing inconsistent on different systems.

sabine:
@SwizzCoder 

- Hover background looks to be `bg-primary-100` (compared to the selected color on the sidebar of the existing ocaml.org documentation pages). Cleaning up the colors and adjusting the `tailwind.config.js` values for them is part of the Dark Mode Outreachy project, so for now we go with the closest color we already have on the page.
- We use heroicons.com and there is a template `icons.eml` that contains all the icons. If the icon in question exists already, it should be used, if not, it should be taken from Heroicons and put there (if available on Heroicons). In rare cases, the Figma design will use an icon that's not on Heroicons, but then the icon should be taken from the Figma document.

SwizzCoder:
Hello @sabine @SaySayo 
I think this PR is ready for review

SwizzCoder:
I had a funny issue with the chevron icons. For instance `Icons.chevron_right` will render the down icon and `Icons.chevron_down` will render the left icon.

src/ocamlorg_frontend/layouts/learn_layout.eml
```diff
@@ -36,26 +36,25 @@ let sidebar_link
...
-    </a>
-  | false ->
-    <a class="font-semibold text-sm text-lighter hover:bg-gray-100 block py-2 px-3" href="<%s href %>">
-      <%s title %>
-    </a>
+    <div class="py-2 px-3 <%s if current then "bg-primary-300 text-primary-600 border-l-2 border-primary-600" else "hover:bg-primary-100 hover:border-l-2 hover:border-primary-600" %>">
+      <a class="focus:text-primary-600 active:text-primary-600 text-sm <%s if current then "text-primary-600 font-semibold" else "font-medium" %>"" href="<%s href %>">
+        <%s title %>
+      </a>
+    </div>
```
sabine: A general rule in UI/UX is this:

When you have a clickable area (e.g. an `a` element), it needs to have a minimum size for people to comfortably click or tap on.

So, here, introducing a wrapping `div` with the padding is not a good idea. It is better to have the styles directly on the `a` element, to ensure that the clickable area is as large as the user expects.

src/ocamlorg_frontend/layouts/learn_layout.eml
```diff
@@ -36,26 +36,25 @@ let sidebar_link
...
-    <%s! extra_html %>
-  </div>
+  <details class="space-y-2 flex flex-col">
+    <summary class="flex gap-4 justify-between text-sm font-semibold px-3 py-2 uppercase focus:text-primary-600 active:text-primary-600 hover:text-primary-300">
+      <%s title %>
+      <%s! Icons.chevron_down "h-6 w-6" %>
+    </summary>
+    <span class="ml-4"><%s! link_html %></span>
+    <span class="ml-4"><%s! extra_html %></span>
+  </details>
```
sabine: Note that this also affects all the other pages on the Learn layout. The only page that should have this new navigation is the Exercises page. This code needs to be moved to `problems.eml` so that the old `sidebar_link_group` can remain as is.

sabine (commented):
Sorry it took so long to review. Here's some feedback / answers:

1. The rotation on the icon comes from `styles.css`
```
details summary svg {
  transform: rotate(90deg);
}
details[open] summary svg {
  transform: rotate(-90deg);
}
```
I think you are right, this is unintuitive and not obvious. It would be better to change this to a 180deg rotation when open. (Requires replacing icons on all other existing instances of `summary` to make sure they're in the expected orientation. It is okay to include in this patch.)

2. Looks good overall, but there is an alignment problem here:
Screenshot 2023-10-10 at 10-54-29 Exercises

Note how in the design, the sections align to the left and have different font styles:
272617525-2c84b311-32a3-4aff-89b0-02dc91c27316

3. It looks to me like we need special styling for the opened `summary` on the Exercises page to achieve the orange highlighting when opened.

4. Gray line on the left is missing. For the orange line the design suggests for the active item: it's fine to omit that at this point.

src/ocamlorg_frontend/layouts/learn_layout.eml
SwizzCoder: Thank you so much for this comment. I have learned so much about these HTML elements just investigating this and seeing how to make it work, but also better. Turns out, we don't even need the DIV element at all.

PR #1538, opened 2023-09-22
Add changelog for Merlin 4.11

voodoos:

PR #1586, opened 2023-10-06
Correct date of merlin 4.11 release

sabine:
I'm assuming the 2023-09-22 from the file name is the correct date for Merlin 4.11 release as added in https://github.com/ocaml/ocaml.org/pull/1538

@voodoos is this correct?


voodoos:
Yes it is ! Sorry for missing it :-)

PR #1571, opened 2023-10-04
Add references for profiling on macOS

leostera:
Hello! 👋🏽 

I was going through this tutorial and realized there's no pointers for how to do profiling on macOS. Figured I'd chime in with a small pointer to the Instrument's app and few screenshots to show how it works.

PR #1596, opened 2023-10-08
Vendor macOS Instruments screenshot

tmattio:
I merged #1571 a tad too quickly and noticed that it was using GitHub URLs for the screenshots instead of adding them to the repository.

I've moved the image to the repository and resized it so we don't embed 1.3M in the executable.
I'm also updating the docs to use only one screenshot: the two were very similar and took a lot of space, so it seemed better to only show the tool once.

cc @leostera


leostera:
Hey no problem :) I took 2 because I wanted to show how the tool looks like while you record and how it looks like when you dig through your own stack traces. I reckon that its not immediately obvious skimming through but if you are actively looking at the images you'd see that the table below changes in content from "random programs in Leandro's computer" to "stack traces sorted by time of an actual OCaml program". 

I'll be doing a writeup on this on the Practical OCaml blog soon tho, so no pressure if you have some length-metric to keep under.

sabine (approved):

christinerose (approved):

PR #1598, opened 2023-10-08
Add @avsm as a codeowner of the policies and Platform roadmap

tmattio:
To make clear that changes to the OCaml.org policies and Platform roadmap need to be approved by the [Owner/Delegate](https://ocaml.org/policies/governance#a.-owner-and-delegates).

@avsm let me know if there are any other pages that should be added here.


avsm (approved):
WFM!

PR #1627, opened 2023-10-12
Add changelog for dune 3.11.1

emillon:


christinerose (approved):

sabine:
@emillon If merging this was not the right call, let me know. I took the linked issue to mean that we should merge this.

tmattio:
Would be great to always have an announcement for the changelogs. At the moment this one is empty: https://ocaml.org/changelog?t=dune. We also now have an RSS feed for the Changelog (which was requested by the community), so we're sending these to RSS clients, would be nice not to send empty news :)

If the release is minor, just a small paragraph is enough.

emillon:
Yes thanks @sabine !

@tmattio I agree that pushing these to RSS or some kind of fediverse thing automatically would be great. Concretely if that could remove the manual step of pushing to discuss it would be even better :) (I think it supports "following" some fediverse events)

PR #1628, opened 2023-10-12
Turn link from 'Unknown docs' to be a span

davesnx:
Looks like detecting if there's Docs or not could disable the button. Not entirely sure if it's desired, but instead of opening an issue, I changed the template directly.

Screenshot 2023-10-12 at 13 48 48


sabine (approved):
Thanks @davesnx, that seems valid to me. Let's merge it.

PR #1643, opened 2023-10-15
Fix constructor name for variant examples

iamargentum:
This PR fixes constructor name used in an example showcasing pattern matching with variants.
Previous example uses the constructor `Error_code` for type `http_response`, so this makes it consistent across examples.


cuihtlauac (approved):
Thanks, this is indeed a bug

PR #1689, opened 2023-10-20
Fix function types in tour of ocaml pattern matching continued section

MeanderingProgrammer:
## Details

Minor issue in types on the following getting started section: https://ocaml.org/docs/tour-of-ocaml#pattern-matching-contd.

Sorry if these kinds of problems are meant to go through a different process, if this is more of a bug I can create an issue instead, or whatever is most appropriate.

## Testing

Validated types are correct in ocaml REPL:

```
# let g x =
  if x = "foo" then 1
  else if x = "bar" then 2
  else if x = "baz" then 3
  else if x = "qux" then 4
  else 0;;
val g : string -> int = <fun>
```

```
# let g' x = match x with
    | "foo" -> 1
    | "bar" -> 2
    | "baz" -> 3
    | "qux" -> 4
    | _ -> 0;;
val g' : string -> int = <fun>
```

### DIFF
--- a/data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
+++ b/data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
@@ -400,7 +400,7 @@ In this other example, the same comparison is made, using `if … then … else 
   else if x = "baz" then 3
   else if x = "qux" then 4
   else 0;;
-val g : int -> string = <fun>
+val g : string -> int = <fun>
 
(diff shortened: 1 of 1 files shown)


cuihtlauac:
Thanks, @MeanderingProgrammer; this is unquestionably a bug. Thanks for reporting it and proposing a fix.

cuihtlauac (approved):

PR #1608, opened 2023-10-10
New Design for Package Search Dropdown

AndroGenius-codes:
Resolves https://github.com/ocaml/ocaml.org/issues/1602.

@sabine  This is a first draft for the search feature. i am going to attach a video of what i have done so far and I need some clarification and thoughts on the design.
1)do i need to change the size of the search bar in the header or should i leave it the way it is.
2) increasing the size of the search bar to match the width of the drop down will make the header too busy so i think we should either reduce the width of the dropdown to match the width of the search bar or let the width of the drop down be wider than that of the search bar.
3)I need your thoughts on the search bar on the package page. what do i need to change.
[search feature.webm](https://github.com/ocaml/ocaml.org/assets/70571088/7483d5cb-1eee-47a0-8d5a-98ca93358ea9)

N/B: i had to use vanilla css because tailwind was a bit limiting, i could not back changes to hover background, i would appreciate a little hint on how to overcome this.

PR #1691, opened 2023-10-20
Package Search UX Improvements

leostera:
Hello ocaml fam 👋🏼 – i had a moment and i figured i'd pitch y'all some ux improvements that would make package search a little more comfortable :)

Let me know what you think and ofc happy to answer any questions and do any tweaks.

### Fix dynamic search results wobblynes

https://github.com/ocaml/ocaml.org/assets/854222/52cb2dfb-e1ec-4b9e-928f-85e27e5255d1

### Make Search Bars the first focused or focusable element

In the top-level search pages, we'd like the Search bar to be already in focus by the time we land. Most people go to packages to search for packages, and this is the default behavior in many package registries (crates.io, rubygems.org, cocoapods.org, pkg.go.dev, etc).

https://github.com/ocaml/ocaml.org/assets/854222/a68b1663-8cee-4beb-877a-f48f05c27a33

### When to show the Search Bar

First, the package search is great, but it doesn't need to be shown twice in any page, so I've made it configurable and removed it from the core search pages.

Screenshot 2023-10-20 at 07 03 39
Screenshot 2023-10-20 at 07 03 57

It is however still available on the slide-over menu: 
Screenshot 2023-10-20 at 07 04 25

### OpenSearch integration

I've also added a small manifest file to allow browsers to show a `Search OCaml Packages` action when someone is typing `ocaml.org` on their browser. This is in the same spirit as Crates.io and other websites use, to make it easier to jump to the search.

### DIFF
--- a/src/ocamlorg_frontend/pages/packages_search.eml
+++ b/src/ocamlorg_frontend/pages/packages_search.eml
@@ -119,3 +119,6 @@ let render ~total ~search ~page ~number_of_pages (packages : Package.package lis
     </div>
   </div>
 </div>
+<script>
+  document.querySelector("main input[type=search]").focus()
+</script>
--- a/asset/opensearch.xml
+++ b/asset/opensearch.xml
@@ -1,0 +1,6 @@
+<OpenSearchDescription xmlns="http://a9.com/-/spec/opensearch/1.1/">
+<ShortName>OCaml Packages</ShortName>
+<Description>Search for packages on ocaml.org</Description>
+<Image type="image/png">https://ocaml.org/asset/logo.svg</Image>
+<Url type="text/html" method="get" template="https://ocaml.org/packages/search?q={searchTerms}"/>
+</OpenSearchDescription>
--- a/src/ocamlorg_frontend/components/header.eml
+++ b/src/ocamlorg_frontend/components/header.eml
@@ -30,8 +30,8 @@ in
 <header
   class="h-20 flex items-center bg-mild-contrast dark:bg-dark-mild-contrast"
   x-data="{ open: false }">
-  <nav class="container-fluid wide header flex justify-between items-center">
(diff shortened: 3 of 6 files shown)


src/ocamlorg_frontend/pages/packages_search.eml
```diff
@@ -1,117 +1,121 @@
...
-            </div>
-          </li>
-          <% ); %>
-        </ol>
-        <% ); %>
-      </div>
-    </div>
-  </div>
-</div>
+let render ~total ~search (packages : Package.package list) = Layout.render ~title:"OCaml Packages · Search Result"
```
leostera: Ah, sorry about this, i'll redo this one.

sabine:
Hey, here's what I love:

1. search input being selected right when you go on /packages - that's seems nice and comfortable. May be a little confusing to people on screen readers, but since the main point of the page may be to advertise package search, and since that's the first interactive element on the page, it probably makes sense to autofocus on the input in this case.
2. Today I learned about https://indieweb.org/OpenSearch, and it looks like a really great idea to add an Opensearch manifest!

Other remarks:
1. removing search bar on /packages and search results page: probably not now. It appears likely that we'll also have to provide documentation search (as a secondary search target) in the main search bar in the navigation and the simpler we keep things here, the easier it's going to be to change.
2. width bug should be fixed by https://github.com/ocaml/ocaml.org/pull/1608 (see staging.ocaml.org)

sabine:
k, here's what we're merging after the CI completes:

1. Move package search input in main nav bar earlier in the DOM, so it gets tab-focus earlier.
2. tabindex="1" on package docs page for in-package search
3. adding an OpenSearch manifest so people can search OCaml packages more easily from browser search bar
4. autofocus search input on packages_search results page, since it's 99% likely that you either want to search again or just go through the search results

Thanks so much @leostera!

sabine (approved):

PR #1694, opened 2023-10-20
An update to the odoc 2.3.0 release announcement

panglesd:
This PR is made of two commits:

- One which adds some links to example for the "render source code" feature. I think it is a nice improvement, since otherwise, there is no way to preview this exciting feature!
- The second one which puts the feature highlights first. I think the roadmap is well explained elsewhere in the site, and that the importance should be given to the changes to odoc!

Feel free to revert the second commit if the current order feels more appropriate to you!


tmattio:
Thanks @panglesd!
@christinerose I'm merging now because this is "only" syncing the changelog with the post that was published on Discuss.

PR #1695, opened 2023-10-20
Use local open

cuihtlauac:
This makes the code easier to read and illustrates the usefulness
of `let open ... in`

### DIFF
--- a/data/tutorials/language/0lg_08_operators.md
+++ b/data/tutorials/language/0lg_08_operators.md
@@ -182,13 +182,14 @@ The `doi_parts` function attempts to extract the registrant and identifier parts
 val ( let* ) : 'a option -> ('a -> 'b option) -> 'b option = <fun>
 
 # let doi_parts s =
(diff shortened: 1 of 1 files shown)


sabine (approved):
Looks fine to me.

Also makes me think of that we need to show people what local opening with brackets is.... 

```
     let identifier = String.(sub s (slash + 1) (length s - slash - 1))
```
Doesn't belong here, but where does this little piece of info belong? It can be quite relevant for people to be able to read existing code.

cuihtlauac:
I believe two things are needed
- Explain different "local open" means avaiable
- Spray several such examples, there is one in operators, we need more

cuihtlauac:
We have another example in "1st program", in the dream example. However, we've removed the explanation on Dream. I believe we've cut too much too fast here.

It is also used in the same file with the module defined in the tutorial, see ll 298

It is used in the error handling and module tutorials as well.

PR #1724, opened 2023-10-30
Packages: Changed the version select dropdown

### LINKED ISSUE #1612: Style the Package Version Select
This is an Outreachy Medium difficulty issue. You must have completed a simple issue (marked with label "outreachy") first, to be eligible for being assigned to this one.

This issue is for the Dark Mode project. Only request to be assigned this issue if you are applying for the Outreachy Dark Mode project.

# Task

On the package documentation pages, there is a select element to switch between different versions of the same package. This should be replaced by a dropdown like this:

|before|after|
|-|-|
|Screenshot from 2023-10-11 11-24-56|Screenshot from 2023-10-11 11-24-45|

[Here is a Figma Document with the styles](https://www.figma.com/file/J0w4b9exaHhVpvabKoGqTn/Design-Github-Issues?type=design&node-id=47%3A274&mode=design&t=q6e3n0Ul6MatGJqY-1).

Constraints:
* Dropdown implementation must be accessible (WCAG).

How to solve this issue:
1. Determine a reasonable, accessible implementation for the version switcher dropdown. AlpineJS may be used.
2. Modify `package_breadcrumbs.eml`.

ademolaomosanya:
Resolves: https://github.com/ocaml/ocaml.org/issues/1612

This is a PR draft on changing the dropdown for version selection on packages page to continually contribute to the Ocaml task.
Used **`Alpine.js`** for the dropdown

changes made are shown in the screenshot below

- Before
Screenshot 2023-10-30 at 16 12 28

- After
Screenshot 2023-10-30 at 16 11 49

Changes yet to be made

- Links to each version are yet to be updated

- Web Content Accessibility Guidelines (WCAG) have not been checked yet.

### DIFF
--- a/src/ocamlorg_frontend/components/package_breadcrumbs.eml
+++ b/src/ocamlorg_frontend/components/package_breadcrumbs.eml
@@ -38,24 +38,56 @@ let render_package_and_version
     | Overview _ -> Url.Package.overview package.name ?version
     | Documentation _ -> Url.Package.documentation package.name ?version ?hash ?page
   in
-  let version_options v =
-    <% if v = package.latest_version then ( %>
-    <option value="<%s url None %>" <%s if package.version = Latest then "selected" else "" %>>
-      <%s "latest (" ^ package.latest_version ^ ")" %>
-    </option>
-    <% ); %>
-    <option value="<%s url (Some v) %>" <%s if package.version = Specific v then "selected" else "" %>>
-      <%s v %>
-    </option>
-  in
-  <div class="flex gap-4">
-    <h1 class="m-0 text-3xl"><span class="sr-only">package </span><a class="font-semibold text-3xl" href="<%s Url.Package.overview package.name ?version %>"><%s package.name %></a></h1>
-    <select id="version" name="version" aria-label="version" onchange="location = this.value;"
-      class="leading-8 appearance-none cursor-pointer py-0 rounded-md border border-gray-400 pr-8"
-      style="background-position: right 0.25rem center">
-      <%s! package.versions |> List.map version_options |> String.concat "" %>
-    </select>
-  </div>
+   <div class="flex gap-4 justify-center">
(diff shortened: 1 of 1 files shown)


sabine:
Hey @ademolaomosanya! That looks pretty good visually already.

Concerning accessibility, the major challenge is to recreate keyboard navigation on the open dropdown. You can see one example of keyboard navigation via alpinejs on the package search dropdown.

When you inspect github's branch dropdown, you can see that it uses `summary` and `details` HTML elements. I would recommend trying that instead of a button controlled by JavaScript.
Screenshot from 2023-11-01 17-31-12

ademolaomosanya:
Thank you for the feedback @sabine. 
Was not aware the chanllenge on accessibility is to recreate keyboard navigation on the open dropdown
I will work on that in the next commit

ademolaomosanya:
https://github.com/ocaml/ocaml.org/assets/87182204/555b9175-026a-4d5d-ae67-988b8608d538

- changed the button and used summary and details instead

- links has not been updated to render version based on selected link

sabine (commented):
This is looking really nice already. :+1: 

I think there are more keyboard interactions that need to be replicated, like
- escape to close the menu
- enter key to navigate to the selected option

When using both the arrow keys and the tab key, two items appear to be selected at the same time. It is not really clear what is happening

The "(latest)" entry seems to be missing.

I think we also need to set an aria attribute that makes the element currently selected via arrow keys visible to assistive technology: See https://www.w3.org/WAI/ARIA/apg/patterns/combobox/examples/combobox-select-only/.

(This is quite a tricky issue in a sense, getting things to work well.)

ademolaomosanya:
thank you @sabine for the feedback.
it is a really tricky issue 
i will check the link out and make adjustment to fix the remaining requirements

sabine:
Closing this because the PR seems abandoned. If not, can reopen.

Either way, thanks @ademolaomosanya for trying, this wasn't a simple issue. :)

PR #1726, opened 2023-10-30
Dedicated book page for individual books

### LINKED ISSUE #1640: Create a Dedicated Page for Individual Books, so the Books on the Learn Overview Page Link Somewhere
Hello @sabine @SaySayo  I was trying to get help from ocaml- documentation and I saw a page for recommended books. Unfortunately there's no way to navigate to the books unless you click on "see more" and a new page opens. 
I don't know if it's by design. 
Here's how to reproduce the error
1. Go to ocaml.org
2. Navigate to "learn"
3. Scroll down to books and try to click on any book.

I was using my  mobile phone so I don't know if it's reproducible with desktop

Oluwabukolab:
Fixes #1640. Still a draft

Created an individual page for each book on the books page. I'm still working on it but wanted to be sure I'm the right path. cc @SaySayo  @sabine

### DIFF
--- a/src/global/url.ml
+++ b/src/global/url.ml
@@ -48,6 +48,7 @@ let manual = "/releases/latest/manual.html"
 let api_with_version v = v2 ^ "/releases/" ^ minor v ^ "/api/index.html"
 let api = "/releases/latest/api/index.html"
 let books = "/books"
+let book slug = books ^ "/" ^ slug
 let changelog = "/changelog"
 let changelog_entry id = "/changelog/" ^ id
 let releases = "/releases"
--- a/src/ocamlorg_frontend/components/book.eml
+++ b/src/ocamlorg_frontend/components/book.eml
@@ -1,0 +1,11 @@
+let render 
+(book : Data.Book.t) =
+  <div class="flex flex-col gap-y-4">
+    <div class="relative w-64 sm:w-72 lg:w-64 max-w-full space-y-5">
+          <%s! if book.pricing = "free" then Learn_components.pricing_tag else "" %>
+          <img src="<%s Ocamlorg_static.Media.url book.cover %>"
+            class="w-full border z-10"
+            alt="">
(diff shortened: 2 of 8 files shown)


sabine:
Hey @sophiatunji, you're on the right path, but the template will have to take a parameter `(book : Data.Book.t)`

Oluwabukolab:
I've successfully created an individual page for each book entry and I added it to the learn home page so each book can be clicked to link to it. I have a few questions on how the page is meant to look. Do we want it to be displayed exactly like each book is on the books page like this screenshot? or is there a proposed design in mind? cc @sabine @SaySayo 
Screenshot 2023-11-09 at 10 23 08

Also, I tried linking each book in the books page too but I'm unsure where it's meant to click from, I currently encompass the entire book entry into an `a` tag, but that just highlights the entire text as a link. 
Screenshot 2023-11-09 at 10 26 11

sabine:
@sophiatunji great progress!

> Do we want it to be displayed exactly like each book is on the books page like this screenshot? or

Yes, I think that would be best, just repeating the design from the `books.eml` template seems fine here.

> Also, I tried linking each book in the books page too but I'm unsure where it's meant to click from

I think we don't want to link to the individual books page from the books page because all information is already laid out on the books page.

The only reason why we need an individual book page is because the Learn landing page links to some books, and we would like to give people the info just about that single book. The alternative to having a dedicated page for each book would be to have a way to link into the books page such that we land at the exact book referenced.

sabine:
@sophiatunji do you plan to continue on this?

Oluwabukolab:
Hi @sabine, yes I do plan to complete this PR, my apologies about the delay.

sabine:
Hello @sophiatunji, I think adding a dedicated page for every book became obsolete with the new design that includes the external links directly in the book blocks. See https://github.com/ocaml/ocaml.org/pull/1834

PR #1729, opened 2023-10-31
Update pagination.eml possible fix to #1728

### LINKED ISSUE #1728: pagination on package search result page shows  extra "1" when result fits into one page
[screen-capture.webm](https://github.com/ocaml/ocaml.org/assets/70571088/43376f30-b236-4329-a341-76006d456abd)

AndroGenius-codes:
Resolves #1728

@sabine @SaySayo  i made a simple ui fix to the pagination function. its a possible fix to the previous issue i opened #1728 
[screen-capture (1).webm](https://github.com/ocaml/ocaml.org/assets/70571088/2ca42cf7-2914-420b-ab83-b8b84d2783f7)

### DIFF
--- a/src/ocamlorg_frontend/components/pagination.eml
+++ b/src/ocamlorg_frontend/components/pagination.eml
@@ -23,6 +23,7 @@ let right_arrow ~base_url page ~query=
 
 let render ~total_page_count ~page_number ~base_url ~query=
   <div class="pt-16 flex items-center justify-center space-x-2.5">
+  <% if  total_page_count > 1 then ( %>
     <%s! if page_number != 1 then (left_arrow ~base_url page_number ~query) else "" %>
     <% if page_number >= 4 then ( %>
     <span class="w-10 h-10 lg:w-14 lg:h-14 border-body-100 rounded-lg flex items-center justify-center text-lighter hover:no-underline">...</span>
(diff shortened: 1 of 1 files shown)


sabine:
That you saw this and tried to make a fix is great @AndroGenius-codes! :heart: 

I noticed one thing, though:

When there's only a single page, I think we do not need to render pagination at all.

So the simple and clean solution here looks to be to make one big `if then else` statement on `total_page_count > 1` to decide whether to render the pagination.

AndroGenius-codes:
> That you saw this and tried to make a fix is great @AndroGenius-codes! ❤️
> 
> I noticed one thing, though:
> 
> When there's only a single page, I think we do not need to render pagination at all.
> 
> So the simple and clean solution here looks to be to make one big `if then else` statement on `total_page_count > 1` to decide whether to render the pagination.

Makes sense. Let me do that

AndroGenius-codes:
> That you saw this and tried to make a fix is great @AndroGenius-codes! ❤️
> 
> I noticed one thing, though:
> 
> When there's only a single page, I think we do not need to render pagination at all.
> 
> So the simple and clean solution here looks to be to make one big `if then else` statement on `total_page_count > 1` to decide whether to render the pagination.

Thank you for the feedback, the new commit accomplishes this.

AndroGenius-codes:
[screen-capture (2).webm](https://github.com/ocaml/ocaml.org/assets/70571088/cc9259e1-1a81-4e36-8123-0f7fdbe5d5f0)
new appearance of pagination feature

sabine (approved):
Thanks @AndroGenius-codes! LGTM

AndroGenius-codes:
> Thanks @AndroGenius-codes! LGTM

Thank you too

PR #1735, opened 2023-11-01
Revert #1051: remove yoshi

tmattio:
All the data source we've added since the introduction of Yoshi have been in plain OCaml. Clearly, Yoshi isn't providing enough value to have be compelling to generate the data modules from the static files.


sabine (approved):
LGTM

PR #1739, opened 2023-11-01
fix: typo in md file

### LINKED ISSUE #1730: Typo in "A Tour of OCaml" Pattern Matching Cont'd section
At the very end of the "Pattern Matching, Cont'd" section:

> Note that OCaml throws a warning when pattern matching does not catch call cases:

I'm assuming the word "call" is intended to be "all".

FatumaA:
closes #1730 
This is a quick fix for the typo reported in A tour of Ocaml markdown file.

### DIFF
--- a/data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
+++ b/data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
@@ -413,7 +413,7 @@ val g' : string -> int = <fun>
 
 The underscore symbol is a catch-all pattern; it matches with anything.
 
-Note that OCaml throws a warning when pattern matching does not catch call cases:
+Note that OCaml throws a warning when pattern matching does not catch all cases:
 ```ocaml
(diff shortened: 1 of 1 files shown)


FatumaA:
Hi @sabine,
Also set up prettier to not auto format files in this project so I think the diff this time would be minimal

christinerose (approved):
Nice catch!

sabine:
Thank you @FatumaA!

sabine (approved):

FatumaA:
> Nice catch!

wasn't me but thanks 😃

FatumaA:
> Hi @sabine, Also set up prettier to not auto format files in this project so I think the diff this time would be minimal

Would this be something worth adding to the project? To overwrite any auto formatting that may be set up on the contributor side?

sabine:
@FatumaA that is a good question. It looks to me like that would require setting many different editor configurations, and create significant noise in the repository. I think it would be better to watch out for this when contributing.

PR #1743, opened 2023-11-02
Improve the `Map` guide

### LINKED ISSUE #1644: Map guide: stop promoting using clear text passwords
The guide for [Map](https://ocaml.org/docs/map) is nicely done but promotes the bad practice of using clear text passwords.

I suppose this is common knowledge but this looks bad as-is IMO, plus there's the off chance that some reader may not know the security implications of doing so.

I propose to take another example for Map: **Books**, having

- an ISBN number for the key
- a title for the value

I can take care of rewriting the page, but getting some feedback first is probably a good idea.

benjamin-thomas:
- remove references to clear text passwords
- use different types for the keys and values

---

This fixes https://github.com/ocaml/ocaml.org/issues/1644

In the end, I didn't go for `Books` as I felt having the key and value types differ would be better.

### DIFF
--- a/data/tutorials/language/1ds_02_map.md
+++ b/data/tutorials/language/1ds_02_map.md
@@ -11,40 +11,44 @@ category: "Data Structures"
 ## Module Map
 
 Map creates a "mapping". For instance, let's say I have some data that is
-users and their associated passwords. I could with the Map module create
-a mapping from user names to their passwords. The mapping module not
-only does this but it does it fairly efficiently. It also does this in a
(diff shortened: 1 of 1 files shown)


sabine:
@benjamin-thomas I love this example! :+1: 

It's much clearer now with different types for the key and value of the `Map` and not showing bad practices!

After review from @christinerose, we'll merge this.

christinerose (approved):

christinerose:
Great suggestion!

sabine (approved):

PR #1657, opened 2023-10-16
Paginate the Package Search Results

Girish-Jangam:
Resolves #1588

## Draft Pull Request

**Problem:**

I modified the packages_search in handler.ml and also rendered pagination on packages_search.eml

But i failed to construct the base URL for the pages of pagination. Please guide me to resolve this issue.

**Related Issues:**

I think, the Dream.query from paginate function which extracts page number from the URL, it appends ?P=number to the Search value if & added as separator to base URL . Therefore further pages of pagination won't work because it keeps appending page query with the Search.

**Testing :**

I tested it by doing Dream.logs on the console, There are no errors for page, current items, number of pages variables. They are getting assigned with correct values.

**Screenshots (if applicable):**

https://github.com/ocaml/ocaml.org/assets/59350195/21d731f7-4dd7-49e7-9dd7-7ac0a31012dc

**Checklist:**

- [x ] I have read and followed the project's contributing guidelines.
- [x ] I have reviewed my code for any potential issues.
- [x ] I have tested the changes locally.
- [ ] I have updated or added relevant documentation.
- [x ] I have added appropriate comments to the code.
- [x ] My code follows the coding style and conventions of the project.
- [ ] The branch is up-to-date with the base branch (usually 'main' or 'master').
- [ ] The code builds and passes all automated tests.
- [ ] I have run linters and code formatters, and the code complies with them.

By submitting this draft pull request, I acknowledge that it is a work in progress, and I'm seeking feedback from maintainers and the community.

Please review and provide feedback on the proposed changes. Your input is greatly appreciated.

PR #1751, opened 2023-11-09
Add Pagination to the Changelog Page

### LINKED ISSUE #1561: Add Pagination to the Changelog
# Task

Currently, all the entries of the changelog are displayed on a single HTML page at https://ocaml.org/changelog.

Since there are a lot of entries, it makes sense to add pagination:
Screenshot 2023-10-03 at 15-13-43 OCaml Blog

As an example of how to do pagination (and for possible code sharing), look at how it is done for the Blog:
1. In `handler.ml`, the list of entries is paginated according to a query parameter "p" from the URL: https://github.com/ocaml/ocaml.org/blob/6a36190efe178a6236a13e4cd79adbf9b42e10c6/src/ocamlorg_web/lib/handler.ml#L191-L218
2. In the template, the pagination component is rendered: https://github.com/ocaml/ocaml.org/blob/6a36190efe178a6236a13e4cd79adbf9b42e10c6/src/ocamlorg_frontend/pages/blog.eml#L67

--

What you need to do:

1. modify the function `changelog` in `handler.ml` to paginate the list of changelog entries.
2. render the pagination component in `changelog.eml`

(Related to https://github.com/ocaml/ocaml.org/issues/1125.)

m-spitfire:
closes #1561

I kept 10 items per page.

Here's how it looks now:
Screenshot 2023-11-10 at 2 18 05 AM

### DIFF
--- a/src/ocamlorg_frontend/components/pagination.eml
+++ b/src/ocamlorg_frontend/components/pagination.eml
@@ -1,40 +1,45 @@
-let paginated_href ~base_url ~page ~query =
-  base_url ^ "?p=" ^ string_of_int page ^ (if query = "" then "" else "&q=" ^ query)
+let rec construct_query_string = function
+  | h :: t -> "&" ^ (fst h) ^ "=" ^ (snd h) ^ construct_query_string t
+  | [] -> ""
 
-let link ~base_url ~page ~current ~query =
+let paginated_href ~base_url ~page ~queries =
+  base_url ^ "?p=" ^ string_of_int page ^ (construct_query_string
+  queries)
+
+let link ~base_url ~page ~current ~queries =
   let active_class = "border-2 border-current text-primary-600" in
   let non_active_class = "border border-body-100 text-lighter hover:bg-gray-50 hover:no-underline" in
-  <a href="<%s (paginated_href ~base_url ~page ~query) %>"
+  <a href="<%s (paginated_href ~base_url ~page ~queries) %>"
     class="w-10 h-10 lg:w-14 lg:h-14 rounded-lg flex items-center justify-center <%s if page = current then active_class else non_active_class %>">
     <%i page %>
   </a>
 
-let left_arrow ~base_url page ~query=
-  <a href="<%s (paginated_href ~base_url ~page:(page - 1) ~query) %>"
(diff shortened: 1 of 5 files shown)


sabine (commented):
Looks pretty great already, just one minor nitpick and one actual concern:

1. Show more items, maybe 30 per page.
2. When a tag is selected, we have to consider that in the pagination hrefs as well. https://github.com/ocaml/ocaml.org/pull/1657/commits/34bcd238a9bb97f0ca236efe70f126457a708860 faced the same problem and introduced the function `paginated_href` to accomodate the `q=` query param. `paginated_href` could be generalized to handle arbitrary query params.

tmattio:
Is loading the page an issue (in that it's become too large)? If not, I'm wondering if an infinite scroll would be a better alternative to pagination for the Changelog. What do you think?

sabine:
@tmattio infinite scroll is a prettier solution in general if done well, but I think it should be another task to transform all the paginations to infinite scroll at the same time. This is a thing where HTMX would likely do well.

Yes, the page is pretty heavy.

tmattio:
> @tmattio infinite scroll is a prettier solution in general if done well, but I think it should be another task to transform all the paginations to infinite scroll at the same time. This is a thing where HTMX would likely do well.

Pagination made more sense to me for the other pages of the site, whereas with the changelog being a timeline, it makes it a better candidate than other paginated content in my mind. But if you prefer implementing pagination first and changing that to infinite scrolling later, that's ok too.

> Yes, the page is pretty heavy.

Can you share numbers (size and time to load)? The page usually loads fast enough for me that I never paid attention to the fact that it was larger than usual. Having everything in one page was also convenient at times, so just to understand how we're improving UX with the pagination

m-spitfire:
@sabine I addressed the issues you mentioned. I'm not sure if I love how I named the search params variable `queries`, looking forward to your review.

Also current implementation assumes that there will be no parameter with the "key" `p`, but I don't think that's a problem.

sabine (approved):
Looks good to me. I have no qualms with the name `queries` - it makes clear that there's a list of query params.

I also agree that assuming that there's no query param `p` is fine.

Thanks for the contribution @m-spitfire! :+1:

Issue #1562, opened 2023-10-03
Implement Capability to Link Exercise to Tutorials


sabine:
# Task

Currently, tutorials and exercises are not related to each other, even though it would be helpful to recommend to people at the end of a tutorial some exercises they should be able to do now.

To achieve that, we need the ability for exercises to reference the tutorials that they should be recommended on.

1. You need to add a field `tutorials: string list` to both the `metadata` and `t` types in `exercise.ml` which holds the slugs of all the tutorials the exercise is linked to. This will make the command-line tool `ood-gen` load the new reference between tutorials and exercises.
2. Add some sensible references between exercises and tutorials in the `data/exercises` folder, for testing purposes.
3. In `handler.ml`, in the function `tutorial`, you need to filter the exercises for those that belong to the tutorial, i.e. (`tutorial.slug` is an element of `exercise.tutorials`).
4. In `tutorial.eml`, render a plain list of links to the exercises at the bottom of the page.

**Important:** When a referenced tutorial slug does not exist, an error should be thrown by `ood-gen` so that we can be sure that the references between tutorials and exercises will stay in a good shape.

biroue10:
@sabine can you assign this task please ?

sabine:
@biroue10 @SaySayo assigned

biroue10:
@sabine @SaySayo can you give me some guidance about this task please

SaySayo:
Hi @biroue10, to get started, 
1. go into the `problems.ml` file, you'd see some predefined type which is a record field called `metadata` and `t`, modify it according to the instructions on the issue, thats the first step. 
2. Then go into the `data/problems` directory, and modify the md files, adding references of what exercises is related to what issue. 
3. In the `handler.ml` file, find the function that handles the `tutorials` and filter out exercises that fits the exercise references.
4. In the `tutorial.eml` file, display these exercises which have been filtered. 
Is this clear enough or do you need more details?

biroue10:
@SaySayo yes i think now it is more clear thank you

sabine:
@biroue10 are you still working on this task? If yes, it would be helpful to open a draft pull request so we can give feedback and guide you better. If no, please let us know so we can assign to someone else.

biroue10:
@sabine you can assign it to someone else

Raven49:
Hi @sabine @SaySayo  , I have made my first contribution as #1658  Kindly assign me any new task so that i can contribute for Outreachy.

[...]

PR #1753, opened 2023-11-09
Link exercises to turorials

florentdrousset:
Related to issue #1562 

I've just added the `tutorials` field to the types `metadata` and `t` to open the draft PR. This is a WIP.

### DIFF
--- a/src/ocamlorg_frontend/pages/tutorial.eml
+++ b/src/ocamlorg_frontend/pages/tutorial.eml
@@ -24,9 +24,23 @@ let of_tutorial_section (s: Data.Tutorial.Section.t) =
   | Platform -> Learn_layout.Platform
   | Guides -> Learn_layout.Guides
 
+let render_related_exercises exercises =
+  if List.length exercises > 0 then
+    <div class='related-exercises'>
+      <h3>Related exercises :</h3>
+      <ul>
+        <% exercises |> List.iter (fun (exercise : Data.Exercise.t) -> %>
+          <li><a href='/exercises#<%s exercise.number %>'><%s exercise.title %></a></li>
+        <% ); %>
+      </ul>
+    </div>
+  else
+    ""
+
 let render
 (tutorial : Data.Tutorial.t)
 ~tutorials
+~related_exercises
 ~canonical
 =
 Learn_layout.render
@@ -39,5 +53,6 @@ Learn_layout.render
 ~current:(of_tutorial_section tutorial.section) @@
   <div class="prose prose-orange max-w-full">
     <%s! tutorial.body_html %>
+    <%s! render_related_exercises related_exercises %>
     <%s! Learn_components.contribute_footer ~href:("https://github.com/ocaml/ocaml.org/blob/"^ Commit.hash ^"/data/"^ tutorial.fpath) %>
   </div>
--- a/data/exercises/001_tail.md
+++ b/data/exercises/001_tail.md
@@ -4,7 +4,7 @@ number: "1"
 difficulty: beginner
 tags: [ "list" ]
 description: "Write a function that returns the last element of a list."
-tutorials: [ "data-types", "functionnal-programming"]
+tutorials: [ "lists"]
 ---
 
 # Solution
--- a/data/exercises/001_tail.md
+++ b/data/exercises/001_tail.md
@@ -4,6 +4,7 @@ number: "1"
 difficulty: beginner
 tags: [ "list" ]
 description: "Write a function that returns the last element of a list."
+tutorials: [ "data-types", "functionnal-programming"]
 ---
 
 # Solution
--- a/data/exercises/002_tail_penultimate.md
+++ b/data/exercises/002_tail_penultimate.md
@@ -4,6 +4,7 @@ number: "2"
 difficulty: beginner
 tags: [ "list" ]
 description: "Find the last and penultimate elements of a list."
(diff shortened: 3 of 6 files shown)


florentdrousset:
@sabine Sorry, it seems that I'm already stuck on this. I've added the field `tutorials` in the types `t` and `metadata` as it can be seen in my commit, but I'm getting an error when building the project : 

```
ood-gen: internal error, uncaught exception:
         Ood_gen.Exn.Decode_error("Didn't find the function for key: tutorials")
         Raised at Ocamlorg__Import.Result.get_ok.(fun) in file "src/global/import.ml", line 103, characters 55-70
         Called from Ood_gen__Exercise.all in file "tool/ood-gen/lib/exercise.ml" (inlined), line 74, characters 13-52
         Called from Ood_gen__Exercise.template in file "tool/ood-gen/lib/exercise.ml", line 99, characters 4-12
         Called from Dune__exe__Gen.cmds.(fun) in file "tool/ood-gen/bin/gen.ml", line 38, characters 48-61
         Called from Cmdliner_term.app.(fun) in file "cmdliner_term.ml", line 24, characters 19-24
         Called from Cmdliner_eval.run_parser in file "cmdliner_eval.ml", line 34, characters 37-44
```

I've tried to add some logic for filtering in the `handler.ml`, but it doesn't seem to relate to my problem. I've looked through the ood-gen tool but haven't been able to locate where or how the exercise data is processed in relation to this field.

sabine:
Hey, nothing to worry about, this part of the site is not very self-explanatory at this point.

What's going on here:

Now that a field `tutorials` has been added, the tool `ood-gen` parses the YAML header of some exercise, and it's not finding a field `tutorials` in that YAML header. So I would recommend changing the type to `string list option`, which means the key `tutorials` is either absent, or there is a list of strings there.

florentdrousset:
Oh right, I understand better now, and it works just fine with `string list option` as a type for `tutorials` ! Thanks a lot, I can advance on this.

florentdrousset:
@sabine I feel like I'm being stuck again 😅 
I've added some filtering logic in `handler.ml` but so far it looks like I'm not able to pass the exercises I want to the template `tutorial.eml`, and I'm getting this error : 

```
File "src/ocamlorg_frontend/pages/tutorial.eml", line 29, characters 40-46:
Error: Unbound record field number
```

I'd like to build my links to the exercises in a similar fashion to what's done in the function `render_exercise_link` in the template `exercises.eml`, but I'm not sure what I'm missing here and why I can't access `exercise.number`. 
I also have a hard time logging and debugging values, so it might just be my filtering logic that could be bugged. 🤔

florentdrousset:
I'm still stuck on this, so if anyone has an idea of what I'm doing wrong here, I'm definitely interested.

src/ocamlorg_frontend/pages/tutorial.eml
```diff
@@ -24,9 +24,17 @@ let of_tutorial_section (s: Data.Tutorial.Section.t) =
   | Platform -> Learn_layout.Platform
   | Guides -> Learn_layout.Guides
 
+let render_related_exercises exercises =
+  let exercise_html = exercises |> List.map (fun exercise ->
```
sabine:
```suggestion
  let exercise_html = exercises |> List.map (fun (exercise : Data.Exercise.t) ->
```

I think this is one of the few situations in OCaml where we have to put a type annotation on the anonymous function. From what I've seen, type inference on anonymous functions seems to be different compared to type inference on functions defined via `let` binding.

Thanks for bearing with me. :slightly_smiling_face:

florentdrousset:
Screenshot_20231114_210859

Alright it seems to be working now, thanks a lot for your help @sabine ! Not sure I would ever found that one by myself. 🤔 
Do we want to render it the way it can be seen on the screenshot ? (ie : <h3>Related exercises</h3> follow by related exercises with a link)

src/ocamlorg_frontend/pages/tutorial.eml
```diff
@@ -24,9 +24,20 @@ let of_tutorial_section (s: Data.Tutorial.Section.t) =
...
 
+let render_related_exercises exercises =
+  let exercise_html = exercises |> List.map (fun (exercise : Data.Exercise.t) ->
+    let link = "/exercises#" ^ exercise.number in
+    "<a href='" ^ link ^ "'>" ^ exercise.title ^ "</a>"
+  ) |> String.concat "<br>" in
+  if List.length exercises > 0 then
+    "<div class='related-exercises'><h3>Related exercises:</h3>" ^ exercise_html ^ "</div>"
+  else
+    ""
```
sabine: While string concatenation may work (it is relatively safe to assume we will not allow XSS attacks via the exercise data), it will be better to make use of the template language to render the HTML and escape the strings we display.

In the template `platform.eml`, there's some examples of iterating over a list rendering templates within a bigger template.

Here, I would recommend to try a similar thing. Instead of concatenating strings with `br`, let's render a `ul` wrapping `li` elements that contain the links.

sabine (approved):
Hey @florentdrousset, thanks for being so patient with me!

This looks fine to me to merge. Thanks for the contribution! :+1:

florentdrousset:
@sabine I'm glad !! And "thanks for being so patient with me!" should be my line I think. 😆 
I'll check for other issues then 👍🏻

Issue #1624, opened 2023-10-12
Reducing the News column on the Blog Page


iamkalio:
@sabine @SaySayo  I hope you're having a great day. I went through the Issues and don't think this issue has been raised and not sure if there is a plan for the  section, but, 
on the Blog page the News column is way longer than the Ocaml planet section and this has a large white noise before you get to the footer.
How about we reduce the array of the News at News Section so that  the number of News rendered can match the Ocaml Planet Column,  the most updated is at the top and other can been seen with the `see more` button  to reduce the space below the Ocaml Planet column

<img width="960" alt="Screenshot 2023-10-12 103323" src="https://github.com/ocaml/ocaml.org/assets/89003403/64a244f5-2e9f-4790-a64f-7d536eb87dc8">

sabine:
Sounds good to me. This page is due for a redesign in the future, but for now reducing the number of news elements to be displayed would improve the page. This makes a good simple outreachy issue.

Abolade017:
can i be assigned this issue, I am interested in it and i will like to make contribution on it @SaySayo

SaySayo:
You will be assigned this issue, @Abolade017. Cc @gpetiot

Abolade017:
> You will be assigned this issue, @Abolade017. Cc @gpetiot

alright

Abolade017:
> Sounds good to me. This page is due for a redesign in the future, but for now reducing the number of news elements to be displayed would improve the page. This makes a good simple outreachy issue.

I have made the news column and the ocaml planet columns to be equal in my first commit and i later found out that the images in mobile view is not aligned well, then i thought of centering it and adjusted the width because the ocaml logo image  width is not equal to the other images

sabine:
I think there is some confusion what should be done here.

The only thing that should be done here is to change the number of news items so that the height of the News column and the OCaml Planet column on the largest screen is approximately the same height.

kiyov09:
Is this issue available @sabine? I would like to help.

sabine:
@kiyov09 sounds great to me.

kiyov09:
Got it. I'll do it as soon as I can. Thnx

kiyov09:
This is fixed and live now. FYI

tmattio:
Thanks @kalio007!

PR #1754, opened 2023-11-09
Reduce the number of News in the Blog page

kiyov09:
I did an analysis and, on average, the smaller `Post` items have a height 1.2 times greater than the `News` items. So, I reduced the number of `News` items in the Blog page to 1.2 of the number of `Post` items. (featured + posts)

This is my fix for #1624

### DIFF
--- a/src/ocamlorg_web/lib/handler.ml
+++ b/src/ocamlorg_web/lib/handler.ml
@@ -246,7 +246,11 @@ let blog req =
          Data.Planet.Post.all)
   in
   let featured = Data.Planet.featured_posts |> List.take 3 in
-  let news = Data.News.all |> List.take 20 in
+  let number_of_news =
+    List.length featured + List.length current_items
+    |> float_of_int |> ( *. ) 1.2 |> int_of_float
+  in
+  let news = Data.News.all |> List.take number_of_news in
   Dream.html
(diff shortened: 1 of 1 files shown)


kiyov09:
@sabine can you take a look when you have time?

kiyov09:
I tried to fill the space as much as possible, but we can also make them 1:1 with the number of posts and that'll make sure the section will never be greater than the posts one

src/ocamlorg_web/lib/handler.ml
```diff
@@ -246,7 +246,11 @@ let blog req =
          Data.Planet.Post.all)
   in
   let featured = Data.Planet.featured_posts |> List.take 3 in
-  let news = Data.News.all |> List.take 20 in
+  let number_of_news =
+    List.length featured + List.length current_items
+    |> float_of_int |> ( *. ) 1.2 |> int_of_float
```
```suggestion
    |> float_of_int |> ( *. ) 1.3 |> int_of_float
```

Looks to me like we can risk a few more items here.

Indeed, it will never be exactly the same height, and this improves things for now, even if in very rare cases the news may be taller.

sabine (approved):
Thanks @kiyov09, I hadn't even considered considering the featured posts, but that makes sense. Will merge in a moment!

PR #1764, opened 2023-11-14
Remove links from outreachy project description

Solar-Rays:
Added a space in the url causing a spill over on mobile


Solar-Rays:
@sabine please review the fix and advise further

data/outreachy.yml
```diff
@@ -172,9 +172,11 @@ rounds:
           for example, in the form of macros or templates. The OCaml compiler
           doesn't provide a full built-in macro system, but the OCaml parser does
           provide syntax for preprocessing purposes: attributes
-          (https://v3.ocaml.org/docs/metaprogramming#attributes-and-derivers) and
+          (https://v3.ocaml.org/docs/metaprogramming
+          #attributes-and-derivers) and
           extension points
-          (https://v3.ocaml.org/docs/metaprogramming#extension-nodes-and-extenders).
+          (https://v3.ocaml.org/docs/metaprogramming
+          #extension-nodes-and-extenders).
```
sabine: I think, since these descriptions are just plain text, they cannot contain URLs in the first place. So the best thing here would be to remove the links.

src/ocamlorg_frontend/pages/outreachy.eml
```diff
@@ -10,24 +10,29 @@ Layout.render
...
+
                         <div>Mentee: <i><%s project.mentee %></i></div>
+                        
                         <div>Mentor(s): <i><%s Fmt.(str "%a" (list ~sep:comma string) project.mentors) %></i></div>
+                        
                         <% if String.length project.description > 240 then ( %>
                             <div :class="is_shown ? '' : 'relative max-h-20 overflow-hidden'">
                                 <div class="absolute bottom-0 h-10 left-0 right-0 bg-gradient-to-t from-white"></div> 
-                                <p> <%s project.description %> </p>
+                                <p class="overflow-fix"> <%s project.description %> </p>
```
sabine: `overflow-fix` doesn't seem defined anywhere and I don't think this file needs to be changed when we remove the offending links from the outreachy descriptions.

Solar-Rays: Oh, I added it during debugging and forgot to remove it when I found the bug. Do you need me to take it out or is that already done?

data/outreachy.yml
Solar-Rays: That is fine then.

src/ocamlorg_frontend/pages/outreachy.eml
sabine: when you look at the "Files changed" tab, you'll see that two files are being edited by this PR. Best is to check out the old version of `outreachy.eml`, so the PR doesn't make any changes to this file.

Solar-Rays:
I have restored outreachy.eml to the previous state. I only made the change to the data file with the rogue url. You can check now please.

sabine (approved):
Thank you @Solar-Rays! Merging this now. :+1:

PR #1776, opened 2023-11-16
Add OCamlPro feed to OCaml Planet

redianthus:
Hi,

The blog of OCamlPro was unfortunately missing from the long list of blogs collected by our dear Tarides maintainers.

Unfortunately, I have no idea how to run the scraper and the website has way too much dependencies so I'd appreciate if someone with a working set-up could run it. ( cc @sabine ? :smile: )

Thanks !


tmattio:
Thanks for the PR @zapashcanon, OCamlPro is a great addition to the OCaml Planet!

Running the scraper unfortunately yields:
```
failed to scrape ocamlpro: Invalid_argument("Syndic.Date.of_string: cannot parse \"2023-09-18\"")
```

Looks like a bug in Syndic, so that'll need to be addressed upstream. Alternatively, a workaround could be to update the OCamlPro RSS feed to be readable by Syndic, that might be quicker if you have control over how the RSS feed is generated.

> The blog of OCamlPro was unfortunately missing from the long list of blogs collected by our dear Tarides maintainers.

For context, I pulled the planet sources from https://github.com/ocaml/v2.ocaml.org/blob/master/planet_feeds.txt (which contained OCamlPro RSS feed before it was removed from V2 because it caused build failures). When I pulled it, I removed the entries that didn't work, that must be when the OCamlPro one got dropped. In hindsight, I should have pinged you when I noticed the feed wasn't working anymore, sorry about that 🙂

redianthus:
Thanks. @RadioPotin wrote a fix for the feed. It should be in production later today.

sabine:
@zapashcanon I tried to scrape the feed this morning in order to merge this. It looks like the feed now has a different problem: `syndic` doesn't seem to parse it as either a RSS2 or Atom feed:

```
Downloading https://ocamlpro.com/blog/feed/ ... done failed to scrape ocamlpro: Failure("Neither Atom nor RSS2 feed")
```

https://github.com/kayceesrk/river/blob/2fa44943d11613a09f6f3bc92012a2b75df654d7/lib/feed.ml#L25-L29

This either means the feed is still not RSS2 or Atom compatible or there's a problem in `syndic`, but I haven't looked deeper. Annoyingly, `river` swallows the error message.

redianthus:
Well the feed is working on any client I've tried. I guess Syndic is broken (or correct but much stricter than most clients).

bclement-ocp:
The W3C has an online service to validate RSS feeds: https://validator.w3.org/feed/

Looks like there are still [issues with our feed](https://validator.w3.org/feed/check.cgi?url=https%3A%2F%2Focamlpro.com%2Fblog%2Ffeed%2F).

> Sorry
>
> This feed does not validate.

RadioPotin:
Greetings,

Thank you @bclement-ocp, I was unaware that that online service existed, I pushed my fix for the date format that was mentioned above yesterday and will look into fixing these errors and pushing a new version of the site today.

Thanks for pointing out these oversights @sabine and @tmattio 

:v:

RadioPotin:
Hello,

Just to let you know that our feed is now compliant with the Atom 1.0 spec.

This should be mergeable now.

Thank you, until next time !

sabine:
@RadioPotin yes, I think the feed is Atom compliant now.

Looking into this, it seems we're currently relying on `link` tags being present to look up the URI.

It seems to me that ocaml.org could be changed to fall back to the `id` tag, when no `link` tag is present. I expect this is going to be a possible issue with other feeds as well.

sabine:
Looking into this upstream at https://github.com/kayceesrk/river/pull/11.

sabine:
Thanks for being so patient with us folks! This is merging as soon as the checks go through.

PR #948, opened 2023-02-23
Import Rewrote Set Tutorial from V2 PR

cuihtlauac:
@NebuPookins rewrote the set tutorial for [V2](https://github.com/ocaml/v2.ocaml.org) in 2021. The PR was neither merged nor rejected:

https://github.com/ocaml/v2.ocaml.org/pull/1596

PR #1778, opened 2023-11-20
Update Modules and Functors Tutorials

cuihtlauac:
Reorganise, update and complete tutorials on Modules and Functors. The draft outline follows. Top-level items are stand-alone tutorials.

- [ ] Introduction to the Module System
    * In “Get Started” or “Module System” (TBD)
    * Story on names & clashes
    * Separate Compilation
    * Separation of Concerns
    * Information Hidding / Encapsulation
    * Each item has min two links
       * Wikipedia or ext. ref
       * Relevant OCaml tutorial or ref. man  
- [ ] Using Modules
    * Using Standard Library & Opam modules (content from 1st program)
- [ ] Writing Modules
    * Reference to “1st” program + Additional Material
- [ ] Module Includes and Hierarchy
    * Submodules (extracted from existing “Modules” doc)
    * Module Inclusion (extracted from existing “Modules” doc)
    * Dune include_subdirs qualified/unqualified
- [ ] Encapsulation
    * Module Signatures
    * Private definitions
    * Abstract Types (extracted from existing “Modules” doc)
    * Keyword private (Ref man chap 12)
- [ ] Using Functors
    * Signatures
    * Instantiating Functors
- [ ] Writing Functors
    * Defining a Functor

Open questions:
  * Reference or duplicate stuff from “Get Started”
    * [Modules and Standard Library](https://ocaml.org/docs/tour-of-ocaml#modules-and-the-standard-library) in “A Tour of OCaml”
    * [Your First OCaml Program](https://ocaml.org/docs/your-first-program)
  * Relationship with `Set` and `Map` tutorials
  * PR https://github.com/ocaml/ocaml.org/pull/948

P.S. As of creating this PR, it doesn't patch anything. Only discussion


data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,183 +8,381 @@ category: "Module System"
...
 
-## What Are Functors and Why Do We Need Them?
+Learning goals:
+- How to use a functor
+- How to write a functor
+- When to use a functor, some cases
 
-A functor is a module that is parametrised by another module, just like a
-function is a value which is parametrised by other values, the arguments.
+A functor is just a parametrized module.
```
leostera: I think the older description was better here since it describes a functor by analogy with something else that people may be familiar with. 

I'd take it a little further and say that functors _are functions_ but that instead of regular values they take modules as inputs and create modules as outputs. 

This intuition I think can help with other things, like the "transitive closures leading to modules" prerequisite.

data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,183 +8,381 @@ category: "Module System"
...
+## Using an Existing Functor: `Set.Make`
+
+The standard library contains a [`Set`](/api/Set.html) module providing a data structure that allows operations like union and intersection. To use the provided type and its associated [functions](/api/Set.S.html), it is required to use the functor provided by `Set`. For reference only, here is a shortened version of the interface of `Set`.
+```ocaml
+module type OrderedType = sig
+  type t
+  val compare : t -> t -> int
+end
+
+module type S = sig
```
leostera:
```suggestion
(** This is the signature of the module returned by applying `Make` *)
module type S = sig
```

data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,183 +8,381 @@ category: "Module System"
...
+  (* ... *)
+end
+
+module Make : functor (Ord : OrderedType) -> S
```
leostera: Here for example thinking of `Make` as a function would be rather useful, since the signature, especially the right side of `functor` looks a lot like a function type.

data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,183 +8,381 @@ category: "Module System"
...
+```
+
+The functor `Set.Make` needs to be passed a module of type `Set.OrderedType` to produce a module of type `Set.S`. Here is how it can look like in our project:
+
+**`funkt.ml`**
+```ocaml
+module StringSet = Set.Make(String)
```
leostera, line 99: I think at this point it'd make sense to create a small inline module to connect `String` to that `OrderedType` interface from above more directly:

```ocaml
module StringSet = Set.Make(struct
  type t = string
  let compare = String.compare
end)
```

You can afterwards do the handwavey "because String already has String.compare and String.t, we can just pass the String module". 

But otherwise it's a little magical.

data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,183 +8,381 @@ category: "Module System"
...
+```
+
+With this, the command `dune exec funkt` shouldn't do anything but it shouldn't fail either. Here is the meaning of that statement
+- The module `funkt.StringSet` is defined
+- The module `String` (from the standard library) is applied to the functor `Set.Make`. This is allowed because the `String` module satisfies the interface `Set.OrderedSet`
+  - It defines a type name `t` (which is an alias for `string`)
+  - It defines a function `compare` of type `t -> t -> bool`, that is the function `String.compare`
+- The result module from the functor application `Set.Make(String)` is bound to the name `StringSet`, it has the signature `Set.S`.
```
leostera: Some notes here:

* I'm not sure about the `dune exec` sentence here, do we need to associate functors with calls to `dune`? Could we just explain the meaning of the `Set.Make` call directly?

* If we use a `struct ... end` to make the input module explicit, we don't have to explain why String magically fits.

data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,183 +8,381 @@ category: "Module System"
...
+
+Add some code to the `funkt.ml` file to produce an executable that does something and check the result.
+
+**`funkt.ml`**
+```ocaml
+module StringSet = Set.Make(String)
+
+let _ =
```
leostera, line 112: I think the whole parsing inputs and splitting by regexp is not adding to working with our functor here, and a hardcoded example would do just the same.

```ocaml
let () = 
  let set = StringSet.of_list ["hello", "world", "world"] in
  StringSet.iter (fun line -> print_endline line) set
```

BUT this makes me think that we haven't given functors a good raison d'être – like why do we even have functors to begin with?

Why does `Set` must be a functor but `Hashtbl` doesn't need to? 

And why does `Hashtbl` offer a `Hashtbl.Make` functor?

I'm realizing I don't know exactly what this is but let's say its because:
* we don't have a polymorphic `compare`
* we do have a polymorphic `hash`

If we had a polymorphic `compare` would we be using `Set` as an example here?

data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,183 +8,381 @@ category: "Module System"
...
+  (name funkt)
+  (public_name funkt)
+  (libraries str))
+```
+
+Check this works using the `dune exec funkt` command, it shouldn't do anything (the empty file is valid OCaml syntax) but it shouldn't fail either. The stanza `libraries str` will be used later.
+
+## Using an Existing Functor: `Set.Make`
+
+The standard library contains a [`Set`](/api/Set.html) module providing a data structure that allows operations like union and intersection. To use the provided type and its associated [functions](/api/Set.S.html), it is required to use the functor provided by `Set`. For reference only, here is a shortened version of the interface of `Set`.
```
leostera: I think I'm missing _why_ we need to use the functor provided by `Set` here – even a small justification for how we want to _reuse_ the functionality of a Set but with a custom `compare` operation.

data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,183 +8,381 @@ category: "Module System"
...
+$ dune exec funkt < dune
+executable
+libraries
+name
+public_name
+str
+funkt
+```
+
+There are no duplicates in a `Set`. Therefore, the string `"funkt"` is only displayed once although it appears twice in the file `dune`.
```
leostera: If we focus on the usage of the new set rather than input, regex splitting, etc, we wouldn't need much of this.

data/tutorials/language/1ms_01_functors.md
cuihtlauac: Sorry, I should have said that what I've written here is interim text. The original intro had some good ideas but it needs to be rewritten. This also applies to the transitive closure thing, of course we're not going to publish this as is.

data/tutorials/language/1ms_01_functors.md
cuihtlauac: You are right, this is good idea.

data/tutorials/language/1ms_01_functors.md
cuihtlauac: * On dune: we just want to have executable examples
* On `struct` : good idea

data/tutorials/language/1ms_01_functors.md
cuihtlauac: What do you think of something like this:
> Most of the implementations of the set operations must use a comparison function. Using the default `Stdlib.compare` would make it impossible to use a user-defined comparison algorithm. Passing the comparaison as higher-order parameter (like `Array.sort` does for instance) would add a lot of boilerplate. Providing set operations as a functor allows to specify the comparison function only once.

data/tutorials/language/1ms_01_functors.md
cuihtlauac: Yes, this is needed, you are right.

data/tutorials/language/1ms_01_functors.md
cuihtlauac: This will be answered afterwards in the “parametrized implementation”. However, we can add a few words saying that being able to use a custom comparison function in all the functions of some sort of set, at once, is justification for set-as-functor.

data/tutorials/language/1ms_01_functors.md
cuihtlauac: You are right. I confess a preference for longer examples, but I see your point.

data/tutorials/language/1ms_01_functors.md
leostera: No worries, and nothing to be sorry for! 🙏🏼

data/tutorials/language/1ms_01_functors.md
leostera: Yes, and this would pave the way for the arguments for functors as a tool for programming-in-the-large! I dig it.

Things like "avoiding repetitive code", "helping establish patterns", or "customizing a lot of code quickly".

cuihtlauac: Text added as a note below

data/tutorials/language/1ms_02_dune.md
```diff
@@ -0,0 +1,241 @@
...
+description: >
+  Learn about the features of Dune that interact with the OCaml module system
+category: "Module System"
+---
+
+# Modules and Libraries in Dune
+
+## Introduction
+
+The goal of this tutorial is
```
christinerose: Do we want to remove this sentence or complete it?

data/tutorials/language/1ms_02_dune.md
```diff
@@ -0,0 +1,241 @@
...
+
+This tutorial uses the [Dune](https://dune.build) build tool. Make sure you have version 3.7 or later installed.
+
+**Requirements**: Modules and Functors
+
+## Minimum Project Setup
+<!-- CONCEPT: Dune setup -->
+
+This section details the structure of an almost minimum Dune project setup. Check [Your First OCaml Program](/docs/your-first-program) for automatic setup using the `dune init proj` command.
```
christinerose:
```suggestion
This section details the structure of an almost-minimum Dune project setup. Check [Your First OCaml Program](/docs/your-first-program) for automatic setup using the `dune init proj` command.
```

almost-minimum is a little awkward .... perhaps "...of a basic Dune project setup" ?

data/tutorials/language/1ms_02_dune.md
```diff
@@ -0,0 +1,241 @@
...
+* Dune creates a module `Wmo` from the contents of folder `lib`.
+* The folder's name (here `lib`) is irrelevant.
+* The library name appears uncapitalised (`wmo`) in `dune` files:
+  - In its definition, in `lib/dune`
+  - When used as a dependency in `dune`
+
+## Library Wrapper Modules
+<!-- CONCEPT: Submodule -->
+
+By default, modules bundled into a library by Dune are wrapped into a module. It is possible to bypass Dune's behaviour by manually writing the wrapper file.
```
christinerose:
```suggestion
By default, when Dune bundles modules bundled into a library, they are wrapped into a module. It is possible to bypass Dune's behaviour by manually writing the wrapper file.
```

How does Dune bundle modules into a module? I mean, it sounds like the same thing. Does it somehow compress many modules into a single module?

data/tutorials/language/1ms_02_dune.md
```diff
@@ -0,0 +1,241 @@
...
+
+This `lib/wmo.ml` file corresponds to the module that Dune automatically generated in the previous section.
+
+**`lib/wmo.ml`**
+```ocaml
+module Cumulus = Cumulus
+module Stratus = Stratus
+```
+
+Here is how this makes sense of these module definitions:
```
christinerose: I don't understand this sentence. Might you mean:
- Here is how this makes sense in these module definitions 
or 
- Here is how to make sense of these module definitions
??

data/tutorials/language/1ms_02_dune.md
```diff
@@ -0,0 +1,241 @@
...
+
+**`wmo.ml`**
+```ocaml
+module Cumulus = Cumulus.M
+module Stratus = Stratus.M
+```
+
+Check with `dune exec nube` that the behaviour of the program is the same as in the previous sections.
+
+## Conclusion
```
christinerose, line 265: Do you want to remove this section or complete it?

data/tutorials/language/1ms_02_dune.md
```diff
@@ -0,0 +1,241 @@
...
+
+# Modules and Libraries in Dune
+
+## Introduction
+
+The goal of this tutorial is
+
+This tutorial uses the [Dune](https://dune.build) build tool. Make sure you have version 3.7 or later installed.
+
+**Requirements**: Modules and Functors
```
christinerose: Perhaps put a link to the tutorial here (when there is one!)

christinerose (commented):
Some initial comments/suggestions/questions 
Haven't double checked the code yet.

data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,183 +8,392 @@ category: "Module System"
...
+end
+
+module type S = sig
+  (* ... *)
+end
+
+module Make : functor (Ord : OrderedType) -> S
+```
+
+**Note**: Most set operation implementations must use a comparison function. Using `Stdlib.compare` would make it impossible to use a user-defined comparison algorithm. Passing the comparison function as a higher-order parameter (like in `Array.sort`, for instance) would add a lot of boilerplate. Providing set operations as a functor allows specifying the comparison function only once.
```
christinerose: "...a lot of boilerplate code." ??
boilerplate is usually an adjective, so would the modified noun be code? file?

data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,183 +8,392 @@ category: "Module System"
...
+  |> String.Set.iter print_endline
+```
+
+This allows the user to seemingly extend the module `String` with a submodule `Set`. Check the behaviour using `dune exec funkt < dune`.
+
+## Functors are Parametrised Modules
+
+### Functors from the Standard Library
+
+Some ”modules” provide operations over an abstract type and need to be supplied with a parameter module used in their implementation. These “modules” are parametrised, in other words, functors. That's the case for the sets, maps, and hash tables provided by the standard library. It works in a contract way:
```
christinerose: What do you mean here by "contract way"? I'm unfamiliar with that term. 
Might you mean "a specific" or "specified" way? Like, it works this way every time?

data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,183 +8,392 @@ category: "Module System"
...
+* If you provide a module that implements what is expected (the parameter interface)
+* The functor returns a module that implements what is promised (the result interface)
+
+Here is the signature of the module that the functors `Set.Make` and `Map.Make` expect:
```
christinerose:
```suggestion
Here is the module's signature that the functors `Set.Make` and `Map.Make` expect:
```
Is this syntax still accurate?

data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,183 +8,392 @@ category: "Module System"
...
+```ocaml
+module type OrderedType = sig
+  type t
+  val compare : t -> t -> int
+end
+```
+
+Here is the signature of the module that the functor `Hashtbl.Make` expects:
```
christinerose:
```suggestion
Here is the functor's signature that the functor `Hashtbl.Make` expects:
```
Still accurate?

data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,183 +8,392 @@ category: "Module System"
...
+* [`Map.S`](/api/Map.S.html)
+* [`Hashtbl.S`](/api/Hashtbl.S.html)
+
+### Writing Your Own Functors
+
+There are many kinds of [heap](https://en.wikipedia.org/wiki/Heap_(data_structure)) data structures. Example include binary heaps, leftist heaps, binomial heaps, or Fibonacci heaps.
+
+The kind of data structures and algorithms used to implement a heap is not discussed in this document.
+
+The common prerequisite to implementing any kind of heap is the availability of a means to compare the elements they contain. That's the same signature as the parameter of `Set.Make` and `Map.Make`:
```
christinerose:
```suggestion
The common prerequisite to implement any heap is the available means to compare the elements contained in that heap. That's the same signature as the parameter of `Set.Make` and `Map.Make`:
```

If this isn't correct, let's work on clarifying and tightening this sentence together.

data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,183 +8,392 @@ category: "Module System"
...
+  type 'a t
+  val f : string t -> unit
+end
+
+module Make(Dep: Iterable) : S with type 'a t := 'a Dep.t = struct
+  let f = Dep.iter (fun s -> Out_channel.output_string stdout (s ^ "\n"))
+end
+```
+
+The module `IterPrint` is refactored into a functor that takes the dependency providing `iter` as a parameter. The `with type 'a t := 'a Dep.t` constraint means the type `t` from the parameter `Dep` replaces the type `t` in the result module. This allows the type of `f` to use the type `t` from the parameter module `Dep`. With this refactoring, `IterPrint` only has one dependency; at the time it is compiled, no implementation of function `iter` is available yet.
```
christinerose:
```suggestion
The module `IterPrint` is refactored into a functor that takes the dependency providing the parameter `iter`. The `with type 'a t := 'a Dep.t` constraint means the type `t` from the parameter `Dep` replaces the type `t` in the result module. This allows the type of `f` to use the type `t` from the parameter module `Dep`. With this refactoring, `IterPrint` only has one dependency. At the time it is compiled, no implementation of function `iter` is available yet.
```

If this isn't accurate, let's work on the syntax together.

data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,183 +8,392 @@ category: "Module System"
...
+  |> List.concat_map Str.(split (regexp "[ \t.,;:()]+"))
+  |> StringSet.of_list
+  |> StringSet.elements
+  |> Array.of_list
+  |> IterPrint.f
+```
+
+Check the program's behaviour using `dune exec funkt < dune`.
+
+**Note**: The functor `IterPrint.Make` returns a module that exposes the type of the injected dependency (here first `List.t` then `Array.t`). That's why a `with type` constraint is needed. If the dependency was an _implementation detail_ that is not exposed in the signature of the initial version of `IterMake` (i.e., in the type of `IterMake.f`), that constraint wouldn't be needed, and the call site of `IterPrint.f` would be unchanged when injecting another dependency.
```
christinerose:
```suggestion
**Note**: The functor `IterPrint.Make` returns a module that exposes the injected dependency type (here, first `List.t` then `Array.t`). That's why a `with type` constraint is needed. If the dependency was an _implementation detail_ that wasn't exposed in the signature of `IterMake`'s initial version (i.e., in the type of `IterMake.f`), that constraint wouldn't be needed, and the call site of `IterPrint.f` would be unchanged when injecting another dependency.
```

If this isn't accurate, let's work on the syntax together!

data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,183 +8,392 @@ category: "Module System"
...
+  end;;
+
+# Array.init 10 Fun.id |> Array.scan_left ( + ) 0;;
+- : int array = [|0; 1; 3; 6; 10; 15; 21; 28; 36; 45|]
+
+# List.init 10 Fun.id |> List.scan_left ( + ) 0;;
+- : int list = [0; 1; 3; 6; 10; 15; 21; 28; 36; 45]
+```
+
+Modules `Array` and `List` appear augmented with `Array.scan_left` and `List.scan_left`.
```
christinerose:
```suggestion
Run the `dune utop` command. Once inside the toplevel, enter the following commands:
```ocaml
# module Array = struct
    include Stdlib.Array
    include ScanLeft.Make(Stdlib.Array)
  end;;

# module List = struct
    include List
    include ScanLeft.Make(struct
      include List
      let of_list = Fun.id
    end)
  end;;

# Array.init 10 Fun.id |> Array.scan_left ( + ) 0;;
- : int array = [|0; 1; 3; 6; 10; 15; 21; 28; 36; 45|]

# List.init 10 Fun.id |> List.scan_left ( + ) 0;;
- : int list = [0; 1; 3; 6; 10; 15; 21; 28; 36; 45]
```

Modules `Array` and `List` appear augmented with `Array.scan_left` and `List.scan_left`. For brevity, the output of the first two toplevel commands is not shown here.
```

Does this make more sense after the example?

File: data/tutorials/language/1ms_01_functors.md [hunk 23]
```diff
@@ -8,183 +8,392 @@ category: "Module System"
...
+
+## Conclusion
```
christinerose, line 518: Should we complete this section or remove it?

christinerose (commented):
Initial comments/suggestions/questions for functors.md

data/tutorials/language/1ms_02_dune.md
cuihtlauac: We need to complete it.

data/tutorials/language/1ms_02_dune.md
cuihtlauac: It needs to be completed

data/tutorials/language/1ms_01_functors.md
cuihtlauac: Is the new text (line 166) an improvement?

data/tutorials/language/1ms_01_functors.md
cuihtlauac: Yes, much better, thanks

data/tutorials/language/1ms_01_functors.md
cuihtlauac: Ok

data/tutorials/language/1ms_01_functors.md
cuihtlauac: I have shortened it further

data/tutorials/language/1ms_01_functors.md
cuihtlauac: Check updated text line 312

data/tutorials/language/1ms_01_functors.md
cuihtlauac: Check updated text line 354

data/tutorials/language/1ms_01_functors.md
cuihtlauac: Yes!

data/tutorials/language/1ms_01_functors.md
cuihtlauac: It needs to be completed, we need some conclusion

data/tutorials/language/1ms_01_functors.md
cuihtlauac: Should we also say something like this?
> The term _functor_ comes from a branch of mathematics called category theory. Although OCaml functors are loosely inspired by them, no knowledge of that math is required to understand and work with OCaml functors. Functors in Haskell more closely represent the Category Theory concept and are, have nothing in common with OCaml functors.

data/tutorials/language/1ms_01_functors.md
```diff
@@ -53,9 +53,8 @@ module type OrderedType = sig
   val compare : t -> t -> int
 end
 
-(** This is the signature of the module returned by applying `Make` *)
 module type S = sig
-  (* ... *)
+  (** This is the signature of the module returned by applying `Make` *)
```
christinerose:
```suggestion
  (** This is the module's signature returned by applying `Make` *)
```
Would this still be accurate?

data/tutorials/language/1ms_01_functors.md
christinerose: Yes! Much more clear!

data/tutorials/language/1ms_01_functors.md
christinerose: Nicely done!

data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,183 +8,407 @@ category: "Module System"
...
+  type 'a t
+  val f : string t -> unit
+end
+
+module Make(Dep: Iterable) : S with type 'a t := 'a Dep.t = struct
+  let f = Dep.iter (fun s -> Out_channel.output_string stdout (s ^ "\n"))
+end
+```
+
+The module `IterPrint` is refactored into a functor that takes as a parameter a module providing the function `iter`. The `with type 'a t := 'a Dep.t` constraint means the type `t` from the parameter `Dep` replaces the type `t` in the result module. This allows the type of `f` to use the type `t` from the parameter module `Dep`. With this refactoring, `IterPrint` only has one dependency; at the time it is compiled, no implementation of function `iter` is available yet.
```
christinerose:
```suggestion
The module `IterPrint` is refactored into a functor that takes a module providing the function `iter` as a parameter. The `with type 'a t := 'a Dep.t` constraint means the type `t` from the parameter `Dep` replaces the type `t` in the result module. This allows the type of `f` to use the type `t` from the parameter module `Dep`. With this refactoring, `IterPrint` only has one dependency. At the time it is compiled, no implementation of function `iter` is available yet.
```

Minor suggestions, if you like it better.

data/tutorials/language/1ms_01_functors.md
christinerose: Much better! I suggested two little tweaks, but it showed up at the bottom of these comments. I'll resolve this one now.

data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,183 +8,407 @@ category: "Module System"
...
+  end;;
+
+# Array.init 10 Fun.id |> Array.scan_left ( + ) 0;;
+- : int array = [|0; 1; 3; 6; 10; 15; 21; 28; 36; 45|]
+
+# List.init 10 Fun.id |> List.scan_left ( + ) 0;;
+- : int list = [0; 1; 3; 6; 10; 15; 21; 28; 36; 45]
+```
+
+Modules `Array` and `List` appear augmented with `Array.scan_left` and `List.scan_left`. For brievety, the output of the first two toplevel commands is not shown here.
```
christinerose:
```suggestion
Modules `Array` and `List` appear augmented with `Array.scan_left` and `List.scan_left`. For brevity, the output of the first two toplevel commands is not shown here.
```

data/tutorials/language/1ms_01_functors.md
christinerose: This is a long sentence with a lot of prepositional phrases: "If the dependency was an _implementation detail_ that is not exposed in the signature of the initial version of `IterMake` (i.e., in the type of `IterMake.f`), that constraint wouldn't be needed, and the call site of `IterPrint.f` would be unchanged when injecting another dependency."

I tried to remove one of them: 
"in the signature of the initial version of `IterMake`"
to
"in the  signature of `IterMake`'s initial version"

It doesn't shorted it much, but it tightens that a little bit. It is correct as it is, but this suggestion makes it a little tighter. Feel free to disregard. 🙂

data/tutorials/language/1ms_02_dune.md
```diff
@@ -0,0 +1,269 @@
+---
+id: libraries-dune
+title: Libraries With Dune
```
sabine, line 3: Should we also consider packages?

Yesterday I did another stream full of skill issue on my part where I tried to create all three kinds of libraries with dune (the singleton, wrapped, and unwrapped sorts) as a package.

Where I got confused and fumbled was when I wanted to install the package containing the three libraries in another switch and I had trouble seeing how to name the libraries so that they show up.

In a sense this topic has some overlap with the OCaml module system, but it also has significant overlap with dune and opam (so kind of feels like it belongs in Platform Tools).

cuihtlauac: There is indeed an overlap. Yes I believe a fourth document is needed on packaging.
1. Units of compilation: module and functors
2. Units of linking & loading: libraries
3. Units of distribution & installation: packages

I wonder if we should rename the "MODULE SYSTEM" section of the "Language" tutorial to reflect on the units listed above

data/tutorials/language/1ms_01_functors.md
```diff
@@ -8,13 +8,14 @@ category: "Module System"
...
 # Functors
 
-
 ## Introduction
 
 In this tutorial, we look at how to use a functor, how to write a functor, and show a couple of use cases involving functors.
 
 As suggested by the name, a _functor_ is almost like a function. However, while functions are between values, functors are between modules. A functor takes a module as a parameter and returns a module as a result. A functor is a parametrised module.
 
+In mathematics, [functor](https://en.wikipedia.org/wiki/Functor) means something different. You don't need to know about those functors to understand OCaml's ones.
```
christinerose:
```suggestion
In mathematics, [functor](https://en.wikipedia.org/wiki/Functor) means something different. You don't need to know about those functors to understand OCaml's.
```

data/tutorials/language/1ms_02_dune.md
```diff
@@ -0,0 +1,269 @@
...
+This is sufficient for `dune build` to work. It will not build anything.
+- `(package (name foo) (allow_empty))` this means we're creating an Opam package named `foo` and we allow it to be empty
+- `(generate_opam_files)` we ask Dune to setup the opam configuration automatically
+
+Here `foo` is the project name and `foo.dir` is its container folder, the names don't have to be the same.
+-->
+
+## Conclusion
+
+The OCaml module system allows organizing a project in many ways. Dune provides several means to arrange modules into libraries.
```
christinerose:
```suggestion
The OCaml module system allows organising a project in many ways. Dune provides several means to arrange modules into libraries.
```

data/tutorials/language/1ms_00_modules.md
```diff
@@ -24,52 +24,52 @@ In OCaml, every piece of code is wrapped into a module. Optionally, a module
 itself can be a submodule of another module, pretty much like directories in a
 file system.
 
-When you write a program, let's say using the two files `amodule.ml` and
-`bmodule.ml`, each automatically defines a module named
-`Amodule` and a module named `Bmodule`, which provides whatever you put into the
+When you write a program, let's say using the two files `athens.ml` and
+`berlin.ml`, each automatically defines a module named
```
christinerose: Do they define modules or create modules (or do those both mean the same thing)?

cuihtlauac: Defines is probably better.

data/tutorials/language/1ms_00_modules.md
```diff
@@ -24,96 +24,92 @@ In OCaml, every piece of code is wrapped into a module. Optionally, a module
...
+`Printf.printf` all over the place.
 ```ocaml
 open Printf
 let data = ["a"; "beautiful"; "day"]
 let () = List.iter (printf "%s\n") data
 ```
 
- The standard library is a module called `Stdlib` where modules `List`, `Option`, `Either`, and others are [submodules](#submodules). Implicitly, all OCaml begins with `open Stdlib`, which avoids writing `Stdlib.List.map`, `Stdlib.Array`, or using `Stdlib.` anywhere.
+ The standard library is a module called `Stdlib` where modules `List`,
+ `Option`, `Either` and others are [submodules](#submodules). Implicitly, all
```
christinerose, line 108:
```suggestion
 `Option`, `Either`, and others are [submodules](#submodules). Implicitly, all
```

data/tutorials/language/1ms_00_modules.md
```diff
@@ -24,96 +24,92 @@ In OCaml, every piece of code is wrapped into a module. Optionally, a module
...
+ OCaml begins with `open Stdlib`. That avoids writing `Stdlib.List.map`,
+ `Stdlib.Array` or using `Stdlib.` anywhere.
```
christinerose, line 110:
```suggestion
 `Stdlib.Array`, or using `Stdlib.` anywhere.
```

data/tutorials/language/1ms_00_modules.md
```diff
@@ -123,94 +119,118 @@ val list_sum_sq : int -> int = <fun>
...
 ```
 
-## Interfaces and Signatures
+## Interfaces and Implementations
+
+By default, anything defined in a module is accessible from other modules.
+Values, functions, types, or submodules, everything is public. This can be
+restricted. That allows distinguishing content provided to other modules from
+internal use content. What is internal is kept private and not available from
+other modules.
```
christinerose:
```suggestion
By default, anything defined in a module is accessible from other modules, including
values, functions, types, or submodules. Everything is public. However, this can be
restricted, so distinguished content provided to other modules can be distinguished from
internal use content. Internal information is kept private and not available from
other modules.
```

Not sure about that second-to-last sentence. If I got it wrong, let's work on it togther.

christinerose (commented):
Suggestions for clarity

data/tutorials/language/1ms_00_modules.md
```diff
@@ -278,9 +276,9 @@ val gimel_of_bool : bool -> gimel
 - : string = "Christine"
 ```
 
-Type `gimel` is _abstract_. Values are available, but only as function results or arguments. Only the provided functions `gimel_of_bool`, `gimel_flip`, and ` gimel_to_string` and polymorphic functions can receive or return `gimel` values.
+Type `gimel` is _abstract_. Values are available, but only as function result or argument. Only the provided functions `gimel_of_bool`, `gimel_flip`, and ` gimel_to_string` and polymorphic functions can receive or return `gimel` values.
```
christinerose: Since "Values" is plural here, it is correct for it to be "function results or arguments," both plural.

data/tutorials/language/1ms_00_modules.md
```diff
@@ -1,545 +1,436 @@
...
-### Submodule Interface
+Definitions from a submodule are accessed by chaining module names, here
+`Florence.Hello.print`.
 
-We can also restrict the interface of a given submodule. It is called a module
-type. Let's do it in our `example.ml` file:
+### Submodule With Signatures
 
+To define an interface to a submodule, we can provide a _module signature_. This
+is done in this second version of the `florence.ml` file:
```
christinerose:
```suggestion
To define a submodule's interface, we can provide a _module signature_. This
is done in this second version of the `florence.ml` file:
```

Would this still be correct?

data/tutorials/language/1ms_00_modules.md
```diff
@@ -1,545 +1,436 @@
...
-The definition of the `Hello` module above is the equivalent of a
-`hello.mli`/`hello.ml` pair of files. Writing all of that in one block of code
-is not elegant so, in general, we prefer to define the module signature
-separately:
+The first version made `Florence.Hello.message` public. In this version it can't be accessed from `glasgow.ml`.
 
-<!-- $MDX skip -->
+### Module Signatures are Types
+
+The role played by module signatures to implementations is akin to the role played by types to values. Here is a third possible way to write file `florence.ml`:
```
christinerose, line 341:
```suggestion
The module signatures' role to implementations is akin to the role played by types to values. Here is a third possible way to write file `florence.ml`:
```
Is this still correct?

data/tutorials/language/1ms_00_modules.md
```diff
@@ -1,545 +1,436 @@
...
+- : int = 89809344
 ```
+
+Values returned by `Random.bits` will differ when you run this code. The first
+and third calls return the same results, showing that the internal state was
+reset.
+
+## Conclusion
+
+In OCaml, modules are the basic means of organising software. To sum up, a module is a collection of definitions wrapped under a name. These definitions can be submodules, which allows the creation of hierarchies of modules. Top-level modules must be files and are the units of compilation. Every module has an interface, which is the list of definitions a module exposes. By default, a module's interface exposes all its definitions, but this can be restricted using the interface syntax.
```
christinerose:
```suggestion
In OCaml, modules are the basic means of organising software. To sum up, a module is a collection of definitions wrapped under a name. These definitions can be submodules, which allow the creation of module hierarchies. Top-level modules must be files and are the units of compilation. Every module has an interface, the list of definitions a module exposes. By default, a module's interface exposes all its definitions, but this can be restricted using the interface syntax.
```

christinerose (commented):
A few suggestions for rewording...

data/tutorials/language/1ms_01_functors.md
```diff
@@ -2,189 +2,522 @@
...
+
+The standard library contains a [`Set`](/api/Set.html) module which is designed to handle sets. This module enables you to perform operations such as union, intersection, and difference on sets. You may check the [Set](/docs/sets) tutorial to learn more about this module, but it is not required to follow the present tutorial.
+
+To create a set module for a given element type (which allows you to use the provided type and its associated [functions](/api/Set.S.html)), it's necessary to use the functor `Set.Make` provided by the `Set` module. For reference only, here is a shortened version of the interface of `Set`:
+```ocaml
+module type OrderedType = sig
+  type t
+  val compare : t -> t -> int
+end
+
```
cuihtlauac, line 54: @sabine Why did you remove this?
```ocaml
module type S = sig
  (** This is the module's signature returned by applying `Make` *)
end
```

Without it, understanding the signature of `Make` is a bit harder. Maybe we could have copied some `S`'s contents to make it easier to understand?

data/tutorials/language/1ms_00_modules.md
cuihtlauac: Yes better

data/tutorials/language/1ms_00_modules.md
cuihtlauac: Yes it is correct. Can we use possessive for the second part of the sentence, after “akin”, the type & value?

data/tutorials/language/1ms_00_modules.md
cuihtlauac: Yes, that's better

PR #1783, opened 2023-11-21
Patching Scraper

sabine:
Relating to https://github.com/ocaml/ocaml.org/pull/1776.

We updated River to fall back to the `id` tag of an entry when no `link`s are provided.

While we're at it: in the watch scraper, we add the missing wrapper object with field `watch:` and reduce the amount of logging in case a posts already exists. This will make it easier to see errors in the scraper log.

River can be unpinned after next release:
- https://github.com/ocaml/opam-repository/pull/24822

### DIFF
--- a/tool/ood-gen/bin/watch_scrape.ml
+++ b/tool/ood-gen/bin/watch_scrape.ml
@@ -52,7 +52,7 @@ let watch_to_yaml t =
         ("category", `String t.category);
       ])
 
-let to_yaml t = `O ["watch", `A (List.map watch_to_yaml t)]
+let to_yaml t = `O [ ("watch", `A (List.map watch_to_yaml t)) ]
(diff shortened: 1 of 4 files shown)


tool/ood-gen/bin/watch_scrape.ml
```diff
@@ -52,7 +52,7 @@ let watch_to_yaml t =
         ("category", `String t.category);
       ])
 
-let to_yaml t = `A (List.map watch_to_yaml t)
+let to_yaml t = `O ["watch", `A (List.map watch_to_yaml t)]
```
tmattio: I'm surprised we can change the yaml without needing to change anything else. Was this broken before?

sabine: I think this broke when yoshi was removed

PR #1839, opened 2023-12-08
OCaml Cookbook

tmattio:
Very rough prototype to start discussions on the addition of a cookbook to the OCaml Learn area.

The goal is to provide short, task-oriented recipes to solve common programming problems, like reading from and writing to a file, querying an HTTP server, parsing JSON, etc.

The idea is that recipes can have variants, that can use different libraries. For instance, we can imagine variants of a recipe for parsing JSON that use Yojson, or Jsonm. Since the language documentation cannot use third-party libraries, the documentation currently doesn't give learners the right tools to get things done in OCaml, and solve real-world problems. The objective is to fill that gap.


sabine:
Ideas for community contributions:
- [ ] previous / next links at the bottom would be nice
- [ ] filtering (categories / tags / package)
- [ ] would be nice to link to cookbook from a package's documentation page
- [ ] have a copy button to copy libraries / ppxes dune file
- [ ] add a table of contents for the categories and subcategories on `cookbook.eml`

tmattio:
Thanks a lot for taking over the cookbook, Sabine! I just ran it locally and it looks great, very exciting!!

Here are a few thoughts I had while looking at it:

- I would keep the cookbook in the Learn area. This is meant to be community driven, and so from a contributor perspective Community makes sense, but I think it's best to think of the sitemap from a visitor's point of view. In that case, I believe Learn is a better section for the cookbook.
- I would simplify the cookbook for now. For instance, I don't think we need to handle files (and certainly not multiple files), sections, explanations for code, etc. Instead, I would start with something deadsimple and only introduce things when we see that we need them (either because we received user feedback, or we have a design from Claire). 
- I think we can get rid of the cookbook categories yaml file. The title can be in the recipe themselves, and the categories (and their names) can be infered from the file structure.

If you agree with the second point, here's a suggestion for what a recipe could look like:

```
---
title: Generate random strings and arrays
packages:
- name: "mirage-crypto-rng"
  version: "0.11.2"
- name: "randomconv"
  version: "0.1.3"
---

\```ocaml
(* Initialize the RNG with the default entropy source. *)
let () = Mirage_crypto_rng_unix.initialize
          (module Mirage_crypto_rng.Fortuna)

(* Generate a `Cstruct.t` with random data. *)
let cstruct n = Mirage_crypto_rng.generate n

(* To get a random `char`, convert a random `int8`: *)
let char () = Char.chr (int8 ())

(* Use `Cstruct`'s conversion methods to obtain random byte arrays, bigarrays or strings. *)
let bytes n = cstruct n |> Cstruct.to_bytes
let bigarray n = cstruct n |> Cstruct.to_bigarray
let string n = cstruct n |> Cstruct.to_string

(* You can also create alphanumeric random characters: *)
let alphanum () =
  Char.chr (48 + Randomconv.int ~bound:74 cstruct)

(* Wrap your random value generator into a sequence to generate as many random values as you want. *)
let seq gen =
  Seq.unfold (fun () -> Some (gen (), ())) ()
let list n gen =
  seq gen |> Seq.take n |> List.of_seq
\```

```

christinerose (approved):

cuihtlauac (approved):

PR #1841, opened 2023-12-10
Fix broken "Install Platform Tools" link

J3RN:
It appears that anchor links are, for better or worse, case-sensitive.


cuihtlauac (approved):

sabine:
Thanks @J3RN!

PR #1843, opened 2023-12-11
Add many packages to "Is OCaml Web Yet?" + rework the page

balat:


sabine (approved):
Thanks @balat, this improves the page significantly!

PR #1846, opened 2023-12-11
Fix escape in "Your First OCaml Program"

J3RN:
I _believe_ that the escaping of the backslash is incorrect here, as it prints the following to the terminal:
```
Hello World!\n
```
I don't think the newline escape is intended to be printed like that (and isn't in later examples).


sabine (approved):
Thanks @J3RN!

PR #1850, opened 2023-12-12
Update to Tailwind CSS 3.3.6

MisterDA:
Both opam and ocaml-ci support pinning tar.gz packages, which are lighter to download than the whole git history.


cuihtlauac:
I've browsed a couple of pages with both versions side by side. I haven't observed any differences.

sabine (approved):

Issue #534, opened 2022-07-26
(cross-ref) online OCaml manual: there should be an easy way to get a fixed-version URL


gasche:
This is a cross-reference to an issue with the OCaml manual which I posted upstream at

https://github.com/ocaml/ocaml/issues/11458

> Suppose I'm browsing a random page of the OCaml manual
> 
>   <https://v2.ocaml.org/manual/types.html#sss:typexpr-sharp-types>
> 
> and I want to refer other people to it. The URL above I found through my search engine does not mention a specific OCaml version, so it is not stable, it may be dead or contain different content next time we look at it.
> 
> Feature wish: it should be easy to get a fixed-version equivalent of the same URL, just like github offers an interface to get a "permalink" from the source view of a versioned file.
> 
> There is a version name in the top-left corner of the page that says "Version 4.14". I would expect that clicking on it brings me exactly to such a fixed-version version of the page, but in fact it brings to <https://ocaml.org/releases/> which is certainly not what I'm looking for.
> 
> Note: this would be easier if the URL scheme was regular among "current versions" and "fixed version" manual pages. Currently we have:
> 
> - <https://v2.ocaml.org/manual/attributes.html>
> - <https://v2.ocaml.org/releases/4.12/manual/attributes.html>
> 
> so there is no way to guess the fixed-version URL from the current-version URL. Instead we could have
> 
> - <https://v2.ocaml.org/releases/latest/manual/attributes.html>
> - <https://v2.ocaml.org/releases/4.12/manual/attributes.html>
> 
> where I could easily guess that replacing "latest" by a version number will give me what I want.

dbuenzli:
> Instead we could have
> 
> * https://v2.ocaml.org/releases/latest/manual/attributes.html
> * https://v2.ocaml.org/releases/4.12/manual/attributes.html

That's not even good. Here's a user interface you will remember without effort (and these `v2` should go #465): 

* https://ocaml.org/manual -- HTTP 303 --> https://ocaml.org/manual/latest
* https://ocaml.org/manual/latest
* https://ocaml.org/manual/$(VERSION)

After all these years of efforts on the ocaml website that's something simple and obvious we could expect.

gasche:
Actually grepping through the ocaml.org sources, it looks like a "latest" schema is already supported

https://github.com/ocaml/ocaml.org/blob/0b9086770d29b45fc8a1dc69dcc580330ba83ecd/src/ocamlorg_frontend/url.ml#L22-L26

[...]

PR #1854, opened 2023-12-13
[Draft] Added HTML manuals/

shakthimaan:
As a follow-up to adding fixed versions for the OCaml manuals (https://github.com/ocaml/ocaml.org/issues/534), I have generated the HTML files for 5.1, 5.0 and 4.14 to be served under ocaml.org.


cuihtlauac:
Brillant!

shakthimaan:
The `/manuals` route now works if served via Dream!

sabine:
Since we decided that we will place the manuals into a different repository and check them out via the Dockerfile, I'm closing this PR for the time being.

(I'm trying to clear out PRs to keep a short list of PRs.)

PR #1860, opened 2023-12-17
fix a typo in the "Values and Functions" chapter

amtoine:
i think i found a typo in the [_Scopes and Environments_ section of _Values and Functions_](https://ocaml.org/docs/values-and-functions#scopes-and-environments), there is no `tau` global variable in the whole page :open_mouth: 
so i replaced it with `twenty`, which makes a bit more sense to me :yum: 

let me know if i did anything wrong, that's the first time i contribute to OCaml :relieved:


cuihtlauac:
Thanks @amtoine, this is indeed a typo. There was a `tau` in a previous version.

amtoine:
that's what i figured, happy to help :yum:

PR #1861, opened 2023-12-17
Add detailed explanation of :: operator in range function

mendesbarreto:
## Description 

Elaborated on the role of the `::` operator within the `range` function. This operator is essential for understanding how the list is built recursively. It's used to prepend the current `lo` value to the front of the list returned by the recursive call to `range`.


data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
```diff
@@ -246,7 +246,7 @@ val range : int -> int -> int list = <fun>
...
+As indicated by its type `int -> int -> int list`, the function `range` takes two integers as parameters and returns a list of integers as result. The first `int` parameter, called `lo`, is the lower bound of the range; the second `int` parameter, called `hi`, is the higher bound of the range. It is assumed that `lo <= hi`. If this isn't the case, the empty range is returned. That's the first branch of the `if … then … else` expression. Otherwise, the `lo` value is prepended at the front of a list that is going to be created by calling range itself. This is achieved by using `::` the "cons" operator in OCaml. It constructs a new list by adding an element at the front of an existing list. However, some progress is made at each call. Here, since `lo` has just been appended at the head of the list, `range` is called with the `lo + 1`. This can be visualised this way:
```
cuihtlauac:
```suggestion
As indicated by its type `int -> int -> int list`, the function `range` takes two integers as parameters and returns a list of integers as result. The first `int` parameter, called `lo`, is the lower bound of the range; the second `int` parameter, called `hi`, is the higher bound. If the case `lo <= hi`, the empty range is returned. That's the first branch of the `if … then … else` expression. Otherwise, the `lo` value is prepended to the list created by calling `range` itself; this is recursion. Preprending is achieved using `::`, the cons operator in OCaml. It constructs a new list by adding an element at the front of an existing list. Progress is made at each call, since `lo` has just been appended at the head of the list, `range` is called with `lo + 1`. This can be visualised this way:
```

The `cons` operator is already mentioned on lines 68--71. But I agree it helps to mention it here too.

cuihtlauac (changes requested):
Thanks @mendesbarreto. I agree with the intent of this PR, explaining :: at this location makes sense. 

The revised text I propose shortens a couple of other sentences too.

christinerose (approved):

data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
mendesbarreto: Thx =) 

I am new to OCaml, I felt something was missing when I encountered this peculiar operator while following this chapter.

data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
```diff
@@ -246,7 +246,7 @@ val range : int -> int -> int list = <fun>
...
+As indicated by its type `int -> int -> int list`, the function `range` takes two integers as parameters and returns a list of integers as result. The first `int` parameter, called `lo`, is the lower bound of the range; the second `int` parameter, called `hi`, is the higher bound. If the case `lo <= hi`, the empty range is returned. That's the first branch of the `if … then … else` expression. Otherwise, the `lo` value is prepended to the list created by calling `range` itself; this is recursion. Preprending is achieved using `::`, the cons operator in OCaml. It constructs a new list by adding an element at the front of an existing list. Progress is made at each call, since `lo` has just been appended at the head of the list, `range` is called with `lo + 1`. This can be visualised this way:
```
cuihtlauac:
```suggestion
As indicated by its type `int -> int -> int list`, the function `range` takes two integers as parameters and returns a list of integers as result. The first `int` parameter, `lo`, is the range's lower bound; the second `int` parameter, `hi`, is the higher bound. If `lo <= hi`, the empty range is returned. That's the first branch of the `if … then … else` expression. Otherwise, the `lo` value is prepended to the list created by calling `range` itself; this is recursion. Preprending is achieved using `::`, the cons operator in OCaml. It constructs a new list by adding an element at the front of an existing list. Progress is made at each call; since `lo` has just been appended at the head of the list, `range` is called with `lo + 1`. This can be visualised this way:
```

cuihtlauac: Some more edits.

cuihtlauac (changes requested):

data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
christinerose: Agree with these edits.

christinerose (approved):

cuihtlauac (approved):

PR #1864, opened 2023-12-18
Add is_ocaml_gui_yet page to ocaml.org

IdaraNabuk:
The page is_ocaml_gui_yet is created and will be added to ocaml.org


sabine:
Yes, this looks useful, but I think there's probably a bunch more libraries that need to be listed.

PR #1887, opened 2023-12-30
(doc) Fix description of range function

vshev4enko:


cuihtlauac (approved):
Thanks, @vshev4enko nicely caught

PR #1898, opened 2024-01-04
Add 2 XenServer positions again

edwintorok:
We have 2 open positions again: one senior,  one junior, and most of our codebase is in OCaml.


sabine (approved):
That's great! ✨

PR #1909, opened 2024-01-10
:memo: mention record update syntax

### LINKED ISSUE #1842: Mention record update `with` syntax
Record update syntax `{ my_record with field_1 = "abc"; field_2 = "def" }` needs to be explained.

- should be addressed in https://github.com/ocaml/ocaml.org/blob/main/data/tutorials/language/0it_01_basic_datatypes.md where records are introduced

srj31:
closes #1842 
Included a simple example of using the record update syntax

### DIFF
--- a/data/tutorials/language/0it_01_basic_datatypes.md
+++ b/data/tutorials/language/0it_01_basic_datatypes.md
@@ -705,6 +705,16 @@ val ghorghor_bey : character =
 - : int = 17
 ```
 
+To construct a new record with different fields without typing in the same fields again we can use record update syntax as shown:
+```ocaml
+# let togrev  = { ghorghor_bey with name = "Togrev"; level = 20; armor_class = -6 };;
+val togrev : character =
+  {name = "Togrev"; level = 20; race = "half-ogre"; class_type = Fighter;
+   alignment = (Chaotic, R_Neutral); armor_class = -6}
+```
+
+
+
 Note that records behave like single constructor variants. That allows pattern matching on them.
 ```ocaml
 # match ghorghor_bey with { level; _ } -> level;;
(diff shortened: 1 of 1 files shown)


data/tutorials/language/0it_01_basic_datatypes.md
```diff
@@ -705,6 +705,14 @@ val ghorghor_bey : character =
 - : int = 17
 ```
 
+To construct a new record with different fields without typing in the same fields again we can use record update syntax as shown:
```
cuihtlauac:
```suggestion
To construct a new record with some field values changed without typing in the unchanged fields we can use record update syntax as shown:
```

cuihtlauac (changes requested):
Half-ogre unite!

cuihtlauac (approved):

data/tutorials/language/0it_01_basic_datatypes.md
```diff
@@ -705,6 +705,14 @@ val ghorghor_bey : character =
 - : int = 17
 ```
 
+To construct a new record with some field values changed without typing in the unchanged fields we can use record update syntax as shown:
```
christinerose, line 708:
```suggestion
We can use record update syntax to construct a new record. This is useful when you need to change some field values without typing in the unchanged fields, as shown below:
```

Do you think this flows better (and is it still accurate)?
Also, is record update a type of syntax, so perhaps _record update_ syntax?

christinerose (approved):
If you don't think my rewrite is more clear, you can still merge it without changing.

data/tutorials/language/0it_01_basic_datatypes.md
sabine: I think saying that we construct a new record first is better because the effect is that there's a new record. When we mention "record update syntax" first, people may think that we are changing the existing value.

sabine:
@srj31 Thanks for the contribution! :sparkles:

PR #1914, opened 2024-01-11
Fix Getting Started documentation

akindofyoga:
I followed the tutorial and did not see any "Entering directory" messages printed from dune. I was confused this. It seems like these docs would be better without these "Entering directory" lines.


cuihtlauac:
Good point @akindofyoga; these messages appeared in previous versions of Dune. Thanks a lot

cuihtlauac:
@akindofyoga: can you update the Dune version required to the version you are using?

akindofyoga:
@cuihtlauac No problem! I have just done this.

cuihtlauac (approved):

PR #1923, opened 2024-01-12
Use a filename different from the library name in "Your First OCaml Program"

### LINKED ISSUE #1856: Dune behaviour confusing in “Your First OCaml Program”
Tutorial does not explain how to create more than exactly one file named `hello.ml` in folder `lib`. This is confusing to some. See:

https://discuss.ocaml.org/t/using-modules-15-characters-long-now/

Add small text or link to other doc to avoid having learners stuck

cuihtlauac:
This discussion: https://discuss.ocaml.org/t/using-modules-15-characters-long-now has shown that the example project in "Your First OCaml Program" suffers from having a file called `lib/hello.ml`. Since it has the same name as the library defined in `lib/dune` (i.e. hello), it turns other file added in the `lib` folder to be private modules, which is confusing for (clever) newbies.

This patch attemps to resolve this issue the following way
- The file is now called `lib/world.ml` and `lib/world.mli`
- Constant exposed in `World` module is called `v`
- Gives rise to modules `Hello` and `Hello.World`

Drawback of this solution is exposure of Dune's wrapping mechanism.

Fix #1856


sabine:
I think exposing dune's wrapping mechanism is probably a good thing.

Should we mention that if you create a module that's named the same as the library, this will make all the other modules inaccessible unless you expose them directly? (Like, have a sentence that mention this and links into the upcoming dune libraries tutorial. Probably as a comment right now.)

cuihtlauac:
> There is a special case, if you call the file `lib/hello.ml`, then refer to the modules/dune tutorial

christinerose (approved):

PR #1927, opened 2024-01-12
fix: Prevent docs link for packages without docs

### LINKED ISSUE #1576: On the package search results page: Show a "docs" link only when the package has documentation
# Task

Currently, every search result on the packages search results page (e.g. https://ocaml.org/packages/search?q=sundial) has a "Documentation" link, but not every package has documentation.

Screenshot 2023-10-05 at 10-41-06 OCaml Packages · Search Result

Instead, when a package has no documentation, no documentation link should be rendered.

How to solve this issue:
1. Look at `handler.ml` for an example how to find out if a package has documentation: https://github.com/ocaml/ocaml.org/blob/fae1912c1882333b7f5d1a778f6056c23bf46f65/src/ocamlorg_web/lib/handler.ml#L731-L741
2. modify the `packages_search` function in `handler.ml` and the `packages_search.eml` template to omit the "Documentation" link when there is no documentation.

kiyov09:
This PR attempt to fix issue #1576 

Edit: Resolves #1576


kiyov09:
@sabine 
Here's a screenshot of the result of the search using the code on this PR 
Screenshot 2024-01-12 at 12 28 11

Can you pls take a look to the code?

kiyov09:
A couple of questions

1. Should I use the same icon + text that's being used in the package page when no docs are available? See screenshot below
Screenshot 2024-01-12 at 12 37 18

2. I noticed that, while the user is typing in the search bar, the search suggestions/results shown in the dropdown also include links to the docs, even if the package doesn't have it. Would you like me to fix that too? (I'm willing to)

Packages page
Screenshot 2024-01-12 at 12 36 12

Main bar
Screenshot 2024-01-12 at 12 35 09

sabine:
Yes, it makes sense to fix it on the search results dropdown as well!

On the search results page, I would go for the grey text color for the "No documentation" notice. I think the class is `text-content`.

For the dropdown, I would grey out the docs and maybe try strike through. 🤔

kiyov09:
Got it. I'll let you know once it's done. All good with the code so far?

src/ocamlorg_frontend/pages/packages_search.eml
```diff
@@ -56,13 +57,17 @@ let render ~total ~search ~page ~number_of_pages (packages : Package.package lis
...
+                <% (match doc_status with | Package.Success -> %>
+                  <a
+                    href="<%s Url.Package.documentation package.name %>"
+                    class="text-primary flex gap-1 text-sm items-center"
+                  >
+                    <%s! Icons.documentation "h-5 w-5" %>
+                    Documentation
+                  </a>
+                <% | _ -> %>
+                  <p class="text-primary text-sm"> No Documentation Available </p>
```
kiyov09: I'll change this to `No documentation`

kiyov09:
Here we go

Screenshot 2024-01-12 at 13 07 16

src/ocamlorg_web/lib/handler.ml
```diff
@@ -637,7 +637,26 @@ let packages_search t req =
...
+    | Some { failed = false; _ } -> Ocamlorg_frontend.Package.Success
+    | Some { failed = true; _ } -> Failure
+    | None -> Unknown
+  in
+
+  let open Lwt.Syntax in
+  let* results =
+    Lwt_list.map_p
+      (fun pkg ->
+        documentation_status pkg >|= fun doc_status ->
```
```suggestion
        documentation_status pkg >|= fun doc_status ->
```

Maybe this can be written without the `>|=` operator. We try to keep things very basic, preferring let* and let+ over these operators.

sabine:
It seems like an auto formatter ran over the template, which makes the diff hard to see. Would be helpful to have a smaller diff. :slightly_smiling_face:

kiyov09:
☹️ maybe because I ran `make fmt`

sabine:
`make fmt` only affects the `.ml` files and is intended to be run. It must be some other autoformatter, since ocamlformat doesn't currently format the templates. :thinking: 

If you find out what it was, maybe we can include a config file that prevents it in the repo (since this is a common problem).

kiyov09:
yeah, I'm noticing that. I'm trying to revert the formating and determine what it was.

kiyov09:
Found it. 

Because the 2nd commit was a really small change I opened it in nvim and, without changing the file type to `ocaml`, I made the change and saved the file. That caused the wrong formating.

By default, my nvim is setting the file type for `.eml` files to `mail`. I'm going to update my setup right away to automatically set the ft of eml files to ocaml to prevent this from happening in the future, but I don't know how it can be prevented for everybody.

src/ocamlorg_web/lib/handler.ml
kiyov09: got it, I'll try to do it without the Infix.

kiyov09:
Fixed the template.

The problem was that the wrong file type made each line end with a dos style (\r\n). I followed [this StackExchange thread](https://superuser.com/questions/395439/m-in-many-lines-in-source-code-files-how-can-i-remove-them) to fix it.

kiyov09:
Ok, here are two possible ways of signaling the package has no doc in the autocomplete. Which one do you like better? I'd go with the one where the link is hidden.

line-through
Screenshot 2024-01-12 at 16 10 30

hidden
Screenshot 2024-01-12 at 16 10 05

kiyov09:
also, if you want, we can later improve the style of the doc link to make it look like the one that's presented in the search results, which I think looks a lot better.

Screenshot 2024-01-12 at 16 15 56

It can be another issue/pr tho

kiyov09:
Ok, I want with the hidden option, but just for completing the commit. I can change it to the line-through variant.

src/ocamlorg_web/lib/handler.ml
```diff
@@ -649,12 +670,31 @@ let packages_autocomplete_fragment t req =
...
+          | Some { failed = true; _ } -> Failure
+          | None -> Unknown)
+      in
+
+      let* top_5 =
+        packages |> List.take 5
+        |> Lwt_list.map_p (fun pkg ->
+               let+ doc_status = documentation_status pkg in
+               (Package_helper.frontend_package t pkg, doc_status))
+      in
```
kiyov09: This and the version in the ```packages_search``` function above are almost identical. I want to remove that duplication but don't know where's the best place (module/file) to put the common logic.

kiyov09:
Also, my solution works but I don't think it's the best. If I'm not mistaken, the code behind builds the URL for the documents and makes a request each time, so I think we should find a way to store somewhere the results of knowing whether a package has documents or not.

I can go into more depth on that if you think it's something that needs to be addressed now.

src/ocamlorg_web/lib/handler.ml
sabine: There's a module `Package_helper` inside `handler.ml` that holds some similar logic of translating between the `Ocamlorg_package` notion of what a package is and what a package looks to `Ocamlorg_frontend.Package`.

sabine:
> If I'm not mistaken, the code behind builds the URL for the documents and makes a request each time

Yes, that's correct.

Considering that we render up to 50 search results at a time, 50 http requests every time seems scary.

Indeed, we need to introduce a cache for the documentation status - the best location for that is most likely inside `Ocamlorg_package`, since that's where we're making the http request to check if documentation exists.

If you're up for it, go for it! Otherwise, I'll have a look next week. :slightly_smiling_face:

kiyov09:
Got it. I'll go for it, but not sure if during the weekend.

kiyov09:
Before start coding anything and go down the wrong rabbit hole I want to discuss my ideas.

As far as I saw in the codebase, the info of the packages that is used to satisfy searches is extracted from Opam on init and saved in the state (and polled and updated after X seconds if polling is not disabeld

kiyov09:
Before start coding anything and go down the wrong rabbit hole I want to discuss my ideas.

As far as I saw in the codebase, the info of the packages that is used to satisfy searches is extracted from Opam on init and saved in the state (and polled and updated after X seconds if polling is not disabled). Let me know if my understanding is wrong here.

So, as we want to cache the doc status somehow I think we could use a similar approach. My two ideas so far are:

1. Cache the result of the calls to `documentation_status`, effectively making just the first call pay the cost of requesting the info. The downside is that till the cache is "populated" will have a pretty large amount of docs requests per search.

2. Spawn (somehow) a task to update the documentation status info of all packages after each opam info update. The downside is that we'll potentially ask for updates on a lot of packages that haven't changed. Maybe there's a way to narrow the ones we want to "update".

What do you think? Would you go in another direction?

sabine:
Hey @kiyov09, thanks for the patience, I am just seeing this now.

Yes, I think you're right, there are these two options.

Option 2 would have to be done upstream on the documentation pipeline at https://github.com/ocurrent/ocaml-docs-ci, since the documentation status is independent from opam-repository. It's not very clear to me right now where exactly to do this. I suppose that documentation status could be emitted on a per-package basis there on a recurring basis. Then we could periodically pull in and refresh this data, similar to how opam-repository is pulled into OCaml.org to generate the package info.

Option 1 (Caching) would be feasible to do on this repository, and would alleviate the problem with having a lot of HTTP requests for now, but is indeed likely to be replaced by option 2 in the long term.

kiyov09:
Go it. I'll start working on an implementation for option 1. Thnx

kiyov09:
Hi @sabine
I have a first draft of the caching strategy. It's not the final solution yet but a starting point.
When you have time, please take a look.

kiyov09:
maybe is a better idea to make the "doc status" available as part of the package info instead of having another field?

src/ocamlorg_package/lib/ocamlorg_package.ml
```diff
@@ -375,28 +380,32 @@ let search_index ~kind t =
...
+  | Some status -> Lwt.return status
+  | None ->
+      let+ content = http_get url in
+      let status =
+        match content with
+        | Ok s ->
+            Some (s |> Yojson.Safe.from_string |> Documentation_status.of_yojson)
+        | _ -> None
+      in
+      state.doc_stats <- DocumentationStatusMap.add url status state.doc_stats;
```
sabine: I see, this caches by URL. I suppose that caching by `Ocamlorg_package.Name.t` and `Version.t` would probably be more obvious.

src/ocamlorg_web/lib/handler.ml
```diff
@@ -637,7 +637,28 @@ let packages_search t req =
...
+  in
+
+  let* results =
+    Lwt_list.map_p
+      (fun pkg ->
+        let+ doc_status = documentation_status pkg in
+        (Package_helper.frontend_package t pkg, doc_status))
+      current_items
+  in
+
```
sabine: We should probably make documentation status part of `Ocamlorg_frontend.Package` so that `Package_helper.frontend_package` can deal with the lookup of documentation status. :thinking:

sabine (commented):
@kiyov09 thanks for your patience, I got sick on Sunday, better now and I'm really behind on things - catching up.

Package info is the data we pull from opam-repository. In the existing code, this is badly named, I think the module would be better named as `opam_repository_info.ml`, rather than `info.ml`.

Documentation status and the actual documentation pages are coming from a different source (docs.ci.ocaml.org). So we could probably extend `Ocamlorg_package.t` to have a mutable field `documentation_status`.

Since the documentation status could change at runtime, e.g. when documentation that previously failed to build or didn't exist yet has been built successfully, we'd just need to expire the cached values sometimes - but we should probably only refetch when someone opens the package overview page, even if that doesn't guarantee 100% consistency on the search results pages.

Reading your code changes here helped me a a lot in figuring out where and how we could do this! :sparkles: Thanks!

src/ocamlorg_package/lib/ocamlorg_package.ml
kiyov09: like in the `packages` field, right?

src/ocamlorg_web/lib/handler.ml
kiyov09: I'm not sure I get this. Do you mean to move this function to the `Ocamlorg_frontend.Package`?

src/ocamlorg_package/lib/ocamlorg_package.ml
kiyov09: I did it via the URL because we need to take the `kind` also into account, and the URL already encodes that info. Do you have an idea/example of that?

kiyov09:
Hi @sabine , glad you're feeling better.

Your idea makes sense to me and I think I like it better too. Summarizing, and making sure I understand it right:

1. Instead of having a mapping storing (potentially) all documentation statuses,  each package (`Ocamlorg_package.t`) will cache its documentation status.
2. Once the documentation status has been assigned to the `Ocamlorg_package.t`, it'll remain the same (and potentially stale) until a user accesses the package overview page.

Is my understanding correct?

sabine:
Yes, that sounds good to me.

I feel like I have to go to the docs pipeline and put the documentation status into the sqlite database that already exists there, so that we can then query docs status of multiple packages/versions in a single http request.

kiyov09:
Ok, so I'll be working on that and ping you when it's ready. Thnx.

The sqlite thing is for another iteration, right?

sabine:
Yeah sqlite would be upstream, but that codebase isn't currently in a shape
that we can make that change.

kiyov09:
> Hi @sabine , glad you're feeling better.
> 
> Your idea makes sense to me and I think I like it better too. Summarizing, and making sure I understand it right:
> 
>     1. Instead of having a mapping storing (potentially) all documentation statuses,  each package (`Ocamlorg_package.t`) will cache its documentation status.
> 
>     2. Once the documentation status has been assigned to the `Ocamlorg_package.t`, it'll remain the same (and potentially stale) until a user accesses the package overview page.
> 
> 
> Is my understanding correct?

Hi @sabine . 

I'm diving into the code again and I think what i commented here won't work by its own.

As long as I can find (pls correct me if I'm wrong) all instances of `Ocamlorg_package.t` are constructed from the info stored inside `Ocamlorg_package.state.packages` so I'm pretty sure we need to:

- Update the definition of `Ocamlorg_packages.state.packages` to store the doc status along side the `Info.t` ( I think this is expensive for I'm able to understand in the code)
- Use something similar to what I have so far to store the doc status of each package and populate the new field on `Ocamlorg_package.t` from there.

What do you think?

kiyov09:
If we choose opt 2 we can keep the new map I created (maybe modifying it to depend on name and version instead of URL) and simplify the code on  the `handler.ml` file

sabine:
I think your option 2 is entirely fine. :+1:

kiyov09:
Hi @sabine, two things

First, I updated the `doc_status` map to depend on `Name.t` and `Version.t`, and everything keeps working fine from what I see. But I noticed the `Documentation_status.t` depends on the `kind` of the package we are dealing with, but I don't know how to encode that in the map.

Second, I'm having a hard time trying to extend `Ocamlord_package.t` with a new field. It's used in so many places, not all of them with the same context. Ex: some function has access to the `Ocamlorg_package.state` but others only to the `Ocamlorg_package.state.package`. I don't have to much experience with Ocaml and would appreciate the help.

kiyov09:
btw, I identified a couple of potential changes that could improve the perceived performance of the search autocomplete. Should I open a new PR to show them to you? or do we need to open a new issue?

src/ocamlorg_package/lib/ocamlorg_package.ml
```diff
@@ -12,11 +12,14 @@ let version t = t.version
 let info t = t.info
 let create ~name ~version info = { name; version; info }
 
+module DocumentationStatusMap = Map.Make (String)
+
```
sabine:
```suggestion
```

seems obsolete now that we're using `Documentation_status.t option Version.Map.t Name.Map.t`

sabine (commented):
> btw, I identified a couple of potential changes that could improve the perceived performance of the search autocomplete. Should I open a new PR to show them to you

Opening a PR on this is great. :sparkles:

> First, I updated the doc_status map to depend on Name.t and Version.t, and everything keeps working fine from what I see

Yup, looks good to me.

> But I noticed the Documentation_status.t depends on the kind of the package we are dealing with, but I don't know how to encode that in the map.

We only need to cache documentation status for ``kind=`Package``, so what you did here is perfectly fine. 

To explain a bit: The `kind` distinguishes between package docs that show up as search results (kind is `` `Package``) and package docs that we only keep around because they're a dependency of some other package (kind is `` `Universe``). The reason this is done is that, due to Module imports and Functors, package docs can be different depending on which exact set of dependencies was used. When this was implemented, people decided it is important to have full consistency of the docs of a package.

> Second, I'm having a hard time trying to extend Ocamlord_package.t with a new field. It's used in so many places, not all of them with the same context. Ex: some function has access to the Ocamlorg_package.state but others only to the Ocamlorg_package.state.package.

Good point. While the compiler will make sure we can't forget any place to update, it's not entirely clear to me right now how this will play out in terms of when the lookup of documentation status happens.

So I think the best thing to do here now is to only make the (optional) documentation status part of `Ocamlorg_frontend.Package.package`:

```
type documentation_status = Success {
  readme_filename : string option;
  license_filename : string option;
  changes_filename : string option;
} | Failure | Unknown

type package = {
  [...]
  documentation_status: documentation_status;
}
```


and refactor `Package_helper.frontend_package` to take the documentation status as an optional parameter in the file `handler.ml`:

```
let frontend_package ?on_latest_url ?documentation_status state
  (package : Ocamlorg_package.t) : Ocamlorg_frontend.Package.package =
```
And then it's in a place where it's good to merge.

src/ocamlorg_package/lib/ocamlorg_package.ml
kiyov09: yeah, need to remove that one.

kiyov09:
Awesome. I'll work on that today in the night (hopefully).

Thanks a lot for the detailed explanation.

kiyov09:
One last thing. We talked previously about re-fetching the status when the user enters the package page. Should I still implement that?

sabine:
Yes, we need to refetch occasionally because otherwise new docs will not show up.

kiyov09:
what about storing the "datetime" along the doc status in the new map and using the "cache info" only it if was fetched X time ago, forcing a refetch otherwise?

Or we can set up an "interval" like the one that is already in place for the packages.

sabine:
Probably good to store the datetime because the cache will not expire at the same time for every package.

kiyov09:
Cool. I'll work on that too and we can review it later. Thnx

kiyov09:
Hi @sabine , thanks for your patience.

I think I'm done with the necessary changes (although I must fix the conflicts this branch has with `ocaml:main` ).

> 
> ```
> type documentation_status = Success {
>   readme_filename : string option;
>   license_filename : string option;
>   changes_filename : string option;
> } | Failure | Unknown
> 
> type package = {
>   [...]
>   documentation_status: documentation_status;
> }
> ```

I didn't add the bunch of fields to the Success variants 'cause the `Package_overview.sidebar_data` includes them all plus the documentation status, so I thought it would be a little confusing. I can refactor all that if you want.

Let me know what you think about the changes and then I can fix the conflict and prepare the PR to be merged.

sabine:
Hey @kiyov09, that looks mergeable to me. Yes, I'll probably refactor more after that, but I also don't want this to be an endless journey, since this is fine.

Do you want me to do the rebase or resolve conflicts on your own? I would probably just squash all this and then rebase on main.

kiyov09:
Awesome!

About the conflicts, any option work for me. Whatever happens first 😅

kiyov09:
@sabine conflicts solved.

sabine (approved):
Awesome! Thanks for this @kiyov09! :sparkles: :+1: 

Merging this now.

Issue #1589, opened 2023-10-06
Add Capability to Recommend Follow-Up Tutorials on Tutorials


sabine:
We would love to have an external contributor work on this issue.

# Task

It would be helpful to recommend to people at the end of a tutorial further tutorials that we recommend them to read after having finished the current one.

To achieve that, we need the ability for tutorials to reference the tutorials that we recommend them to read next.

1. You need to add a field `recommended_next_tutorials: string list` to both the `metadata` and `t` types in `tutorial.ml`. which holds the slugs of all the tutorials we recommend reading after this one.
2. Add some sensible references between tutorials in the `data/tutorials` folder, for testing purposes.
3. In `handler.ml`, in the function `tutorial`, you need to filter the tutorials for those that should be recommended as next reading, i.e. (`slug` is an element of `tutorial.recommended_next_tutorials`).
4. In `tutorial.eml`, render a plain list of links to the recommended tutorials at the bottom of the page.

**Important:** When a referenced tutorial slug does not exist, an error should be thrown by `ood-gen` so that we can be sure that the references between tutorials will stay in a good shape.

iamkalio:
Hello @sabine @SaySayo , I'm interested  and would like to work on this task. Could you please assign it to me? Thank you
I've created a pull-request for the first task while its on review  I can start working on this if assigned. thanks

sabine:
@kalio007 You are now assigned to this issue. Have fun! :slightly_smiling_face:

sabine:
@kalio007 are you working on this issue? Please confirm if you still plan to work on this issue.

You can open a draft pull request if you need feedback / help.

iamkalio:
@sabine I did have a couple issues, but ill create a draft pull request now. I've been going through GUI projects to understand them hence the delay

PoorlyDefinedBehaviour:
I'll do it if no one else wants to (since I have already done another one on Dune)

enoonan:
I started hacking on this but I think there's something I don't understand about these `@@deriving` annotations. I'm assuming it works similarly to `derive` macros in Rust but that's about as far as I go.

After step 1, the error I get on compile is `Didn't find the function for key: recommended_next_tutorials` which I can just grep for and see it comes from the `ppx_deriving_yaml` library. So I tried extracting recommended_next_tutorials into a type like so:

```
type recommended_next_tutorials = string list
[@@deriving of_yaml, show { with_path = false }]
```

[...]

PR #1928, opened 2024-01-12
(feat) add recommended_next_tutorials capability

enoonan:
Adds ability to recommend next tutorial in the head / metadata of an existing tutorial.md file. E.g.:

`recommended_next_tutorials: 
  [
    "values-and-functions",
    "basic-data-types",
    "if-statements-and-loops",
    "lists"
  ]`

Examples have been added to "A Tour of OCaml" and "Your First Ocaml Program"

### DIFF
--- a/data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
+++ b/data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
@@ -4,6 +4,13 @@ title: A Tour of OCaml
 description: >
   Hop on the OCaml sightseeing bus. This absolute beginner tutorial will drive you through the marvels and wonders of OCaml. We'll have a look at the most commonly used language features.
 category: "First Steps"
+recommended_next_tutorials: 
+  [
+    "values-and-functions",
+    "basic-data-types",
+    "if-statements-and-loops",
+    "lists"
+  ]
 ---
 
 # A Tour of OCaml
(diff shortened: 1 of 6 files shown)


enoonan:
Addresses issue https://github.com/ocaml/ocaml.org/issues/1589

christinerose:
This LGTM on my end. @cuihtlauac ?

christinerose (approved):

sabine (approved):
Thanks @enoonan, this looks great! I have no change requests here! :sparkles: Merging.

The only thing we could do to make this more foolproof (for people editing the docs) is if we check that the documents in question exist. But I think that should probably be done in another PR, since there's more than just the tutorials that would benefit from this.

sabine:
I noticed, the CI will complain because of formatting. Running `make fmt` once helps.

enoonan:
Yay! I just ran `make fmt` and pushed that so we'll see whether that does the trick :)

enoonan:
Re: checking that documents exist @sabine - where would that check happen? I'm still a little unclear on when and where things execute vis a vis the code gen that's happening.

sabine:
Yep, in tool/ood-gen in tutorial.ml is where the codegen happens, and the codegen is probably the right moment to check if the recommended tutorials exist. Probably after all the tutorials have been decoded and before saving them as an OCaml module.

enoonan:
Would you prefer that to be a separate PR? I'm happy to give it a go but am a little nervous to roll it into this one just because I think I'll be doing a lot of trial and error familiarizing myself with the code gen part of it.

sabine:
Sure, let's merge this PR and if you run into trouble, it's perfectly fine to ask more questions!

PR #1943, opened 2024-01-15
Add pre-compile check for Recommended Next Tutorials

enoonan:
Not totally sure this is the approach you all prefer but I told @sabine I would take a crack at it. 

If someone fills out the "recommended_next_tutorials" portion of a tutorial's metadata, and they provide a tutorial slug that doesn't exist, this PR will cause compilation to fail with a helpful "Missing_Tutorial" exception and error message. 

Currently a misspelled tutorial slug will just not appear in the final page.

I also removed the `related-exercises` class from the rendered "Recommended next tutorials" output - it wasn't doing anything and I only copied it in there accidentally in the first place!


sabine (commented):
The approach looks great to me, doing the simplest thing here makes sense!

Having an error message is very helpful. Does it show in which document the error occurs? (I don't remember if we have that for these documents.)

I think there's some possibility to simplify the logic (List.exists three times seems suspicious). Will take a closer look later.

enoonan:
> Having an error message is very helpful. Does it show in which document the error occurs?
It doesn't but it could

> I think there's some possibility to simplify the logic (List.exists three times seems suspicious).
Agreed!

sabine:
Yup, so from a perspective of adding or changing data, making the filename part of the error message would be excellent, I think.

Regarding code simplifications, we're happy to do those.

enoonan:
Added a more informative error message and took a shot at simplifying the logic. It's made especially easier by the fact that `recommended_next_tutorials` got changed from an option to a string list. Also I spent some time reading the List module docs / letting it simmer / learning new tricks.

PS apologies for so many commits. It has been a long time since I worked on a large collaborative project and I keep messing up my gits / needing to tweak things

tool/ood-gen/lib/tutorial.ml
```diff
@@ -120,6 +120,23 @@ let toc ?(start_level = 1) ?(max_level = 2) doc =
     invalid_arg "toc: ~max_level must be >= ~start_level";
   create_toc ~max_level start_level (headers doc)
 
+exception Missing_Tutorial of string
+
+let any_recommended_next_tuts_are_missing_exn all =
```
cuihtlauac: PR #1965 will add prerequisites tutorials. Those list will require the same check. Let's name this accordingly.
```suggestion
let check_tutorial_references all =
```

tool/ood-gen/lib/tutorial.ml
```diff
@@ -138,6 +155,8 @@ let all () =
   |> List.sort (fun t1 t2 -> String.compare t1.fpath t2.fpath)
 
 let template () =
+  let _ = any_recommended_next_tuts_are_missing_exn @@ all () in
```
cuihtlauac:
```suggestion
  let _ = check_tutorial_references @@ all () in
```

cuihtlauac (changes requested):

cuihtlauac (approved):
Thanks @enoonan this looks good to me. This will be useful to maintain data consistency.

data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
```diff
@@ -406,8 +403,7 @@ This function operates not just on lists of integers but on any kind of list. It
 
 #### Defining a Higher-Order Function
 
-It is possible to pass a function as a parameter to another function. Functions taking other functions as parameters are called _higher-order_ functions. This was illustrated earlier using function `List.map`. Here is how `map` can be written using pattern matching on lists.
-
+It is possible to pass a function as argument to another function. Functions having other functions as parameters are called _higher-order_ functions. This was illustrated earlier using function `List.map`. Here is how `map` can be written using pattern matching on lists.
```
christinerose:
```suggestion
It is possible to pass a function as argument to another function. Functions that have other functions as parameters are called _higher-order_ functions. This was illustrated earlier using function `List.map`. Here is how `map` can be written using pattern matching on lists.
```

data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
```diff
@@ -187,15 +186,13 @@ We can write anonymous functions and immediately apply them to a value:
...
-
+A function may have several parameters, separated by spaces.
 ```ocaml
 # let cat a b = a ^ " " ^ b;;
 val cat : string -> string -> string = <fun>
 ```
 
-The function `cat` takes two `string` arguments, `a` and `b`, and returns a value of type `string`.
-
+The function `cat` has two `string` parameters, `a` and `b`, and returns a value of type `string`.
```
christinerose: Perhaps we should change this example away from "a" and "b" as we've been doing to avoid confusion with `'a` and `b`.

enoonan: Christine I think you might be looking at an older commit. At one point I kind of screwed up my merging and then didn't realize my editor had auto-formatted that specific tutorial when I was testing things out. In the latest commit the only affected file should be tutorial.ml

christinerose: Oh! That makes sense! I thought we had caught them all. :) 
Can you disregard that or should I delete it?

enoonan: Hmmm... will it work if I just click "resolve conversation"? 

(I haven't worked on a big team project for a while and forget how this is supposed to go 😅)

christinerose: Absolutely. Just resolved it, so it should just stay the same. 🙂 
No worries about forgetting...I forget from day to day!

PR #1836, opened 2023-12-07
 Implement dark mode on learn area landing page

oyenuga17:
@SaySayo @sabine @punchagan please review.
- To test this uncomment line https://github.com/oyenuga17/ocaml.org/blob/14e93c389c0921d0ff9eaa1ae5ba2149a7862206/src/ocamlorg_frontend/layouts/layout.eml#L103 in layout.eml

PR #1957, opened 2024-01-16
Include text on if-then-else and begin-end

### LINKED ISSUE #1851: Explain `if then else`
Explain `if ... then ... else ...` wherever needed, with some redundancy (that's ok).

OCaml has both `if ... then ... else ...` and `if ... then ...` (which results in a unit value, and thus should only be mentioned in the Mutability & Imperative tutorial).

- [ ] Basic Datatypes: if then else as syntactic sugar on pattern matching over bool
- [ ] Mutability and imperative: if then w/o else
- [ ] In values & functions: if then else and if then are expressions
- [ ] Check with existing text in “Tour of OCaml” and “If & Recusion” docs

(all the files are in the folder https://github.com/ocaml/ocaml.org/tree/main/data/tutorials/language)

cuihtlauac:
Fix #1836
Fix #1851


data/tutorials/language/0it_06_imperative.md
```diff
@@ -331,6 +331,138 @@ In this example, the first two expressions are `print_endline` function calls, w
...
+
+This is arguably convoluted and does not work:
+```ocaml
+let f r = r := incr r; 2 * !r;;
+Error: This expression has type unit but an expression was expected of type int
+```
+
+But here is how it can be made to work:
+```ocaml
+# let f r = r := begin incr r; 2 * !r end;;
```
christinerose:
```suggestion
# let f r = r := begin incr r; 2 * !r end;;
val f : int ref -> unit = <fun>
```

data/tutorials/language/0it_06_imperative.md
```diff
@@ -331,6 +331,139 @@ In this example, the first two expressions are `print_endline` function calls, w
...
+val f : int ref -> unit = <fun>
+```
+
+The error came from assign `:=` associating more that semicolum `;`. Here is what we want to do, in order:
+1. Increment `r`
+2. Compute `2 * !r`
+3. Assign into `r`
+
+Remember the value of a semicolon-separated sequence is the value of its last expression. Grouping the first two steps suffice to fix the error. Usually, this is done using parenthesis. But surrounding by parenthesis or `begin … end` is the same. The latter is preferred to highlight expressions having side effects.
```
christinerose:
```suggestion
Remember the value of a semicolon-separated sequence is the value of its last expression. Grouping the first two steps fix the error. Usually, this is done using parentheses, but surrounding it with `begin … end` works, too. The latter is preferred to highlight expressions that have side effects.
```

data/tutorials/language/0it_06_imperative.md
```diff
@@ -331,6 +331,139 @@ In this example, the first two expressions are `print_endline` function calls, w
...
+
+# if false then print_endline "A" else print_endline "B"; print_endline "C";;
+B
+C
+- : unit = ()
+```
+
+Here `; print_endline "C"` is executed after the whole conditional expression, not after `print_endline "B"`.
+
+The sequence must form a single subexpression to have two prints in one branch. Parenthesis can be used, but it is customary to use `begin … end` in an imperative context.
```
christinerose:
```suggestion
The sequence must form a single subexpression to have two prints in one branch. Parentheses can be used, but it is customary to use `begin … end` in an imperative context.
```

data/tutorials/language/0it_06_imperative.md
```diff
@@ -331,6 +331,139 @@ In this example, the first two expressions are `print_endline` function calls, w
...
+
+Grouping the statements of sequence is also needed for the first branch of the conditional:
+```ocaml
+# if true then begin
+    print_endline "A";
+    print_endline "C"
+  end else
+    print_endline "B";
+  ;;
+A
```
christinerose:
```suggestion
A
C
```

data/tutorials/language/0it_06_imperative.md
```diff
@@ -331,6 +331,139 @@ In this example, the first two expressions are `print_endline` function calls, w
...
+    print_endline "A";
+    print_endline "C"
+  else
+    print_endline "B";;
+B
+C
+- : unit = ()
+```
+
+Failing to do it will result in a syntax error. What's before the semicolon is parsed as an `if … then … ` without else expression. What's after the semicolon appears as a [dangling](https://en.wikipedia.org/wiki/Dangling_else) `else`.
```
christinerose:
```suggestion
Failing to do it will result in a syntax error. What's before the semicolon is parsed as an `if … then … ` without an `else` expression. What's after the semicolon appears as a [dangling](https://en.wikipedia.org/wiki/Dangling_else) `else`.
```

data/tutorials/language/0it_06_imperative.md
```diff
@@ -331,6 +331,140 @@ In this example, the first two expressions are `print_endline` function calls, w
...
+- : unit = ()
+
+# if false then
+    print_endline "A";
+    print_endline "C"
+  else
+    print_endline "B";;
+B
+C
+- : unit = ()
```
christinerose: I get a syntax error here

christinerose (commented):
Made a few changes.
One comment open where I get a syntax error

data/tutorials/language/0it_06_imperative.md
cuihtlauac: Yes, this is on purpose. The syntax error is expected. I messed up with code blocks. The block lines 458--464 should replace the block lines 446--454. Once done, the second (and therefore redundant) block should be deleted.

data/tutorials/language/0it_06_imperative.md
```diff
@@ -331,6 +331,140 @@ In this example, the first two expressions are `print_endline` function calls, w
...
+
+Remember the value of a semicolon-separated sequence is the value of its last expression. Grouping the first two steps fix the error. Usually, this is done using parentheses, but surrounding it with `begin … end` works, too. The latter is preferred to highlight expressions that have side effects.
+
+**Fun fact**: `begin … end` and parenthesis are badly the same:
+```ocaml
+# begin end;;
+- : unit = ()
+```
+
+### Conditional Execution
```
sabine: This title is indirect, doesn't make clear that this is about If-Then-Else. From the title alone, it could also be about match expressions.

sabine (commented):
- PR lacks a redirect from `/docs/if-statements-and-recursion` to `/docs/loops`, and I am not sure why the new slug is `loops`, and not `loops-and-recursion`.

- Is it clear that moving `begin` ... `end` syntax and `if`/`then`/`else` explanations from this document to the mutable state tutorial is a good idea? I ask because I think `begin` ... `end` is very fundamental and should probably be introduced early. Both concepts are independent from mutability or imperative programing - `begin`/`end` is just a fancy way to avoid brackets, and `if`/`then`/`else` is just an expression.

I understand the old document here as being about the topic of control flow. And this seems a good topic to cover early.

I would be tempted to move the control flow related sections from the mutability tutorial, rather than the other way around.

data/tutorials/language/0it_06_imperative.md
```diff
@@ -331,6 +331,107 @@ In this example, the first two expressions are `print_endline` function calls, w
...
+```
+
+The error came from assign `:=` associating stronger that semicolon `;`. Here is what we want to do, in order:
+1. Increment `r`
+2. Compute `2 * !r`
+3. Assign into `r`
+
+Remember the value of a semicolon-separated sequence is the value of its last expression. Grouping the first two steps with `begin … end` fixes the error.
+
+**Fun fact**: `begin … end` and parentheses are literally the same:
```
christinerose: This was said on li 336. If you meant to repeat it here, then that's fine, although you might want to rephrase it to something like " **Remember: .... **

christinerose:
li 551 in imperative.md says something about "other languages." Is it necessary to have a note about other languages here?

christinerose (approved):
All LGTM, just two questions in the Conversation ....

data/tutorials/language/0it_06_imperative.md
cuihtlauac: This is repeated here because the `unit` value, usually written `()` can also be written `begin end`. Useless by funny

cuihtlauac:
> li 551 in imperative.md says something about "other languages." Is it necessary to have a note about other languages here?

Yes, this is intended

data/tutorials/language/0it_06_imperative.md
```diff
@@ -331,6 +331,107 @@ In this example, the first two expressions are `print_endline` function calls, w
...
+Error: This expression has type unit but an expression was expected of type int
+```
+
+But here is how it can be made to work:
+```ocaml
+# let f r = r := begin incr r; 2 * !r end;;
+val f : int ref -> unit = <fun>
+```
+
+The error came from assign `:=` associating stronger that semicolon `;`. Here is what we want to do, in order:
```
cuihtlauac:
```suggestion
The error came from assign `:=`, which associates stronger than a semicolon `;`. Here is what we want to do, in order:
```

PR #1961, opened 2024-01-16
(data) Added priver.dev ocaml feed

emilpriver:
hello,
I created a ocaml rss feed with all my ocaml posts and added it to planet-source.yml.

Hope it's interesting :D

### DIFF
--- a/data/planet-sources.yml
+++ b/data/planet-sources.yml
@@ -182,4 +182,4 @@
   url: https://blog.osau.re/feed.xml
 - id: emilpriver
   name: Emil Privér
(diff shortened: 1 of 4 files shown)


data/planet-sources.yml
```diff
@@ -180,3 +180,6 @@
 - id: dinosaure
   name: Romain Calascibetta
   url: https://blog.osau.re/feed.xml
+- id: emilpriver
+  name: Emil Privér
+  urk: https://priver.dev/tags/ocaml/index.xml
```
sabine:
```suggestion
  url: https://priver.dev/tags/ocaml/index.xml
```

PR #1978, opened 2024-01-20
(docs) Include as syntax for function parameters

### LINKED ISSUE #1852: Explain `as` Syntax for Aliasing in Pattern Matching
OCaml provides Syntax for defining an alias for some parameters you pattern match on.

For example:

```
let f ((x, y) as arg) = ...

let f ({a;b} as arg2) = ...

let f = function Some _ as opt -> opt | None -> Some 42;;
```

So, it's possible to both pattern match on a function parameter and give the parameter a name.

Relevant documents where this needs to be explained:
- https://github.com/ocaml/ocaml.org/blob/main/data/tutorials/language/0it_00_values_functions.md
- https://github.com/ocaml/ocaml.org/blob/main/data/tutorials/language/0it_01_basic_datatypes.md
- https://v2.ocaml.org/releases/5.1/htmlman/patterns.html#sss:pat-alias (Alias patterns)

srj31:
closes https://github.com/ocaml/ocaml.org/issues/1852

- Added use of pattern matching in function parameters
- Mention Record type for function parameters
- Function Parameter Alias for tuples and records
- Use alias in variant values for parameter

### DIFF
--- a/data/tutorials/language/0it_00_values_functions.md
+++ b/data/tutorials/language/0it_00_values_functions.md
@@ -601,6 +601,7 @@ val spicy_cat : string * string -> string = <fun>
 # spicy_cat ("hello", "world");;
 - : string = "hello world"
 ```
+
 It looks like two arguments have been passed: `"hello"` and `"world"`. However, only one, the `("hello", "world")` tuple, has been passed. Inspection of the generated assembly would show it isn't the same function as `sweet_cat`. It contains some more code. The contents of the tuple passed to `spicy_cat` (`x` and `y`) must be extracted before evaluation of the `x ^ " " ^ y` expression. This is the role of the additional assembly instructions.
 
 In many imperative languages, the `spicy_cat ("hello", "world")` syntax reads as a function call with two parameters; but in OCaml, it denotes applying the function `spicy_cat` to a tuple containing `"hello"` and `"world"`.
@@ -751,3 +752,4 @@ One may wonder:
 
 The answer to this question goes beyond the scope of this tutorial. This comes from the [λ-calculus](https://en.wikipedia.org/wiki/Lambda_calculus), the mathematical theory underneath functional programming. In that formalism, there are nothing but functions. Everything, including data, is a function, and computation reduces to parameter passing. In functional programming (and thus OCaml), having functions and values at the same level is an invitation to think this way. This is different from the imperative programming approach where everything reduces to reading and writing into the memory.
 -->
+
--- a/data/tutorials/language/0it_00_values_functions.md
+++ b/data/tutorials/language/0it_00_values_functions.md
@@ -180,6 +180,19 @@ This also works with user-defined types.
 # type live_person = int * name;;
 type live_person = int * name
 
+# let robin_with_age : live_person = (20, robin);;
+val robin_with_age : live_person = (20, {first = "Robin"; last = "Milner"})
+
+# let (years, { first; last }) = robin_with_age;;
+val years : int = 20
+val first : string = "Robin"
+val last : string = "Milner"
+```
+
+### Pattern Matching on Function Parameters
+
+Pattern matching can also be used on function parameters.
+```ocaml
 # let age ((years, { first; last }) : live_person) = years;;
 val age : live_person -> int = <fun>
 ```
--- a/data/tutorials/language/0it_00_values_functions.md
+++ b/data/tutorials/language/0it_00_values_functions.md
@@ -156,6 +156,23 @@ val first : string = "Robin"
 val last : string = "Milner"
 ```
 
+### Pattern Matching in Function Parameters
+
+It is possible to use pattern matching to specify function parameters. For tuples:
+```ocaml
+# type live_person = int * name;;
+type live_person = int * name
+
+# let age ((years, full_name) : live_person) = years;;
+val age : live_person -> int = <fun>
+```
+
+For records:
+```ocaml
+# let introduce ({first; last}) : name) = "I am " ^ first ^ " " ^ last;;
+val introduce : name -> string = <fun>
+```
+
 ### Pattern Matching on `unit`
 <!--Unit example-->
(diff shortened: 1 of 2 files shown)


data/tutorials/language/0it_00_values_functions.md
```diff
@@ -593,6 +606,20 @@ It looks like two arguments have been passed: `"hello"` and `"world"`. However,
...
+# type cat_args = {x : string; y : string};;
+type cat_args = { x : string; y : string; }
+
+# let happy_cat ({x; y}: cat_args) = x ^ " " ^ y;;
+val happy_cat : cat_args -> string = <fun>
+
+# happy_cat {x = "hello"; y = "world"};;
+- : string = "hello world
+```
+
```
cuihtlauac: Adding this section is a good idea. I propose
- Move it right after "pattern matching on records"
- Use the `age` type
- Use the title "Pattern Matching in Function Parameters" 
- Add an example with a tuple
- Update text accordingly

cuihtlauac (changes requested):
The section "Pattern Matching in Definitions" addresses all the cases where not exactly one name is defined using a `let` or a parameter definition (maybe we should state that explicitly?).

We have addressed
- `let (x, y) =`
- `let {x, y} =`
- `let () =`
- `let _ =`
- `let f (x, y) =` (sort-of)

However, we've mixed examples with `let` and parameter definitions. This whole section needs to be improved to reflect the various cases. We don't need a section for each case, but having a section dedicated to pattern matching in parameter definition, as you proposed, is an excellent idea.

christinerose (approved):

sabine (approved):
This seems nice. Is there anything else that needs to be done here before merge? From my side this looks good.

data/tutorials/language/0it_00_values_functions.md
```diff
@@ -156,6 +156,23 @@ val first : string = "Robin"
...
+# let age ((years, full_name) : live_person) = years;;
+val age : live_person -> int = <fun>
+```
+
+For records:
+```ocaml
+# let introduce ({first; last}) : name) = "I am " ^ first ^ " " ^ last;;
+val introduce : name -> string = <fun>
+```
```
cuihtlauac:
````suggestion
### Pattern Matching in Function Parameters

Single-case pattern matching can also be used for parameter declaration.

Here is an example with tuples:
```ocaml
# let get_country ((country, { first; last }) : citizen) = country;;
val get_countruy : citizen -> string = <fun>
```

Here is an example with the `name` record:
```ocaml
# let introduce {first; last} = "I am " ^ first ^ " " ^ last;;
val introduce : name -> string = <fun>
```

**Note** Using the `discard` pattern for parameter declaration is also possible.
```ocaml
# let get_meaning _ = 42;;
val get_meaning : 'a -> int = <fun>
```

````

This section should go at the end of the section _Pattern Matching in Definition_ (after _Discarding Values Using Pattern Matching_). That section should be renamed _Pattern Matching in Definitions and Function Parameters_

The text (line 124):
> Some definitions can introduce more than one name or no name.

should be replaced by:
> When pattern matching only has one case it can be used in `let` definitions. In that case, less or more than one name may be defined. This applies to tuples, records and custom single-variant types.

The example in the section _Pattern Matching on Records_ should be changed into
```ocaml
# type name = { first : string; last: string };;
type name = { first : string; last : string; }

# let robin = { first = "Robin"; last = "Milner" };;
val robin : name = {first = "Robin"; last = "Milner"}

# let { first = given_name; last = surname } = robin;;
val given_name : string = "Robin"
val surname : string = "Milner"
```

The example in section _Pattern Matching on User-Defined Types_ should be changed into:
```ocaml
# type citizen = string * name;;
type citizen = string * name

# let (country, { last; first }) = ("United Kingdom", robin);;
val country : string = "United Kingdom"
val first : string = "Robin"
val last : string = "Milner"
```

data/tutorials/language/0it_00_values_functions.md
```diff
@@ -177,11 +194,13 @@ Above, the pattern does not contain any identifier, meaning no name is defined.
 
 This also works with user-defined types.
 ```ocaml
-# type live_person = int * name;;
-type live_person = int * name
+# let robin_with_age : live_person = (20, robin);;
+val robin_with_age : live_person = (20, {first = "Robin"; last = "Milner"})
```
cuihtlauac, line 198: See suggestion above.

data/tutorials/language/0it_00_values_functions.md
```diff
@@ -177,11 +194,13 @@ Above, the pattern does not contain any identifier, meaning no name is defined.
...
 
-# let age ((years, { first; last }) : live_person) = years;;
-val age : live_person -> int = <fun>
+# let (years, { first; last }) = robin_with_age;;
+val years : int = 20
+val first : string = "Robin"
+val last : string = "Milner"
```
cuihtlauac, line 203: Same as previous

cuihtlauac (changes requested):
Sorry @srj31 for the late review. It took me way too long to put things together. Thanks for submitting this. It greatly helped understanding what was wrong in the current text

data/tutorials/language/0it_00_values_functions.md
cuihtlauac:
```suggestion
# type citizen = string * name;;
type citizen = string * name
```

data/tutorials/language/0it_00_values_functions.md
cuihtlauac:
```suggestion
# let ((country, { first = forename; last = surname }) : citizen) = ("United Kingdom", robin);;
val country : string = "United Kingdom"
val forename : string = "Robin"
val surname : string = "Milner"
```

cuihtlauac (changes requested):

cuihtlauac (approved):

PR #1984, opened 2024-01-22
changelog: infer date from slug if not in metadata

emillon:
The date is present in 2 places: in the slug and in the yaml metadata.

This is redundant and creates extra work for contributors.

This makes the yaml metadata optional; in that case the date is inferred from the slug.


emillon:
For dune I checked that the metadata was consistently duplicated (there was one discrepancy but the slug version was correct).
I didn't want to change everything else, but if you prefer we can cleanup the rest of the sources and always source from the slug.

christinerose (approved):

cuihtlauac:
Here are the discrepancies I'm finding on the whole set:
```
$ git grep '^date:' -- data/changelog | awk 'match($0, /([0-9]{4}-[0-9]{2}-[0-9]{2}).*([0-9]{4}-[0-9]{2}-[0-9]{2})/, arr) { if (arr[1] != arr[2]) print }'
data/changelog/dune/2022-02-11-dune-3.0.0.md:date: "2022-02-19"
data/changelog/infra/2023-12-04-relocate-services.md:date: "2023-12-05"
data/changelog/ocaml/2023-04-14-ocaml-5.1.alpha1.md:date: "2023-04-15"
```

I'm tempted to keep the slug date everywhere.

emillon:
OK let's do this

emillon:
Done. I had to sanitize a bit the input when parsing because some slugs look like `2023-1-2` with no zeroes, which is not parsed correctly later.

cuihtlauac (approved):
This looks good to me. In 1984, we rewrite history to remove redundancy. I like that.

emillon:
it's ok to change the slugs as well, but this will break some links so that's more risky. up to you!

cuihtlauac:
Consistency primes accuracy, let's leave the slugs unchanged.

PR #1987, opened 2024-01-22
ood-gen: call crunch on each directory separately

emillon:
This has several benefits:
- the various directories can be processed in parallel
- the generated modules are smaller, so the superlinear behavior of the
  compiler is less visible
- dependencies are more precise, so fewer recompilations should be
  necessary

### DIFF
--- a/tool/ood-gen/lib/data.ml
+++ b/tool/ood-gen/lib/data.ml
@@ -1,0 +1,46 @@
+module type DB = sig
+  val read : string -> string option
+  val file_list : string list
+end
+
+let dbs : (_ * (module DB)) list =
+  [
+    ("academic_institutions", (module Data_academic_institutions));
+    ("books", (module Data_books));
+    ("changelog", (module Data_changelog));
+    ("code_examples", (module Data_code_examples));
+    ("events", (module Data_events));
+    ("exercises", (module Data_exercises));
+    ("industrial_users", (module Data_industrial_users));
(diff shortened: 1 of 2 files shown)


emillon:
This PR is more to start a discussion. While contributing I noticed that a single command was taking a long time. The ocaml compiler has trouble with large generated modules. So instead of doing that, this PR generates smaller modules. They build more efficiently, can be built in parallel and have better incremental properties.

Here's a quick benchmark:
- this PR: dune build  51.69s user 7.19s system 318% cpu 18.475 total
- main: dune build  71.25s user 7.92s system 200% cpu 39.436 total
(so we can see that there's both less computation to do, and it is also more parallel)

There are several places we can go with that:
- we can make a dune rule generator to remove some boilerplate and make the code easier to adapt when the directory hierarchy will change
- we could experiment with different scheduling strategies to make the data buckets more even (for example, add more layers of dispatch, or dispatch based on the filename hash)
- maybe this should be upstreamed to crunch or a new tool

Let me know what you think.

cuihtlauac:
This is great, I think it is the way to go

sabine (approved):
Thank you @emillon!

This makes compilation noticeably faster and since @cuihtlauac also approves of it, let's merge it!

before:
```
real	0m34.138s
user	0m56.223s
sys	0m10.930s
```
after:
```
real	0m16.128s
user	0m43.780s
sys	0m11.639s
```

PR #1988, opened 2024-01-23
Remove references to the platform installer

cuihtlauac:
- Text in /docs/platform
- Url and routes

### DIFF
--- a/src/global/url.ml
+++ b/src/global/url.ml
@@ -80,12 +80,8 @@ let tutorial name = "/docs/" ^ name
 let getting_started = "/docs/get-started"
 let installing_ocaml = "/docs/installing-ocaml"
 let exercises = "/exercises"
-let installer = "/install-platform.sh"
 let outreachy = "/outreachy"
 let logos = "/logo"
-
-let github_installer =
-  "https://github.com/tarides/ocaml-platform-installer/releases/latest/download/installer.sh"
 
 let github_opam_file package_name package_version =
   Printf.sprintf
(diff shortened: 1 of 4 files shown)


panglesd:
Just to satisfy my curiosity, what are the main reasons to remove those references?

cuihtlauac:
Sorry, this PR should been a draft. What are the plans regarding the installer?

tmattio:
With the ongoing work to add package management to Dune, which will make Dune the recommended tool for people to build projects, the way to get started and install developer tools will change. In the meantime, the recommended way is to install them manually with opam, and the easiest (but unofficial) way still seems to be to use the Platform installer, so the documentation seems correct. 

I initially thought that in light of the above, we shoul keep the documentation as-is, but:
- Since ocaml-platform-installer is being less maintained (as a result of very few opened issues, but I don't think we plan to spend a lot of time on the tool in the future given the focus on Dune 4)
- We've had a very high bar for what gets into the official docs so as to keep OCaml.org unbiased, and while ocaml-platform-installer is a great tool, I'm not sure it meets the standards we've applied to other tools
- There seem to be very few users of the tool, despite Tarides advertising for it

I would suggest that we go through with that PR. Does that sound good to you @panglesd? I know we discussed the opposite a few days ago.

panglesd:
That sounds good to me! I think ocaml.org being unbiased is an important goal, and the installer has not being through any process to enter the platform tools.

It had its chance, being in the official doc for some time, but did not gain much attention, so I also think going on with this PR won't change the user experience a lot when installing OCaml.

sabine (approved):
Alright, then let's merge.

PR #1999, opened 2024-01-25
Consider Dune blocks as Lisp source

cuihtlauac:
Use `lisp` instead of `dune` as markdown code block header. This
allows blocks to be highlighted.


sabine:
Only works in HackMD, right? `hilite` currently supports neither `dune` nor `lisp` syntax highlighting.

PR #2002, opened 2024-01-25
fix typo in the tutorial on optional params

julbinb:
the example contains `?init:(x=0)` but the sentence below it uses 1 instead of 0


cuihtlauac (approved):
Thanks, @julbinb; this is indeed a bug!

PR #2023, opened 2024-02-03
Improve some wording in the functors tutorial

neuroevolutus:
This PR adds some minor grammatical and wording changes to the functors tutorial that will hopefully make things clearer for new users of OCaml.

In particular, I made some changes to the aside about the `with type` constraint that may need particular review to ensure I did not state something that is semantically incorrect.

### DIFF
--- a/data/tutorials/language/0it_01_basic_datatypes.md
+++ b/data/tutorials/language/0it_01_basic_datatypes.md
@@ -482,16 +482,6 @@ Note that:
 - `unit` is a variant with a unique constructor, which does not carry data: `()`.
 - `bool` is also a variant with two constructors that doesn't carry data: `true` and `false`.
 
-#### Empty Variants
-
-Variants can optionally be defined with absolutely no constructors at all.
-
-```ocaml
-type void = |
-```
-
-Such types are not ordinarily useful in OCaml programs (as they do not have any constructible values), but they can be useful as temporary placeholders when defining types in a [functor](/docs/functors#writing-your-own-functors).
-
 #### Constructors With Data
 
 It is possible to wrap data in constructors. The following type has several constructors with data (e.g., `Hash of string`) and some without (e.g., `Head`). It represents the different means to refer to a Git [revision](https://git-scm.com/docs/gitrevisions).
--- a/data/tutorials/language/0it_01_basic_datatypes.md
+++ b/data/tutorials/language/0it_01_basic_datatypes.md
@@ -482,6 +482,16 @@ Note that:
 - `unit` is a variant with a unique constructor, which does not carry data: `()`.
 - `bool` is also a variant with two constructors that doesn't carry data: `true` and `false`.
 
+#### Empty Variants
+
+Variants can optionally be defined with absolutely no constructors at all.
+
+```ocaml
+type void = |
+```
+
+Such types are not ordinarily useful in OCaml programs (as they do not have any constructible values), but they can be useful as temporary placeholders when defining types in a [functor](/docs/functors#writing-your-own-functors).
+
 #### Constructors With Data
 
 It is possible to wrap data in constructors. The following type has several constructors with data (e.g., `Hash of string`) and some without (e.g., `Head`). It represents the different means to refer to a Git [revision](https://git-scm.com/docs/gitrevisions).
--- a/data/tutorials/language/1ms_01_functors.md
+++ b/data/tutorials/language/1ms_01_functors.md
@@ -242,8 +242,8 @@ module type S = sig
 end
 
 module Binary(Elt: OrderedType) : S = struct
-  type elt = | (* Replace by your own *)
-  type t = | (* Replace by your own *)
+  type elt (* = replace by your own *)
+  type t (* = replace by your own *)
   (* Add private functions here *)
   let empty = failwith "Not yet implemented"
   let is_empty h = failwith "Not yet implemented"
--- a/data/tutorials/language/1ms_01_functors.md
+++ b/data/tutorials/language/1ms_01_functors.md
@@ -374,6 +374,70 @@ Check the program's behaviour using `opam exec -- dune exec funkt < dune`.
 
 **Note**: Modules received and returned by `IterPrint.Make` both have a type `t`. The `with type ... :=` constraint is needed to make the two `t` identical. This allows functions from the injected dependency and result module to use the same type. When the parameter's contained type is not exposed by the result module (i.e. when it is an _implementation detail_), the `with type` constraint is unnecessary. 
 
+### Naming and Scoping
+
+In the previous section, we learned how to use the `with type` constraint in order to unify a type `t` contained within both the parameter and result module of a functor. Let's go over a few more details concerning naming and scoping to get a better grasp of the mechanics of this constraint.
+
+When reading the source of `iterPrint.ml`, it may have seemed curious as to why we could not have simply defined `Make` as follows:
+
+```ocaml
+module Make(Dep: Iterable) : S = struct
+  type 'a t = 'a Dep.t
+  let f = Dep.iter (fun s -> Out_channel.output_string stdout (s ^ "\n"))
+end
+```
+
+In the absence of client code that utilises the function `f` provided by the output of `Make`, the project would compile without error.
+
+However, since `Make` is invoked to create module `IterPrint` in `funkt.ml`, the project will fail to compile:
+
+```shell
+5 | ..stdin
+6 |   |> In_channel.input_lines
+7 |   |> List.concat_map Str.(split (regexp "[ \t.,;:()]+"))
+8 |   |> StringSet.of_list
+9 |   |> StringSet.elements
+Error: This expression has type string list
+       but an expression was expected of type string IterPrint.t
+```
+
+This may seem odd given the fact that the modified `Make` functor successfully compiles on its own. Indeed, if we had used an inappropriate definition such as `type 'a t = MkT`, the functor itself would have failed to compile:
+
+```shell
+11 | .................................struct
+12 |   type 'a t = MkT
+13 |   let f = Dep.iter (fun s -> Out_channel.output_string stdout (s ^ "\n"))
+14 | end
+Error: Signature mismatch:
+       Modules do not match:
+         sig type 'a t = MkT val f : string Dep.t -> unit end
+       is not included in
+         S
+       Values do not match:
+         val f : string Dep.t -> unit
+       is not included in
+         val f : string t -> unit
+       The type string Dep.t -> unit is not compatible with the type
+         string t -> unit
+       Type string Dep.t is not compatible with type string t
+```
+
+The key thing to realise is that, outside the functor, client code is not privy to the fact that `type 'a t` is being set equal to `Dep.t`. In `funkt.ml`, `IterPrint.t` simply appears as an abstract type exposed by the result of `Make`. This is precisely why the `with type` constraint is needed to propagate the knowledge that `IterPrint.t` is the same as the instantiation of `Dep.t` (`List.t` in this case).
+
+Another property of the `with type` constraint is that the type it exposes won't be shadowed by definitions within the functor body. In fact, the `Make` functor could be redefined as follows without preventing the successful compilation of module `Funkt`:
+
+```ocaml
+module Make(Dep: Iterable) : S with type 'a t := 'a Dep.t = struct
+  type 'a t = MkT
+  let g _ = "MkT"
+  let f = Dep.iter(fun s ->
+    Out_channel.output_string stdout (g MkT ^ "\n");
+    Out_channel.output_string stdout (s ^ "\n"))
+end
+```
+
+In the example above, the local type `t` does not shadow the type `t` exposed by the `with type` constraint and can be used for the implementation of the functor. However, it is generally better to avoid availing of this behaviour since it may make your code more difficult to understand.
+
 ## Write a Functor to Extend Modules
 
 In this section, we define a functor to extend several modules in the same way. This is the same idea as in the [Extending a Module with a Standard Library Functor](#extending-a-module-with-a-standard-library-functor), except we write the functor ourselves.
--- a/data/tutorials/language/1ms_01_functors.md
+++ b/data/tutorials/language/1ms_01_functors.md
@@ -415,7 +415,7 @@ module Make(Dep: Iterable) : S with type 'a t := 'a Dep.t = struct
 end
 ```
 
-In the example above, the local type `t` does not shadow the type `t` exposed by the `with type` constraint and can be used for the implementation of the functor. However, it is generally better to avoid availing of this behaviour since it may make your code more difficult to understand.
+In the example above, `t` from `with type` takes precedence over the local `t`, which only has a local scope. However, it is generally better to avoid availing of this behaviour since it may make your code more difficult to understand.
 
 ## Write a Functor to Extend Modules
 
--- a/data/tutorials/language/1ms_01_functors.md
+++ b/data/tutorials/language/1ms_01_functors.md
@@ -428,10 +428,10 @@ Another property of the `with type` constraint is that the type it exposes won't
 
 ```ocaml
 module Make(Dep: Iterable) : S with type 'a t := 'a Dep.t = struct
-  type 'a t = MkT
-  let g _ = "MkT"
+  type 'a t = LocalType
+  let g LocalType = "LocalType"
   let f = Dep.iter(fun s ->
-    Out_channel.output_string stdout (g MkT ^ "\n");
+    Out_channel.output_string stdout (g LocalType ^ "\n");
     Out_channel.output_string stdout (s ^ "\n"))
 end
 ```
--- a/data/tutorials/language/1ms_01_functors.md
+++ b/data/tutorials/language/1ms_01_functors.md
@@ -63,7 +63,7 @@ Here is how this reads (starting from the bottom, then going up):
 
 **Note**: Most set operations need to compare elements to check if they are the same. To allow using a user-defined comparison algorithm, the `Set.Make` functor takes a module the specifies both the element type `t` and the `compare` function. Passing the comparison function as a higher-order parameter, as done in `Array.sort`, for example, would add a lot of boilerplate code. Providing set operations as a functor allows specifying the comparison function only once.
 
-Here is an example how to use `Set.Make`:
+Here is an example of how to use `Set.Make`:
 
 **`funkt.ml`**
 
--- a/data/tutorials/language/1ms_01_functors.md
+++ b/data/tutorials/language/1ms_01_functors.md
@@ -80,10 +80,6 @@ This defines a module `Funkt.StringSet`. What `Set.Make` needs are:
 - Type `t`, here `string`
 - Function allowing to compare two values of type `t`, here `String.compare`
 
-However, since the module `String` defines
-- Type name `t`, which is an alias for `string`
-- Function `compare` of type `t -> t -> bool` compares two strings
-
 This can be simplified using an _anonymous module_ expression:
 ```ocaml
 module StringSet = Set.Make(struct
@@ -93,6 +89,10 @@ end)
 ```
 
 The module expression `struct ... end` is inlined in the `Set.Make` call.
+
+However, since the module `String` already defines
+- Type name `t`, which is an alias for `string`
+- Function `compare` of type `t -> t -> bool` compares two strings
 
 This can be simplified even further into this:
 ```ocaml
--- a/data/tutorials/language/1ms_01_functors.md
+++ b/data/tutorials/language/1ms_01_functors.md
@@ -242,8 +242,8 @@ module type S = sig
 end
 
 module Binary(Elt: OrderedType) : S = struct
-  type elt (* = replace by your own *)
-  type t (* = replace by your own *)
+  type elt (* Add your own type definition *)
+  type t (* Add your own type definition *)
   (* Add private functions here *)
   let empty = failwith "Not yet implemented"
   let is_empty h = failwith "Not yet implemented"
@@ -377,7 +377,6 @@ Check the program's behaviour using `opam exec -- dune exec funkt < dune`.
 ### Naming and Scoping
 
 The `with type` constraint unifies types within a functor's parameter and result modules. We've used that in the previous section. This section addresses the naming and scoping mechanics of this constraint.
-``
 
 Naively, we might have defined `Iter.Make` as follows:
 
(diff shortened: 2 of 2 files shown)


data/tutorials/language/1ms_01_functors.md
```diff
@@ -371,7 +371,7 @@ let _ =
 
 Check the program's behaviour using `opam exec -- dune exec funkt < dune`.
 
-**Note**: The functor `IterPrint.Make` returns a module that exposes the type from the injected dependency (here first `List.t` then `Array.t`). That's why a `with type` constraint is needed. When parametrising other something not exposed by the module (and _implementation detail_), the `with type` constraint is not needed.
+**Note**: The functor `IterPrint.Make` exposes the type from the injected dependency (here first `List.t` then `Array.t`) as the type `t` required of the result module. That's why a `with type` constraint is needed. When customising a type that's not exposed by the result module (i.e. an _implementation detail_), the `with type` constraint is not needed. As a further consequence of this, we could even define a local type `t` within the `struct` block for use within the implementation, and such a type `t` would not shadow the type `t` of the result module.
```
cuihtlauac: Thanks, @neuroevolutus, that note needs to be clarified. Your text is better than is initial one, do you think this improves on it:

```suggestion
**Note**: Modules received and returned by `IterPrint.Make` both have a type `t`. The `with type ... :=` constraint is needed to make the two `t` identical. This allows functions from the injected dependency and result module to use the same type. When the parameter's contained type is not exposed by the result module (i.e. when it is an _implementation detail_), the `with type` constraint is unnecessary. 
```

Regarding your last sentence:

> As a further consequence, we could even define a local type `t` within the `struct` block for use within the implementation, and such a type `t` would not shadow the type `t` of the result module.

This is a bit hard to grasp because it requires 1/ figuring out what the code would look like if modified with a local `t` and 2/ figuring out the shadowing concern.

However, I believe you have a valid point. This is a matter that needs to be addressed. Could we address it somewhere else in the document or do we need an additional example? We may even need a complete section to address naming and shadowing concerns in relationship with functor definitions.

What is our opinion on this?

cuihtlauac (changes requested):

data/tutorials/language/1ms_01_functors.md
neuroevolutus: Thanks so much for the rewording. I think it's a lot more succinct and even clearer now. I made sure to commit your suggestion.

> This is a bit hard to grasp because it requires 1/ figuring out what the code would look like if modified with a local t and 2/ figuring out the shadowing concern.

I agree; that is a good point :sweat_smile:. I think adding a new section with a concrete example to address these concerns would be really nice. I'll work on adding this section and make a new commit to get your feedback.

neuroevolutus: As a small aside, I added a small section on empty variants to the data types tutorial to give context for the use of a singular `|` as a placeholder type definition.

data/tutorials/language/1ms_01_functors.md
```diff
@@ -157,14 +157,14 @@ let _ =
 
 This allows the user to seemingly extend the module `String` with a submodule `Set`. Check the behaviour using `opam exec -- dune exec funkt < dune`.
 
-## Functors Allows Parametrising Modules
+## Functors Facilitate Parametrising Modules
```
cuihtlauac: What do you think of this?
```suggestion
## Parametrising Modules with Functors 
```

data/tutorials/language/1ms_01_functors.md
```diff
@@ -371,7 +372,71 @@ let _ =
 
 Check the program's behaviour using `opam exec -- dune exec funkt < dune`.
 
-**Note**: The functor `IterPrint.Make` returns a module that exposes the type from the injected dependency (here first `List.t` then `Array.t`). That's why a `with type` constraint is needed. When parametrising other something not exposed by the module (and _implementation detail_), the `with type` constraint is not needed.
+**Note**: Modules received and returned by `IterPrint.Make` both have a type `t`. The `with type ... :=` constraint is needed to make the two `t` identical. This allows functions from the injected dependency and result module to use the same type. When the parameter's contained type is not exposed by the result module (i.e. when it is an _implementation detail_), the `with type` constraint is unnecessary. 
+
+### Naming and Scoping
+
+In the previous section, we learned how to use the `with type` constraint in order to unify a type `t` contained within both the parameter and result module of a functor. Let's go over a few more details concerning naming and scoping to get a better grasp of the mechanics of this constraint.
```
```suggestion
In the previous section, we learned how to use the `with type` constraint in order to unify types contained within both the parameter and result module of a functor. Let's go over a few more details concerning naming and scoping to get a better grasp of the mechanics of this constraint.
```

data/tutorials/language/1ms_01_functors.md
```diff
@@ -371,7 +372,71 @@ let _ =
...
+When reading the source of `iterPrint.ml`, it may have seemed curious as to why we could not have simply defined `Make` as follows:
+
+```ocaml
+module Make(Dep: Iterable) : S = struct
+  type 'a t = 'a Dep.t
+  let f = Dep.iter (fun s -> Out_channel.output_string stdout (s ^ "\n"))
+end
+```
+
+In the absence of client code that utilises the function `f` provided by the output of `Make`, the project would compile without error.
```
```suggestion
In the absence of code that utilises the function `f` provided by the output of `Make`, the project would compile without error.
```

data/tutorials/language/1ms_01_functors.md
```diff
@@ -371,7 +372,71 @@ let _ =
...
+
+However, since `Make` is invoked to create module `IterPrint` in `funkt.ml`, the project will fail to compile:
```
```suggestion
However, since `Make` is invoked to create module `IterPrint` in `funkt.ml`, the project will fail to compile with the following error message:
```

data/tutorials/language/1ms_01_functors.md
```diff
@@ -371,7 +372,71 @@ let _ =
...
+       is not included in
+         S
+       Values do not match:
+         val f : string Dep.t -> unit
+       is not included in
+         val f : string t -> unit
+       The type string Dep.t -> unit is not compatible with the type
+         string t -> unit
+       Type string Dep.t is not compatible with type string t
+```
```
cuihtlauac: Do you think this example is needed? Let's focus on a single story, the unification constraint.

data/tutorials/language/1ms_01_functors.md
```diff
@@ -371,7 +372,71 @@ let _ =
...
+The key thing to realise is that, outside the functor, client code is not privy to the fact that `type 'a t` is being set equal to `Dep.t`. In `funkt.ml`, `IterPrint.t` simply appears as an abstract type exposed by the result of `Make`. This is precisely why the `with type` constraint is needed to propagate the knowledge that `IterPrint.t` is the same as the instantiation of `Dep.t` (`List.t` in this case).
+
+Another property of the `with type` constraint is that the type it exposes won't be shadowed by definitions within the functor body. In fact, the `Make` functor could be redefined as follows without preventing the successful compilation of module `Funkt`:
+
+```ocaml
+module Make(Dep: Iterable) : S with type 'a t := 'a Dep.t = struct
+  type 'a t = MkT
+  let g _ = "MkT"
+  let f = Dep.iter(fun s ->
+    Out_channel.output_string stdout (g MkT ^ "\n");
```
```suggestion
  type 'a t = LocalType
  let g LocalType = "LocalType"
  let f = Dep.iter(fun s ->
    Out_channel.output_string stdout (g LocalType ^ "\n");
```

data/tutorials/language/1ms_01_functors.md
```diff
@@ -371,7 +372,71 @@ let _ =
...
+    Out_channel.output_string stdout (s ^ "\n"))
+end
+```
+
+In the example above, the local type `t` does not shadow the type `t` exposed by the `with type` constraint and can be used for the implementation of the functor. However, it is generally better to avoid availing of this behaviour since it may make your code more difficult to understand.
```
cuihtlauac: I understand your point. I must admit I had never fallen into that pitfall and wasn't aware of it. 

What do you think about turning the narrative into something like:

> `t` from `with type` takes over local `t` which only has a local scope?

Again, the idea is to make things more direct.

cuihtlauac (changes requested):
Thanks, @neuroevolutus, this is great! I wonder if the difference between `=` and `:=` type constraint is needed. But overall, you are absolutely right, this matter is missing in the current version.

data/tutorials/language/1ms_01_functors.md
```diff
@@ -371,7 +371,7 @@ let _ =
 
 Check the program's behaviour using `opam exec -- dune exec funkt < dune`.
 
-**Note**: The functor `IterPrint.Make` exposes the type from the injected dependency (here first `List.t` then `Array.t`) as the type `t` required of the result module. That's why a `with type` constraint is needed. When customising a type that's not exposed by the result module (i.e. an _implementation detail_), the `with type` constraint is not needed. As a further consequence of this, we could even define a local type `t` within the `struct` block for use within the implementation, and such a type `t` would not shadow the type `t` of the result module.
+**Note**: Modules received and returned by `IterPrint.Make` both have a type `t`. The `with type ... :=` constraint is needed to make the two `t` identical. This allows functions from the injected dependency and result module to use the same type. When the parameter's contained type is not exposed by the result module (i.e. when it is an _implementation detail_), the `with type` constraint is unnecessary.
```
christinerose:
```suggestion
**Note**: Modules received and returned by `IterPrint.Make` both have a type `t`. The `with type ... :=` constraint is needed to make the two `t` modules identical. This allows functions from the injected dependency and result module to use the same type. When the parameter's contained type is not exposed by the result module (i.e., when it is an _implementation detail_), the `with type` constraint is unnecessary. 
```
It reads funny to me to have a plural (two) with just `t` -- is it accurate to have `t` modules instead? ...or `t` types?

Regardless, there needs to be a comma after i.e.

data/tutorials/language/1ms_01_functors.md
neuroevolutus: Yeah, that's way better :smile:. Short and sweet.

data/tutorials/language/1ms_01_functors.md
neuroevolutus: For sure, frankly I agree with you :+1:. This was mostly me digressing as a result of the fiddling I was doing trying to understand the code locally on my computer :sweat_smile:. I'll add a new commit to undo this.

data/tutorials/language/1ms_01_functors.md
neuroevolutus: I agree; that sounds good. It sounds clearer to me as a reader.

neuroevolutus: I used "takes precedence over" in the rewording, but let me know if you feel it's clearer with just saying "takes over."

neuroevolutus:
> I wonder if the difference between `=` and `:=` type constraint is needed. But overall, you are absolutely right, this matter is missing in the current version.

Actually, this is something I was struggling with a bit. I realised when playing around with the code that `with type ... = ...` parses just fine but creates a compilation error. I read the section on modules in the OCaml manual, but it seemed to describe the `with type ... = ...` constraint and not the `with type ... := ...` one. Is there some intuition I'm missing on the difference between `=` and `:=`?

data/tutorials/language/0it_01_basic_datatypes.md
```diff
@@ -482,6 +482,16 @@ Note that:
...
+#### Empty Variants
+
+Variants can optionally be defined with absolutely no constructors at all.
+
+```ocaml
+type void = |
+```
+
+Such types are not ordinarily useful in OCaml programs (as they do not have any constructible values), but they can be useful as temporary placeholders when defining types in a [functor](/docs/functors#writing-your-own-functors).
+
```
Octachron: This is not the intended use case for empty types (see https://ocaml.org/manual/emptyvariants.html for an explanation of why they were introduced). An abstract type works better as a placeholder for the use case you are describing. I would thus avoid mentioning empty variant types before the GADT introduction.

data/tutorials/language/1ms_01_functors.md
neuroevolutus: Thanks for catching that. I made sure to add the comma in that part.

With regards to the `t`, I think it may be fine to keep it as is since the `t` refers to two types rather than modules per se. What are your thoughts, @cuihtlauac?

data/tutorials/language/0it_01_basic_datatypes.md
neuroevolutus: I see. That would make sense. I'll revert the corresponding commit.

neuroevolutus: I've edited the `Binary` functor example to make `elt` and `t` abstract types.

Thanks for the link to the manual! I'll check that out.

Octachron: To give a better example of what I meant by "abstract type are better placeholder", if you write
```ocaml
type empty = |
let f ([]: empty list) = ()
```
the function `f` is total: the only list of empty values is the empty list. But this is property is only true for empty types. In other words, by using empty types as placeholders one might accidentally capture properties that are only true for empty types.

neuroevolutus: I see. I think I better understand now. As an aside, are there any examples in the wild where using empty types and refutation cases for them come in handy when it comes to GADTs? I think I'm still fuzzy on the motivation for empty types.

data/tutorials/language/1ms_01_functors.md
```diff
@@ -371,7 +372,50 @@ let _ =
 
 Check the program's behaviour using `opam exec -- dune exec funkt < dune`.
 
-**Note**: The functor `IterPrint.Make` returns a module that exposes the type from the injected dependency (here first `List.t` then `Array.t`). That's why a `with type` constraint is needed. When parametrising other something not exposed by the module (and _implementation detail_), the `with type` constraint is not needed.
+**Note**: Modules received and returned by `IterPrint.Make` both have a type `t`. The `with type ... :=` constraint is needed to make the two `t` identical. This allows functions from the injected dependency and result module to use the same type. When the parameter's contained type is not exposed by the result module (i.e., when it is an _implementation detail_), the `with type` constraint is unnecessary.
+
+### Naming and Scoping
+
+In the previous section, we learned how to use the `with type` constraint in order to unify types contained within both the parameter and result module of a functor. Let's go over a few more details concerning naming and scoping to get a better grasp of the mechanics of this constraint.
```
```suggestion
The `with type` constraint unifies types within a functor's parameter and result modules. We've used that in the previous section. This section addresses the naming and scoping mechanics of this constraint.
``

File: data/tutorials/language/1ms_01_functors.md [hunk 12]
```diff
@@ -371,7 +372,50 @@ let _ =
...
+
+When reading the source of `iterPrint.ml`, it may have seemed curious as to why we could not have simply defined `Make` as follows:
```
```suggestion
Naively, we might have defined `Iter.Make` as follows:
```

"Simple", "obvious", "evident", and others are forbidden words :-) 

Also, shorter is (almost) always better.

data/tutorials/language/1ms_01_functors.md
```diff
@@ -371,7 +372,50 @@ let _ =
...
+
+```ocaml
+module Make(Dep: Iterable) : S = struct
+  type 'a t = 'a Dep.t
+  let f = Dep.iter (fun s -> Out_channel.output_string stdout (s ^ "\n"))
+end
+```
+
+In the absence of code that utilises the function `f` provided by the output of `Make`, the project would compile without error.
```
```suggestion
In the absence of code that utilises the function `f` provided by the output of `Make`, the project compiles without error.
```

data/tutorials/language/1ms_01_functors.md
```diff
@@ -371,7 +372,50 @@ let _ =
...
+
+However, since `Make` is invoked to create module `IterPrint` in `funkt.ml`, the project will fail to compile with the following error message:
```
```suggestion
However, since `Make` is invoked to create module `IterPrint` in `funkt.ml`, the project fails to compile with the following error message:
```

data/tutorials/language/1ms_01_functors.md
```diff
@@ -371,7 +372,50 @@ let _ =
...
+5 | ..stdin
+6 |   |> In_channel.input_lines
+7 |   |> List.concat_map Str.(split (regexp "[ \t.,;:()]+"))
+8 |   |> StringSet.of_list
+9 |   |> StringSet.elements
+Error: This expression has type string list
+       but an expression was expected of type string IterPrint.t
+```
+
+The key thing to realise is that, outside the functor, client code is not privy to the fact that `type 'a t` is being set equal to `Dep.t`. In `funkt.ml`, `IterPrint.t` simply appears as an abstract type exposed by the result of `Make`. This is precisely why the `with type` constraint is needed to propagate the knowledge that `IterPrint.t` is the same as the instantiation of `Dep.t` (`List.t` in this case).
```
```suggestion
Outside the functor, it is not known that `type 'a t` is set to `Dep.t`. In `funkt.ml`, `IterPrint.t` appears as an abstract type exposed by the result of `Make`. This is why the `with type` constraint is needed. It propagates the knowledge that `IterPrint.t` is the same type as `Dep.t` (`List.t` in this case).
```

data/tutorials/language/1ms_01_functors.md
```diff
@@ -371,7 +372,50 @@ let _ =
...
+
+Another property of the `with type` constraint is that the type it exposes won't be shadowed by definitions within the functor body. In fact, the `Make` functor could be redefined as follows without preventing the successful compilation of module `Funkt`:
```
```suggestion
The type constrained using `with type` isn't shadowed by definitions within the functor body. In the example, the `Make` functor can be redefined as follows:
```

data/tutorials/language/1ms_01_functors.md
```diff
@@ -371,7 +372,50 @@ let _ =
...
+module Make(Dep: Iterable) : S with type 'a t := 'a Dep.t = struct
+  type 'a t = LocalType
+  let g LocalType = "LocalType"
+  let f = Dep.iter(fun s ->
+    Out_channel.output_string stdout (g LocalType ^ "\n");
+    Out_channel.output_string stdout (s ^ "\n"))
+end
+```
+
+In the example above, `t` from `with type` takes precedence over the local `t`, which only has a local scope. However, it is generally better to avoid availing of this behaviour since it may make your code more difficult to understand.
```
```suggestion
In the example above, `t` from `with type` takes precedence over the local `t`, which only has a local scope. Don't shadow names too often because it makes the code harder to understand.
```

cuihtlauac (changes requested):

data/tutorials/language/1ms_01_functors.md
```diff
@@ -371,7 +372,50 @@ let _ =
 
 Check the program's behaviour using `opam exec -- dune exec funkt < dune`.
 
-**Note**: The functor `IterPrint.Make` returns a module that exposes the type from the injected dependency (here first `List.t` then `Array.t`). That's why a `with type` constraint is needed. When parametrising other something not exposed by the module (and _implementation detail_), the `with type` constraint is not needed.
+**Note**: Modules received and returned by `IterPrint.Make` both have a type `t`. The `with type ... :=` constraint is needed to make the two `t` identical. This allows functions from the injected dependency and result module to use the same type. When the parameter's contained type is not exposed by the result module (i.e., when it is an _implementation detail_), the `with type` constraint is unnecessary.
```
```suggestion
**Note**: Modules received and returned by `IterPrint.Make` both have a type `t`. The `with type ... := ...` constraint exposes that the two types `t` are the same. This makes functions from the injected dependency and result module use the exact same type. When the parameter's contained type is not exposed by the result module (i.e., when it is an _implementation detail_), the `with type` constraint is not necessary.
```

data/tutorials/language/1ms_01_functors.md
```diff
@@ -371,7 +372,51 @@ let _ =
...
+Naively, we might have defined `Iter.Make` as follows:
+
+```ocaml
+module Make(Dep: Iterable) : S = struct
+  type 'a t = 'a Dep.t
+  let f = Dep.iter (fun s -> Out_channel.output_string stdout (s ^ "\n"))
+end
+```
+
+In the absence of code that utilises the function `f` provided by the output of `Make`, the project compiles without error.
```
```suggestion
In the function `f` isn't used, the project compiles without error.
```

cuihtlauac (changes requested):

cuihtlauac (approved):

cuihtlauac:
> > I wonder if the difference between `=` and `:=` type constraint is needed. But overall, you are absolutely right, this matter is missing in the current version.
> 
> Actually, this is something I was struggling with a bit. I realised when playing around with the code that `with type ... = ...` parses just fine but creates a compilation error. I read the section on modules in the OCaml manual, but it seemed to describe the `with type ... = ...` constraint and not the `with type ... := ...` one. Is there some intuition I'm missing on the difference between `=` and `:=`?

I believe we can move forward with this PR without exposing the difference between the
`=` and `:=` constraints (see section [10.4](https://v2.ocaml.org/releases/5.1/htmlman/modtypes.html#ss%3Amty-with) of the manual). We can add more material on that later

data/tutorials/language/1ms_01_functors.md
```diff
@@ -242,9 +242,10 @@ module type S = sig
 end
 
 module Binary(Elt: OrderedType) : S = struct
-  type elt = | (* Replace by your own *)
-  type t = | (* Replace by your own *)
+  type elt (* = replace by your own *)
+  type t (* = replace by your own *)
```
```suggestion
  type elt (* Add your own type definition *)
  type t (* Add your own type definition *)
```

data/tutorials/language/1ms_01_functors.md
```diff
@@ -371,7 +372,51 @@ let _ =
 
 Check the program's behaviour using `opam exec -- dune exec funkt < dune`.
 
-**Note**: The functor `IterPrint.Make` returns a module that exposes the type from the injected dependency (here first `List.t` then `Array.t`). That's why a `with type` constraint is needed. When parametrising other something not exposed by the module (and _implementation detail_), the `with type` constraint is not needed.
+**Note**: Modules received and returned by `IterPrint.Make` both have a type `t`. The `with type ... := ...` constraint exposes that the two types `t` are the same. This makes functions from the injected dependency and result module use the exact same type. When the parameter's contained type is not exposed by the result module (i.e., when it is an _implementation detail_), the `with type` constraint is not necessary.
+
+### Naming and Scoping
+
+The `with type` constraint unifies types within a functor's parameter and result modules. We've used that in the previous section. This section addresses the naming and scoping mechanics of this constraint.
+``
```
```suggestion
```

cuihtlauac (changes requested):

data/tutorials/language/1ms_01_functors.md
```diff
@@ -157,14 +157,14 @@ let _ =
 
 This allows the user to seemingly extend the module `String` with a submodule `Set`. Check the behaviour using `opam exec -- dune exec funkt < dune`.
 
-## Functors Allows Parametrising Modules
+## Parametrising Modules with Functors
```
```suggestion
## Parametrising Modules with Functors
```

cuihtlauac (changes requested):

cuihtlauac (approved):
This looks good to me. Thnaks @neuroevolutus

neuroevolutus:
> This looks good to me. Thnaks @neuroevolutus

Thank you so much for all the feedback! I'm so happy to have had the chance to contribute.

PR #2036, opened 2024-02-08
Add changelog entry for ppxlib.0.32.0 release

NathanReb:
We just released ppxlib.0.32.0 to opam so here's the changelog entry for ocaml.org!


cuihtlauac:
Typo fixed in #2037

PR #2048, opened 2024-02-11
make the github links on ocaml.org point to the ocaml org

### LINKED ISSUE #2043: ocaml.org should link to the ocaml.org github repo
I've been meaning to report a few things about ocaml.org, but had no idea how to contact anyone responsible for the site : I looked at https://ocaml.org, https://ocaml.org/packages, https://v3.ocaml.org/packages, https://ocaml.org/community, I searched for "contact" or similar in the pages. I only found this repo by just guessing the url.

I think there should be a link to here in the footer, at a minimum. In the community page would be another option, but that's harder to find. What do you think?

v-gb:
Rather than just the ocaml repository (closes #2043).

### DIFF
--- a/src/ocamlorg_frontend/components/footer.eml
+++ b/src/ocamlorg_frontend/components/footer.eml
@@ -34,7 +34,7 @@ let policies = [
   ]
 
 let socials = [
-  ("https://github.com/ocaml/ocaml", "GitHub", Icons.github);
+  ("https://github.com/ocaml", "GitHub", Icons.github);
   ("https://discord.gg/cCYQbqN", "Discord", Icons.discord);
   ("https://twitter.com/ocamllang", "Twitter", Icons.twitter);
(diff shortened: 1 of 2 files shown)


cuihtlauac (approved):
That looks good to me, thanks @The-Amoghavarsha.

We should merge this. However, it looks as a minimum/interim solution. I believe we need more:

- Text, not mere links
- Links to
  - The OCaml GitHub organisation: https://github.com/ocaml
  - The OCaml compiler: https://github.com/ocaml/ocaml
  - The ocaml.org project: 
  - The Ocaml Foundation: https://ocaml-sf.org/

Some need to go into Community, some in Platform and some in each page's footer.

sabine (approved):
Good interim solution because the org page on GitHub is actually really nice because the main repos are pinned. Let's merge.

Issue #2051, opened 2024-02-12
Replace `bool option` by `bool` in data


cuihtlauac:
Data served by ocaml.org is static. Yaml files' headers from the [`/data`](../tree/main/data) folder are parsed and linked into the binary. The [`ood-gen`](../tree/main/tool/ood-gen) tool turns those files into caml sources compiled with the rest of the project.

The Yaml syntax is implicitly described by the `metadata` types in the files from the [`tool/ood-gen/lib`](../tree/main/tool/ood-gen/lib) folder. Parsing returns values of this type. However, at run time, another type is used. It is the type named `t` in the same file.

Some Yaml headers have optional fields containing optional toggles. Their parsers produce values of type `bool option`; this is not optimal. It adds unnecessary processing at runtime and code complexity.

This issue is about changing `bool option` typed fields in `t` types in the relevant files:

* [ ] [`planet.ml`](../blob/main/tool/ood-gen/lib/planet.ml)
* [x] [`release.ml`](../blob/main/tool/ood-gen/lib/release.ml)
* [X] [`workshop.ml`](../blob/main/tool/ood-gen/lib/workshop.ml)

The key idea is to replace `None` values by the default value (either `true` or `false`).

This is similar to PRs #1967, #2039 and #2042 (except they dealt about `list option`). Note that each type change affects other files and that the `metadata` types must not be changed.

egmaleta:
Hi, I can handle this too.

Regarding changes in `workshop.ml`, they will be similar to the ones made in 31384d1 so I think is valuable to either wait for #2042 to merge or to make these particular changes in that pr's branch. The type adjustement for `Workshop.presentation.poster` can be trivially made inside the `transform_presentation` function I introduced at 31384d1. What should I do?

sabine:
Thanks for you patience @egmaleta! We merged #2042 in the meantime. :+1:

Rebasing and resolving conflicts is usually not a big deal with PRs this small, so either of those ways of doing it would've been fine.

egmaleta:
Good to know for future contributions, thanks :)

egmaleta:
Hello, the 3rd task was completed in #2068.

In regards of the first one, what is the type to modify its `bool option` fields?
It seems to me that all types in task no.1 are metadata types and so they shouldn't be modified.

cuihtlauac:
> In regards of the first one, what is the type to modify its bool option fields? It seems to me that all types in task no.1 are metadata types and so they shouldn't be modified.

[...]

PR #2042, opened 2024-02-09
(ood-gen) Adjust type of fields with `list option` type to `list` in Changelog, News, Planet, and Workshop

egmaleta:
Removed `option` from type of fields:
- `Changelog.t.authors`
- `News.t.authors`
- `Planet.t.authors`
- `Workshop.presentations.additional_links`

This and #2039 might close #1970.

PR #2055, opened 2024-02-13
(ood-gen) Adjust type of fields with bool option type to bool in Release

egmaleta:
Solves task no. 2 of #2051, for now.

In `planet.ml` (task no. 1) there is nothing to do because all fields with type `bool option` belong to `metadata`-like types (I am including here `External.Source.t`)

In `workshop.ml` (task no. 3) I can adjust the type of `Workshop.presentation.poster` but the [solution might not improve the code](https://github.com/ocaml/ocaml.org/pull/2042#discussion_r1487478833) as well. IMO the only way to handle the fields of `Workshop.presentation` with `option` type is by creating another type from it, and use the latter in `Workshop.t.presentations` instead, only this time I would use ppx_stable and rename `presentation` to `presentation_metadata` or something more readable.

### DIFF
--- a/src/ocamlorg_data/data.mli
+++ b/src/ocamlorg_data/data.mli
@@ -392,8 +392,8 @@ module Release : sig
     kind : kind;
     version : string;
     date : string;
-    is_latest : bool option;
-    is_lts : bool option;
+    is_latest : bool;
+    is_lts : bool;
     intro_md : string;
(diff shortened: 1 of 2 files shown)


sabine (approved):
Thanks @egmaleta! :sparkles: Merging!

PR #2060, opened 2024-02-15
(docs) Basic Data Types: Add link to module Str

cuihtlauac:


data/tutorials/language/0it_01_basic_datatypes.md
```diff
@@ -131,7 +131,7 @@ Indexed access to string characters is possible using the following syntax:
 # "buenos dias".[4];;
 - : char : 'o'
 ```
-Operations on `string` values are provided by the [`Stdlib`](/api/Stdlib.html) and the [`String`](/api/String.html) modules.
+String related operations are provided by the [`Stdlib`](/api/Stdlib.html), the [`String`](/api/String.html), and [`Str`](/api/Stdlib.html) modules.
```
Octachron: The `Str` module is *not* part of the standard library. It is a separate library with known design issues. I don't think it is a good idea to cite it here.

cuihtlauac: I was not aware of the issues. What's planned deprecating or fixing?

Our scope is the OCaml _platform_ not only the standard library. Are you considering removing it from the platform?

P.S. Fixed Str link

Octachron: At the scope of the opam ecosystem, `Str` is not the reference library for regular expression either. 
Moreover, I don't think that the `Str` library has ever been part of the OCaml platform?
The future status of `Str` is mostly in limbo, by lack of people working on it.

cuihtlauac: > At the scope of the opam ecosystem, Str is not the reference library for regular expression either.

We probably don't want to pick an opam package outside of the platform. Except as an example of use. Opam is part of the platform.

> Moreover, I don't think that the Str library has ever been part of the OCaml platform?

Since it is bundled with the compiler and mentioned in [The OCaml API](https://v2.ocaml.org/releases/5.1/api/index.html), I assumed it was.

> The future status of Str is mostly in limbo, by lack of people working on it.

:-(

Octachron: Part of the problem is that writing a modern regexp library (with unicode property classes support) is a very large effort, and such effort is probably better hosted outside of the compiler.
And it is unclear if it would be worthwhile to invest efforts into very slowly updating the Str API to remove its reliance on global state since the end result would be a more usable but still ascii-only (and not PCRE) regexp library.

sabine:
Ok so for practical reasons... this is in the manual right now, and people use this successfully. Do we link or don't we link? IMO either of those is fine as long as we make a quick decision.

Octachron:
I know people that uses `Str` in anger (it is one of the dependencies of `ocamldoc` for instance), but I am not sure who is using the library "successfully"? And my personal advice for any newcomers trying to use `Str` is to switch to either another regexp library or a parser library.

sabine:
I take that as a point for closing this and not sending people to the `Str` module. Instead we can finish the cookbook and add an entry there that highlights such a library.

PR #2068, opened 2024-02-19
(ood-gen) Adjust type of fields with `option` type in Workshop

egmaleta:
Handles task no. 3 of #2051 and task left behind in #2042 (merged).

I think this is cleaner than my attempt at #2042 because of the use of `ppx_stable` to transform presentation metadata


sabine (approved):
Thanks @egmaleta! Yes, having separate metadata and output type here is good!

PR #2075, opened 2024-02-20
Advertise atom feed in <head> using <link>

### LINKED ISSUE #2074: Advertise atom feed in <head> using <link>
I was curious if there was a feed for https://ocaml.org/blog somewhere. My browser didn't notify me of any feeds. I thought that was odd, so I started trying different URLs without luck. Eventually I started grepping the ocaml.org source code and found https://ocaml.org/feed.xml

Anyway I think it would be great to add in `<head>`something like  `<link rel="alternate" href="/feed.xml" title="The OCaml Planet" type="application/atom+xml"/>` so the next person doesn't have to spend some 20 minutes searching for a feed.

cuihtlauac:
Fix #2074

### DIFF
--- a/src/ocamlorg_frontend/layouts/layout.eml
+++ b/src/ocamlorg_frontend/layouts/layout.eml
@@ -45,6 +45,7 @@ let base
 ?styles
 ~title
 ?canonical
+?alternate
 ?(active_top_nav_item: Header.active_nav_item option)
 ?(footer_html = "")
 inner =
@@ -54,7 +55,7 @@ inner =
       <meta charset="utf-8">
       <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
       <% (match description with | Some description -> %>
-      <meta name="description" content="<%s description %>">
+      <meta name="description" content="<%s description %>./src/ocamlorg_frontend/layouts/layout.eml">
       <% | None -> ()); %>
       <meta name="twitter:title" content="<%s title %>">
       <% (match description with | Some description -> %>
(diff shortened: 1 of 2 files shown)


src/ocamlorg_frontend/layouts/layout.eml
```diff
@@ -54,7 +55,7 @@ inner =
       <meta charset="utf-8">
       <meta name="viewport" content="width=device-width, initial-scale=1, shrink-to-fit=no">
       <% (match description with | Some description -> %>
-      <meta name="description" content="<%s description %>">
+      <meta name="description" content="<%s description %>./src/ocamlorg_frontend/layouts/layout.eml">
```
sabine:
```suggestion
      <meta name="description" content="<%s description %>">
```

looks like an accidental change

src/ocamlorg_frontend/layouts/layout.eml
```diff
@@ -73,6 +74,9 @@ inner =
       <% (match canonical with | Some canonical -> %>
       <link rel="canonical" href="https://ocaml.org<%s canonical %>">
       <% | None -> ()); %>
+      <% (match alternate with | Some alternate -> %>
+      <link rel="alternate" href="<%s alternate %>" title="The OCaml Planet" type="application/atom+xml">
+      <% | None -> ()); %>
```
sabine: Seems to be the wrong place to assume that alternate always is a link to the OCaml Planet.

I would rather add an `?(extra_head_html = "")` parameter on the main layout where we can add all kinds of varying content into the `<head>` section of the page. Either that, or make alternate a record `{href : string; title: string; type_ : string}`. However, I think this is a one-off situation to place a `<link rel="alternate" ...>`, so `extra_head_html` seems less complex and more flexible.

sabine (approved):

PR #2076, opened 2024-02-20
feat: create ocaml playground mobile view 

### LINKED ISSUE #1391: Create a usable mobile view for the Playground
The playground currently cannot be used on mobile devices. Instead of providing a mobile view, it displays a message that a mobile support is "coming soon".

This task requires making a layout inside `ocamlorg_frontend/src/pages/playground.eml` that works for the interaction with the Playground on mobile devices.

If you want to work on this, reply to this issue so we can talk a bit how this could look like.

The first step would be coming up a good layout (for example, having the code view on top, and the output window on the bottom?), and then our designer can help create styles for this layout, or you can experiment in Figma.

FatumaA:
This PR closes #1391 by Implementing a mobile view for the playground as per the provided design.

The PR:
- removes mobile check
- removes hidden class on mobile from the playground
- stacks the page view
- adds hero icons
- make button style changes
- refactors the positioning of the buttons and removes some unneeded logic

Screenshots:

Screenshot 2024-02-21 at 00 13 48


cuihtlauac:
I've tested it with my Pixel 8 Pro. I managed to write code and run it. First time I write caml code on a phone. This is awesome. I love it! Thanks @FatumaA 

Would it be possible to have another example? The one we have is nice, but it has a lot of comments, which is painful on a phone's screen.

cuihtlauac (approved):

FatumaA:
Thank you @cuihtlauac 😊
I think that's a good idea, I was also thinking of making the output pane draggable on mobile. Such that you can pull up the output screen if necessary.

What are your thoughts?

sabine:
Hey @FatumaA, this is really great! Seems to work just perfect! :sparkles: 

I removed some of the old inset shadows and reduced the padding on the code area. Merging this now. :+1:

Not sure if the added complexity of draggable output pane is worth it - however, you could give it a try in another PR - it's just not certain that it can be merged since we try and keep things as simple as we can. Maybe a simple button to collapse / expand the output panel would work?

PR #2080, opened 2024-02-21
New submission of the Database tutorial

F-Loyer:
Note: I may have "Resolved" proposition thinking they would be merged. (I am a beginner with GitHub)


cuihtlauac:
Can you rebase this? I've merged the fix on the typo in error guide in PR #2087

F-Loyer:
I have used the "Sync fork" option from the GitHub.

cuihtlauac:
@F-Loyer: We've been reviewing this beyond editorial edits. We believe this material would be better presented as a set of related cookbook entries.
1. Insert and create data
1. Create databases
etc.

Lwt processing should be a different and dedicated set of recipes, that can be referred to from the database receipes.

Custom binders and monadic stuff should refer to existing documentation and tutorials.

Your PR is beneficial because it greatly helps us understand and refine the future design of the cookbook. For instance, we now understand the set of used packages must be metadata, that will allow to display of recipes related to a package. In the case of your's doing so would more or less recreate this document.

P.S. afk next week

F-Loyer:
I guess that before adapting the database tutorial, I would need a rough approved plan.

Where Lwt processing (ppx_rapper has only Lwt/Async support) should be placed ? Cookbook/guide ? If in Cookbook, I guess I should have a pattern about the recipes which start with an introduction (I mean here "why dealing with Lwt ?") I guess most people use Lwt because it is demanded by other pieces of code (ppx_rapper, dream...). It will be more true with OCaml 5 and its simpler to use Eio.

I had a look to: https://rust-lang-nursery.github.io/rust-cookbook/database/postgres.html 

Especially: `for row in client.query("SELECT id, name, country FROM author", &[])?`

The `?` is related to the implicit error handling. If we want to match it with an OCaml notation, it will be the `let*` from the `Lwt_result` monad in a context of Lwt queries. (Without the Lwt context, it would have been a Result monad or a less acurate Result.get_ok post processing). Shall the recipices use this operator with a link to the Lwt section?

In many cases, snipsets are ok, but when using libraries or ppx_rewriter, the `utop` syntax is quite different than with dune. (`utop` use `#require` to load libraries and ppx... where dune uses two different stanzas). I guess that we should introduce a formalism about dune requirements. One should not have to infer them from the `utop` snipsets. (Rust seems simple: we have a crate... install it and use it. The OCaml platform with package/libraries/modules is far more complicated).

Note, the cookbook is not part of the main branch. Shall I have to propose something in the `cookbook` branch ? `staging` branch?


In comparison, the Rust cookbook only focus on SQLite and Postgres with two different libraries. I could have done the same. Probably no Lwt issues, but the added value of Caqti/ppx_rapper is the portability and the implicit type conversion. No need of `value.to_string()` like in the Rust cookbook. An Lwt version seems to be required since some environment (dream server...) demands it. 

(NOTA: I am working on a ppx_rapper_eio...).

PR #2095, opened 2024-02-26
Update typo + change from append to prepend

danipoma:


christinerose (approved):

data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
```diff
@@ -68,7 +68,7 @@ val u : int list = [1; 2; 3; 4]
 - : string list = ["this"; "is"; "mambo"]
 ```
 
-The lists' types, `int list` and `string list`, have been inferred from the type of their elements. Lists can be empty `[]` (pronounced “nil”). Note that the first list has been given a name using the `let … = …` construction, which is detailed below. The most primitive operation on lists is appending a new element at the front of an existing list. This is done using the “cons” operator, written with the double colon operator `::`.
+The lists' types, `int list` and `string list`, have been inferred from the type of their elements. Lists can be empty `[]` (pronounced “nil”). Note that the first list has been given a name using the `let … = …` construction, which is detailed below. The most primitive operation on lists is prepending a new element at the front of an existing list. This is done using the “cons” operator, written with the double colon operator `::`.
```
cuihtlauac: > The most primitive operation on lists is prepending a new element at the front of an existing list.

Replacing *appending* with *prepending* sounds good. However, isn't the second part of the sentence redundant if we do that? Also, isn't _preprend_ a more advanced term? The text needs to be exact and easy to understand.

data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
```diff
@@ -260,7 +260,7 @@ val range : int -> int -> int list = <fun>
...
+As indicated by its type `int -> int -> int list`, the function `range` takes two integers as arguments and returns a list of integers as result. The first `int` parameter, `lo`, is the range's lower bound; the second `int` parameter, `hi`, is the higher bound. If `lo > hi`, the empty range is returned. That's the first branch of the `if … then … else` expression. Otherwise, the `lo` value is prepended to the list created by calling `range` itself; this is recursion. Prepending is achieved using `::`, the cons operator in OCaml. It constructs a new list by adding an element at the front of an existing list. Progress is made at each call; since `lo` has just been prepended at the head of the list, `range` is called with `lo + 1`. This can be visualised this way:
```
cuihtlauac: Same here. We should probably have either: 

> has been prepended to the list

or 

> has been appended at the head of the list

cuihtlauac (changes requested):

cuihtlauac (approved):

data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
christinerose: Agreed about it being a more advanced term. 
How about: 

The most primitive operation on lists is to add a new element at the front of an existing list.

data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
christinerose: Let's look at this section a little more closely. 

"Otherwise, the `lo` value is prepended to the list created by calling `range` itself; this is recursion."

This is trying to say that the `lo` value is added to the front of the list that was previously created by calling `range` --- correct? If so, I'll work on making that more clear. Is the "itself" necessary here?

"Prepending is achieved using `::`, the cons operator in OCaml. It constructs a new list by adding an element at the front of an existing list."

IIUC, this says: The OCaml cons operator (`::`) constructs a new list by adding an element at the front of an existing list" 

This is currently how it reads. The "It" in the second sentence refers to `::`, right?


Unless "prepending" is a specific term for this action in programming, it might be best to avoid it.

data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
sabine:
```suggestion
The lists' types, `int list` and `string list`, have been inferred from the type of their elements. Lists can be empty `[]` (pronounced “nil”). Note that the first list has been given a name using the `let … = …` construction, which is detailed below. The most primitive operation on lists is to add a new element at the front of an existing list. This is done using the “cons” operator, written with the double colon operator `::`.
```

sabine (approved):

PR #2099, opened 2024-02-26
Cookbook : filesystem

F-Loyer:
I am not sure this cookbook is very useful... The `Stdlib.Sys` if fair enough (and more precise).

Perhaps I should add `is_regular_file`, but it is only available since 5.1.


christinerose (approved):

christinerose (approved):

cuihtlauac:
Set upstream from ocaml:cookbook-filesystem

cuihtlauac (changes requested):
Set upstream from https://github.com/ocaml/ocaml.org/tree/cookbook-filesystem

PR #2117, opened 2024-02-29
Omit mention of Merlin

yawaramin:
There's no need to mention Merlin at the beginning stage. It will just confuse people eg https://x.com/synecdokey/status/1763039629047894273?s=20


sabine (approved):
Thanks @yawaramin ! :sparkles:

PR #2123, opened 2024-02-29
Cookbook : Web / simple HTTP client

F-Loyer:
A very simple cookbook...


data/cookbook/web-programming/simple-http-client/00-camlnet.ml
```diff
@@ -0,0 +1,26 @@
...
+packages:
+- name: "camlnet"
+  version: "4.1.9-2"
+- name: "conf-gnutls"
+  version: "1"
+libraries:
+- netclient
+  nettls-gnutls
+discussion: |
+  - **Understanding `ocamlnet`:** The `ocamlnet` provides a rich set of networking libraries that cover many protocols. It provides some modules with a rich set of parameters. But for simple needs, the `Nethttp_client.Convenience` provides simple functions for basic interaction with the server. Note: These functions don't handle redirections.
```
christinerose:
```suggestion
  - **Understanding `ocamlnet`:** The `ocamlnet` provides a rich set of networking libraries that cover many protocols. It has some modules with a rich set of parameters. But for simple needs, the `Nethttp_client.Convenience` gives simple functions for basic interaction with the server. Note: These functions don't handle redirections.
```
Suggestion to avoid the use of "provide" multiple times

data/cookbook/web-programming/simple-http-client/00-camlnet.ml
```diff
@@ -0,0 +1,26 @@
...
+  - **Reference:** The `Convenience` module used for simple HTTP interaction is documented on [the ocamlnet site](http://projects.camlcity.org/projects/dl/ocamlnet-4.1.9/doc/html-main/Nethttp_client.Convenience.html)
```
christinerose: Is `ocamlnet` the official term (in monospace). I didn't see it on that linked website. If so, this one also needs to be in monospace. If not, it would follow the OCaml format used in OBuilder, OCamlFormat, etc., to write it as OCamlNet. @cuihtlauac, what do you think?

christinerose (commented):
A few questions & suggestions

samoht:
I'm not convinced that `ocamlnet` should be the framework recommended for new users. Is the goal to show all the existing frameworks? Or should we recommend one?

data/cookbook/web-programming/simple-http-client/00-camlnet.ml
F-Loyer: `ocamlnet` is the name of the package from a, Opam view. [The Ocamlnet](http://projects.camlcity.org/projects/ocamlnet.html) spell it Ocamlnet.

F-Loyer:
> I'm not convinced that `ocamlnet` should be the framework recommended for new users. Is the goal to show all the existing frameworks? Or should we recommend one?

But the convenience module of Ocamlnet is "convenient" - see the proposed snippet : 2 lines - (but quite basic). I have a Cohttp/Lwt snippet if it is prefered...

tmattio:
I think ocamlnet is unmaintained and incompatible with OCaml 5, so I would also suggest not advertising it as a viable option in the Cookbook :)

data/cookbook/web-programming/simple-http-client/00-camlnet.ml
F-Loyer: The Ocamlnet library has been changed for cohttp-lwt... it doesn't fit in a 2 lines program anymore...

cuihtlauac:
Rebased branch available in: https://github.com/ocaml/ocaml.org/tree/cookbook-httpclient

cuihtlauac (changes requested):
Set upstream from https://github.com/ocaml/ocaml.org/tree/cookbook-httpclient

cuihtlauac:
Set upstream from https://github.com/ocaml/ocaml.org/tree/cookbook-httpclient

PR #2127, opened 2024-03-01
Cookbook : fix in Lwt (type mismatch with iter_s/iter_p functions)

F-Loyer:


cuihtlauac (approved):
Thanks @F-Loyer, this was bugged.

PR #2132, opened 2024-03-04
Improve Events Directory

sabine:
* removes past event
* makes date and time format more clear - all times and dates are to be entered in UTC
* display start and end times, if applicable
* move logic that chooses which events to display to `handler.ml`


tmattio:
I would recommend not deleting historical data. Maybe a flag to hide past events would be more appropriate?

sabine:
They're hidden anyways since their start/end dates are in the past. But if you prefer to keep them around, no problem.

PR #2133, opened 2024-03-04
Cookbook compression / decompression

F-Loyer:


data/cookbook/text-processing/regular-expressions/00-str.ml
```diff
@@ -0,0 +1,33 @@
+---
+packages: []
+libraries:
+- str
+discussion: |
+  - **Understanding `Str`: `str` is a library which is brought with OCaml. It contains many functions which deal with regular expression. The documentation of the `Str` module is in the [API reference](https://v2.ocaml.org/api/Str.html).
```
christinerose, line 6:
```suggestion
  - **Understanding `Str`: `str` is a library that comes with OCaml. It contains many functions that deal with regular expression. The documentation of the `Str` module is in the [API reference](https://v2.ocaml.org/api/Str.html).
```

data/cookbook/text-processing/utf8-text-processing/00-camomile.ml
```diff
@@ -0,0 +1,64 @@
...
+   any Unicode character: *)
+let () = assert (Camomile.UChar.code (Camomile.UTF8.get "déjà" 3)
+                    = 224)
+
+(* Using bytes oriented indexes. Using functions which works
+   directly with the byte indexes of characters can be more efficient
+   than using functions like `get` or `length` that have to parse
+   the string counting characaters. `first`, `next`, `prev`, `last`
+   and `look` all deals with bytes indexes. `nth` convert of position
+   expressed as a character index to a byte index. *)
```
christinerose, line 35: Should `nth` be in monospace like it is here? If so, does it also need to be in monospace in li 24?

This sentence doesn't make sense. I suspect there's a typo.
"`nth` convert of position expressed as a character index to a byte index."

Should it be "`nth` converts a position expressed as a character index to a byte index" ?

data/cookbook/compression/gzip-decompress/00-camlzip.ml
```diff
@@ -0,0 +1,30 @@
...
+- name: "camlzip"
+  version: "1.11"
+libraries:
+- camlzip
+discussion: |
+  - **About `camlzip`:** `camlzip` is a package that brings 3 modules: `Zip`, `Zlib`, and `Gzip`, which permit compression and decompression. The `Gzip` module has two decompression functions, depending on the origin of the compressed stream (file or output channel).
+  - **Reference:** The [`Gzip` `.mli` file](https://github.com/xavierleroy/camlzip/blob/master/gzip.mli) is simple and well documented.
+---
+
+(* The gunzip compression function opens both files (`source` and `dest`) and transfers bytes through a bytes buffer with the appropriate libraries *)
```
christinerose:
```suggestion
(* The `Gunzip` compression function opens both files (`source` and `dest`) and transfers bytes through a bytes buffer with the appropriate libraries *)
```
I suspect this needs to be in monspace, but should it also be capitalised like `Gzip` is?

christinerose (commented):
A few questions for clarity. I'll review again after they're addressed.

F-Loyer:
I don't understand why are there so many conflicts... I may probably have to extract the new files, and create a new branch from a fresh cookbook branch.  To do it, perhaps I should wait the cookbook branch to be updated (merging the other PR)

christinerose:
> I don't understand why are there so many conflicts... I may probably have to extract the new files, and create a new branch from a fresh cookbook branch. To do it, perhaps I should wait the cookbook branch to be updated (merging the other PR)

So sorry if I did something wrong! Please let me know if you figure it out, and I'll make a note not to do that again!

F-Loyer:
A have created a new fresh branch from cookbook. added the compression cookbook files... commit and push -f. It seems ok.

```
 1408  git checkout cookbook
 1409  git branch -D cookbook-unzip
 1410  git checkout -b cookbook-unzip
 1413  cp -r /tmp/compression .
 1416  cp /tmp/cookbook_categories.yml .
 1418  git add .
 1419  git status
 1423  git commit -m 'Cookbook / Compression'
 1425  git push -f floyer
```

cuihtlauac:
The cookbook design is not yet final. Thanks to your contributions we are building a better understanding of what we envision.

Based on where we are now, I believe this PR should be only two tasks:
- compress single file
- decompress single file

Then, there are different ways to have several things:
- per flavour, e.g. GZIP vs ZIP or others -> this should remain inside each recipe. Currently we can't address this because the current implementation of recipes can only have a single code sample. This will be changed
- per task, e.g. several file or folders -> one group of recipes per task
- per package, e.g. curly vs cohttp -> one recipe per package

In UML (sorry) terminology:
* a recipe has a task
* a task has a category

Hierarchically:
- Category: Compression
  - Task: Compress File
    - Recipe: Using Camlzip 
  - Task: Decompress File
    - Recipe: Using Camlzip

In that framework:
  * Addresssing compression of folders adds two tasks with corresponding recipes.
  * Supporting another library adds one recipe per task

F-Loyer:
I guess that about compression/decompression, a decomposition depending of the format has a sense. When we decompress a file, the format is imposed, then a recipes for ZIP and an other for Gzip would address two different usages. For compression, we may have the choice... expected when compressing multiple files (ZIP handles archives... Gzip must be used with tar (not sure it is quite handy to program).

However, studiying the decompostion case would be more interesting with Regular expressions. I have proposed 2 recipes for the same objective. One with the vanilla OCaml and its Str module. And the other, simpler, with ppx_regexp+re. Here, the decomposition is mandated because of the different sets of involved packages/libraries.

On the https://rust-lang-nursery.github.io/rust-cookbook/  we have Category / Recipes, recipes are quite decomposed. Perhaps too much (1/ sorting integer, 2/ sorting floats, 3/ sorting struct ... where I have proposed a unique recipe : list/array/with custom `compare` function). Idem for the database (5 recipes... 2 for SQlite, 2 for Postgres compared to 2 : Sqlite and Caqti). Note: I plan to add transaction/Caqti... and I am not sure of the way to proposed it. A dedicated recipes would share many lines with the Caqti recipes. Note: the Caqti recipes tell us how to execute the different queries (this could have been spread among different recipes), but also how to sequence multiple queries (with `lwt` an example would be best). This explain the unique Caqti recipe.

cuihtlauac:
Rebased branch available in: https://github.com/ocaml/ocaml.org/tree/cookbook-rebased

cuihtlauac (changes requested):
Set upstream from https://github.com/ocaml/ocaml.org/tree/cookbook-unzip

cuihtlauac:
Set upstream from https://github.com/ocaml/ocaml.org/tree/cookbook-unzip

PR #2146, opened 2024-03-07
(docs) Add link for instances of Array

### LINKED ISSUE #2088: Refer to both tutorials and API references as links
1. Outside of code blocks, transform all references to standard manual modules, types and functions into links
2. Sentences referring to tutorials matching an standard library module should also contain a direct link to the API reference

Examples:

In [https://ocaml.org/docs/basic-data-types](../blob/main/data/tutorials/language/0it_01_basic_datatypes.md)

Good, links:

```md
Operations on `float` are provided by the [`Stdlib`](/api/Stdlib.html) and the [`Float`](/api/Float.html) modules.
```

Bad, no links:

```md
The predefined function `fst` returns the first element of a pair, while `snd` returns the second element of a pair.
```

rmeis06:
Hello! I'm an outreachy applicant.

I've updated mentions of the Array module to be a link where they were not already. This should partially fix https://github.com/ocaml/ocaml.org/issues/2088

@cuihtlauac  @SaySayo Does this look correct to yinz?

### DIFF
--- a/data/tutorials/language/0it_01_basic_datatypes.md
+++ b/data/tutorials/language/0it_01_basic_datatypes.md
@@ -575,7 +575,7 @@ type json =
   | Object of (string * json) list
 ```
 
-Both constructors [`Array`](/api/Array.html) and `Object` contain values of type `json`.
+Both constructors `Array` and `Object` contain values of type `json`.
 
 Functions defined using pattern matching on recursive variants are often recursive too. This function checks if a name is present in a whole JSON tree:
 ```ocaml
@@ -589,7 +589,7 @@ Functions defined using pattern matching on recursive variants are often recursi
 val has_field : string -> json -> bool = <fun>
 ```
 
-Here, the last pattern uses the symbol `_`, which catches everything. It returns `false` on all data that is neither [`Array`](/api/Array.html) nor `Object`.
+Here, the last pattern uses the symbol `_`, which catches everything. It returns `false` on all data that is neither `Array` nor `Object`.
 
 ### Polymorphic Data Types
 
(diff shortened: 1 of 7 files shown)


christinerose (approved):

data/tutorials/language/0it_01_basic_datatypes.md
cuihtlauac, line 1: These occurences of the string `Array` do not refer to the Array module from the standard library but to the Array constructor (line 566) from the type json (line 560). Links should not be added here

data/tutorials/language/0it_05_imperative.md
cuihtlauac, line 1: This one is tricky, the two occurences of Array here refer to the module defined line 743, not the module from the standard library, they should not be changed.

data/tutorials/language/1ms_01_functors.md
```diff
@@ -477,7 +477,7 @@ Run the `dune utop` command. Once inside the toplevel, enter the following comma
 - : int list = [0; 1; 3; 6; 10; 15; 21; 28; 36; 45]
 ```
 
-Modules `Array` and `List` appear augmented with `Array.scan_left` and `List.scan_left`. For brevity, the output of the first two toplevel commands is not shown here.
+Modules [`Array`](/api/Array.html) and `List` appear augmented with `Array.scan_left` and `List.scan_left`. For brevity, the output of the first two toplevel commands is not shown here.
```
```suggestion
Modules `Array` and `List` appear augmented with `Array.scan_left` and `List.scan_left`. For brevity, the output of the first two toplevel commands is not shown here.
```

This one does not refer to the version from the standard library. No link should be added.

cuihtlauac (changes requested):

rmeis06:
@cuihtlauac I believe I have the documents updated as described. A question for me to better understand ocaml, a standard library can be overwritten in a local program using `module <module name>`? And I think I'll have to learn more about what a constructor is to understand the Array that was defined in json.

cuihtlauac (approved):
Thanks @rmeis06

PR #2150, opened 2024-03-08
Serve OCaml Compiler Manuals

mtelvers:
* Include the precompiled html manuals from [ocaml-web/html-compiler-manuals](https://github.comocaml-web/html-compiler-manuals) into the `Dockerfile`.  The `RUN` line determines which branches exist and clones them all into local directories: `/manual/4.14`, `/manual/5.0`, etc; updating this is unnecessary when a new version is released.
* Updated the code to serve `/manual` and below from the local filesystem
* Added default name `index.html` when the directory is requested
* Updated the default manual link `/manual` to serve from these files rather than redirect to v2.ocaml.org


src/ocamlorg_web/lib/redirection.ml
```diff
@@ -449,6 +449,25 @@ let from_v2 =
     ("/releases/latest/api/index.html", Url.api_with_version latest_version);
   ]
 
+let default_index_html =
+  [
+    ("/manual", Url.manual_lang latest_version);
```
shakthimaan: We will need to add one for `latest`.

sabine: We don't need a redirect for `latest`, we need to serve this via Dream under `/manual/latest`. A redirect would mean that the URL in the browser would change to the specific version.

src/ocamlorg_web/lib/redirection.ml
```diff
@@ -459,12 +478,13 @@ let redirect_p pattern =
 let fwd_v2 origin =
   Dream.get origin (fun req -> Dream.redirect req (Url.v2 ^ origin))
 
+let serve url path = Dream.get url (Dream.static path)
+
 let manual =
   [
+    serve "/manual/**" "/manual/";
```
sabine: I don't think we should be configuring pages to be served from the `redirection.ml` module. Instead, this should be put into `router.ml`.

src/global/url.ml
```diff
@@ -46,7 +46,9 @@ let minor v =
 
 let v2 = "https://v2.ocaml.org"
 let manual_with_version v = v2 ^ "/releases/" ^ minor v ^ "/htmlman/index.html"
-let manual = "/releases/latest/manual.html"
+let manual = "/manual"
+let manual_lang v = "/manual/" ^ minor v ^ "/index.html"
+let manual_api v = "/manual/" ^ minor v ^ "/api/index.html"
```
sabine: `index.html` is the default page looked up by the browser, so we don't need to append this here. Shorter URLs tend to be better. I would rather redirect the other way.

cuihtlauac:
Using the following on the upstream `https://github.com/ocaml-web`:
```bash
git grep -ho 'href=\"https://[._a-z0-9/\-]*' | sort -u
```

I found the following hardcoded urls:
```
https://ocaml.org/
https://ocaml.org/releases/
https://ocaml.org/releases/4.13/ocaml-4.13-refman-html.tar.gz
https://ocaml.org/releases/4.13/ocaml-4.13-refman.info.tar.gz
https://ocaml.org/releases/4.13/ocaml-4.13-refman.pdf
https://ocaml.org/releases/4.13/ocaml-4.13-refman.txt
https://ocaml.org/releases/4.14/ocaml-4.14-refman-html.tar.gz
https://ocaml.org/releases/4.14/ocaml-4.14-refman.info.tar.gz
https://ocaml.org/releases/4.14/ocaml-4.14-refman.pdf
https://ocaml.org/releases/4.14/ocaml-4.14-refman.txt
https://ocaml.org/releases/5.0/ocaml-5.0-refman-html.tar.gz
https://ocaml.org/releases/5.0/ocaml-5.0-refman.info.tar.gz
https://ocaml.org/releases/5.0/ocaml-5.0-refman.pdf
https://ocaml.org/releases/5.0/ocaml-5.0-refman.txt
https://ocaml.org/releases/5.1/ocaml-5.1-refman-html.tar.gz
https://ocaml.org/releases/5.1/ocaml-5.1-refman.info.tar.gz
https://ocaml.org/releases/5.1/ocaml-5.1-refman.pdf
https://ocaml.org/releases/5.1/ocaml-5.1-refman.txt
https://ocaml.org/releases/5.2/ocaml-5.2-refman-html.tar.gz
https://ocaml.org/releases/5.2/ocaml-5.2-refman.info.tar.gz
https://ocaml.org/releases/5.2/ocaml-5.2-refman.pdf
https://ocaml.org/releases/5.2/ocaml-5.2-refman.txt
https://ocaml.org/releases/5.3/ocaml-5.3-refman-html.tar.gz
https://ocaml.org/releases/5.3/ocaml-5.3-refman.info.tar.gz
https://ocaml.org/releases/5.3/ocaml-5.3-refman.pdf
https://ocaml.org/releases/5.3/ocaml-5.3-refman.txt
```

All but the first two trampolines to v2. We should all turn them into relative links.

Dockerfile
```diff
@@ -34,6 +34,8 @@ COPY --from=build /home/opam/_build/default/src/ocamlorg_web/bin/main.exe /bin/s
 
 COPY playground/asset playground/asset
 
+RUN git clone https://github.com/ocaml-web/html-compiler-manuals /manual && for a in $(git -C /manual branch -r --format='%(refname:lstrip=3)') ; do git -C /manual submodule add ./ $a && git -C /manual/$a checkout $a; done
```
cuihtlauac: This seems consistent with the `README.md` in `https://github.com/ocaml-web/html-compiler-manuals`. It says:
> For every compiler version, there is a branch named like exactly like the compiler version. The branch contains the HTML pages for the compiler manual to be served by OCaml.org.

However, I only see a single branch upstream. Am I missing something or can we do the `git clone` without iterating over branches?

src/global/url.ml
```diff
@@ -46,7 +46,9 @@ let minor v =
...
 let api_with_version v = v2 ^ "/releases/" ^ minor v ^ "/api/index.html"
 let api = "/releases/latest/api/index.html"
```
cuihtlauac:
```suggestion
let api_with_version v = "/releases/" ^ minor v ^ "/api"
let api = "/releases/latest/api"
```

cuihtlauac (dismissed):

Dockerfile
shakthimaan: The requirements have changed. We will need to update the README to reflect the same. Thanks for your review!

src/global/url.ml
mtelvers: The browser doesn't request the default page.  If configured, the server returns a default page when the browser doesn't specify a specific page.

src/global/url.ml
mtelvers: I agree with this change, but I think it's outside the scope of this PR.  A whole exercise will need to be carried out to remove redirects to `v2`.  See `fwd_v2` in `src/ocamlorg_web/lib/redirection.ml`.

src/ocamlorg_web/lib/redirection.ml
mtelvers: I have moved this to `router.ml`

Dockerfile
mtelvers: With the new repo layout the `RUN` line can be greatly simplified.

src/global/url.ml
cuihtlauac: I agree removing obsolete redirects should be in another PR. But this only allows the links to the Standard Library API (such as found in https://ocaml.org/docs, the footer and several other places) to point to the local content instead of V2.

src/global/url.ml
shakthimaan: When manually testing instructions given in the Dockerfile, I am getting a 404 for /manual/5.1/index.html. What could I be missing?

mtelvers: Are your manuals cloned to the top level root, aka /, of your test machine?

cuihtlauac: Could we have a setup with an environment variable (maybe `OCAMLORG_MANUAL_PATH` ?) such as in `src/ocamlorg_web_lib/config.ml`? That would be easier for debugging.

shakthimaan: > Are your manuals cloned to the top level root, aka /, of your test machine?

Ohh! Okay. After copying to /manuals, it works fine. Thanks!

cuihtlauac: This branch: https://github.com/ocaml/ocaml.org/tree/html-compiler-manuals-config contains a patch that allows to set the manual path in a shell environment variable

mtelvers: Thanks, I'm on leave this week.  I have cherry-picked your commit to this PR.

cuihtlauac:
The changes I'm proposing to handle api urls are in this branch: https://github.com/ocaml/ocaml.org/tree/html-compiler-manuals-api

sabine:
@cuihtlauac I don't see a reason why we should introduce a config variable `OCAMLORG_MANUAL_PATH` for the manual path. The rest looks useful on first glance, but haven't checked in detail.

cuihtlauac:
> @cuihtlauac I don't see a reason why we should introduce a config variable `OCAMLORG_MANUAL_PATH` for the manual path. The rest looks useful on first glance, but haven't checked in detail.

That allows developers to put their local copy of `html-compiler-manual` wherever they want/need (not only `/`).

mtelvers:
I have added @cuihtlauac latest commit to this PR and included a commit to update line 50 of `src/global/url.ml` as the refactoring of these lines has made `api_with_version` the same as `manual_with_version` where they should be different:

```
-let api_with_version v = "/manual/" ^ minor v ^ "/index.html"
+let api_with_version v = "/manual/" ^ minor v ^ "/api/index.html"
```

@shakthimaan I note that `search_icon.svg` is missing from the `api` folder for all the versions of the manuals except for 4.12 and 4.13.  I think it needs to be included.

shakthimaan:
> search_icon.svg is missing from the api folder

The search_icon.svg has been added to the respective api/ folders.

sabine (approved):
LGTM, on staging this seems to work fine

Octachron:
Thanks a lot for the hard work on integrating the reference manual !

PR #2160, opened 2024-03-11
(ui) Activate Dark Mode

sabine:
Depends on https://github.com/ocaml/ocaml.org/pull/2159. Also, all the dark mode pages on staging.ocaml.org need to be reviewed.

### DIFF
--- a/src/ocamlorg_frontend/layouts/layout.eml
+++ b/src/ocamlorg_frontend/layouts/layout.eml
@@ -103,7 +103,12 @@ inner =
       <title><%s title %></title>
     </head>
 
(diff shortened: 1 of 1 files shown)


tmattio:
Had a look at staging, the only place where I could see an issue was the package documentation pages.

sabine:
still, after this new patch I just merged a minute before opening this PR? I think you probably saw the pre-deployment staging

sabine:
I saw and fixed a little issue with package documentation breadcrumbs, looks to me like this is good enough to deploy.

Issue #2141, opened 2024-03-06
Outreachy Summer 2024


SaySayo:
Hi all,

here's some information about the Outreachy application period. First of all: welcome! 👋

For non-applicants confused by this issue, see [this section](#for-non-applicants).

## For Applicants

Thanks for choosing to contribute to the OCaml ecosystem! This is the right place to start if you're considering applying to the following project: 
- Run OCaml Exercises Everywhere

### Quick Installation
The quickest way to install OCaml currently is using the OCaml Platform Installer.

You'll need to install some dependencies if they're not already installed. This can be done easily using a package manager.

If you're running an Ubuntu/Debian system you can run the following command in your terminal. If you're using Windows, here are the [specific instructions](https://ocaml.org/docs/ocaml-on-windows), but it is probably easier and quicker to use WSL2 on Windows and then using the Linux/macOS instructions.

`sudo apt install bzip2 make gcc bubblewrap rsync patch curl unzip`

After your dependencies are installed run the below install script.

`bash < <(curl -sL https://ocaml.org/install-platform.sh)`

Install the Platform tools.

`ocaml-platform`

### Other Installation Methods
If the OCaml Platform installation method does not suit you then you can take a look at the other installation methods described on ocaml.org: [here](https://ocaml.org/install) or [there](https://ocaml.org/docs/up-and-running#installing-ocaml).

## Contributions
Before being able to fully contribute to the project, here are simple steps you must follow!

### First step: seting up OCaml
If you want to apply to any of the those projects, the first step for you will be to compile OCaml. Please, read through [Installing OCaml on Your System](https://github.com/ocaml/ocaml.org/issues/1553#Installing-OCaml-on-Your-System). If you're using Windows, here are the [specific instructions](https://ocaml.org/docs/ocaml-on-windows), but it is probably easier and quicker to use WSL2 on Windows and then using the Linux/macOS instructions.

Now you should be able to go through the basics of OCaml, please follow this tutorial: [Your First Day with OCaml](https://ocaml.org/docs/first-hour).

For more details about the OCaml language you can read this guide [about the language](https://ocaml.org/docs/language).

Try to pick up a few [exercises to practice](https://ocaml.org/exercises) along the way!

### Second step: seting up ocaml.org locally
Once OCaml is successfully installed, you also need to install an opam switch that uses a version of OCaml that ocaml.org currently supports, and install the dependencies.

[...]

PR #2163, opened 2024-03-11
Linking Bigarray references

### LINKED ISSUE #2088: Refer to both tutorials and API references as links

rmeis06:
This should partially fix https://github.com/ocaml/ocaml.org/issues/2088

I found some references to Bigarray in the following files:
news/ocaml/ocaml-2022-02.md
releases/4.07.0.md

It didn't seem like I should update them because they're release notes and news, so I left them as is. But I can update them if that's desired.

@cuihtlauac @SaySayo How does this look?

### DIFF
--- a/data/tutorials/language/5rt_00_memory_representation.md
+++ b/data/tutorials/language/5rt_00_memory_representation.md
@@ -179,7 +179,7 @@ collection process described earlier.
 The `size` field records the length of the block in memory words. This is 22
 bits on 32-bit platforms, which is the reason OCaml strings are limited to 16
 MB on that architecture. If you need bigger strings, either switch to a
-64-bit host, or use the `Bigarray` module.
+64-bit host, or use the [`Bigarray`](/api/Bigarray.html) module.
 
 The 2-bit `color` field is used by the GC to keep track of its state during
 mark-and-sweep collection.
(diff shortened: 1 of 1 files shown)


SaySayo:
Hi @rmeis06 its good practice to indicate on the issues prior to attempting it as sometimes other applicants have indicated prior interest in attempting it.

rmeis06:
Alright, sorry I don't know that I understood the instructions in https://github.com/ocaml/ocaml.org/issues/2141 under "Third step: a good first issue" as it asks only to open a PR, rather than requesting to be assigned an issue.

christinerose (approved):

cuihtlauac (approved):
Looks good to me. Thnaks @rmeis06

PR #2164, opened 2024-03-11
Suggest colors for syntax

Ozyugoo:


practice/004/ex.ml
```diff
@@ -0,0 +1,16 @@
...
+open OUnit2
+
+module type Testable = sig
+  val length : 'a list -> int
+end
+
+module Make(Tested: Testable) : sig val v : test end = struct
+  let v = "length" >::: [
+    "nil" >:: (fun _ -> assert_equal (Tested.length []));
+    "cons" >:: (fun _ -> assert_equal (Tested.length ["a"; "b"; "c"]));
```
cuihtlauac, line 11: You must specify the expected return value in both cases here

cuihtlauac (changes requested):
Thanks @Ozyugoo. Yes this is going in the right direction. Make sure compilation is successful before pushing a PR. Also you should and newlines characters at end of file, when missing

cuihtlauac:
You have mixed two PRs in this one. Work on PR #2168 must be made in a separate PR

practice/004/ex.ml
Ozyugoo: Thank you, on it.

Ozyugoo:
> You have mixed two PRs in this one. Work on PR #2168 must be made in a separate PR

I will do that, thank you

PR #2175, opened 2024-03-13
Implement copy functionality for exercises

### LINKED ISSUE #891: Add exercises tests

ghost:
To fix #891

### DIFF
--- a/src/ocamlorg_frontend/css/styles.css
+++ b/src/ocamlorg_frontend/css/styles.css
@@ -39,7 +39,7 @@ body {
 }
 
 .container-fluid.wide {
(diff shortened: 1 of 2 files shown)


cuihtlauac:
Thanks, @amarachigoodness74, this is going into the right direction.

Could you reuse the same icon as in package search?

sabine:
Closing because the PR appears abandoned.

PR #2186, opened 2024-03-13
Chore: suggest color scheme for syntax.

Ozyugoo:


Ozyugoo:
@sabine kindly check this if it aligns with what you expect.

Issue #2188, opened 2024-03-13
The code in Modules chapter doesn't compile


mikhailazaryan:
The code provided [here](https://ocaml.org/docs/modules#module-signatures-are-types) doesn't compile with the following message:
Signature mismatch:
       Modules do not match:
         sig val message : string val print : unit -> unit end
       is not included in
         HelloType
       The value `hello' is required but not provided

cuihtlauac:
Thanks @mikhailazaryan, this is indeed a bug.

mikhailazaryan:
I guess it works with the code from the pull request linked here, doesn't it? I mean, if the "print" function would just be renamed to "hello"

mikhailazaryan:
Alright, seen your answer

PR #2189, opened 2024-03-13
Minor fix

mikhailazaryan:
Solves #2188 by means of refactoring

### DIFF
--- a/data/tutorials/language/1ms_00_modules.md
+++ b/data/tutorials/language/1ms_00_modules.md
@@ -345,7 +345,7 @@ end
 
 module Hello : HelloType = struct
   let message = "Hello"
-  let print () = print_endline message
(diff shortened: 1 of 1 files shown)


cuihtlauac (changes requested):
Shouldn't it be the signature `HelloType` that's changed to list `print` instead of `hello`?

It would allow compiling `glasgow.ml` against the 3rd version of `florence.ml` without changes. Also, it would be consistent with the text:
> First, we define a module type called HelloType, which defines _the same module interface_ as before.

mikhailazaryan:
Yes, certainly

cuihtlauac:
Great! Thanks @mikhailazaryan

cuihtlauac (approved):

PR #2191, opened 2024-03-13
Correctly invert text on "Is OCaml Web" page

### LINKED ISSUE #2201: Dark mode unreadable in some pages
Concretely, I see dark text over dark mode on the page https://ocaml.org/docs/is-ocaml-web-yet.

Screenshot from 2024-03-14 11-18-03

Relatedly, it will be useful to have some sort of toggle button to enable and disable dark mode. If I understood correctly, right now the mode is being pulled from the system defaults.

SquidDev:
The "Is OCaml Xyz Yet?" pages (currently only [Is OCaml Web Yet?](https://ocaml.org/docs/is-ocaml-web-yet#lower-web-stack)) are not correctly styled in dark mode, leading to hard to read text.

A screenshot of 'Is OCaml Web Yet?' The page background is in black (dark mode being enabled), but the body still uses the light-mode theme (also being dark), making it unreadable.

This commit adds the missing tag `dark:prose-invert` class to the page - a rudimentary search shows that this is the only page missing it. In order for the heading and tagline to also be styled, I've moved them inside the prose tag (following `tutorial.eml`, which is used for most pages in the "Guides" section).

This does slightly change the layout of the page (the header is slightly smaller), but this is more consistent with other pages in this section, so feels like a win. Happy to revert this if preferred though!

A screenshot of 'Is OCaml Web Yet?' with this commit applied. Once again, the page is in dark mode, but now the text is white and thus readable!

### DIFF
--- a/src/ocamlorg_frontend/pages/is_ocaml_yet.eml
+++ b/src/ocamlorg_frontend/pages/is_ocaml_yet.eml
@@ -22,9 +22,10 @@ Learn_layout.three_column_layout
 ~current:Learn_layout.Guides
 ~left_sidebar_html:(Some(left_sidebar ~current_tutorial:(Some (Printf.sprintf "is-ocaml-%s-yet" meta.id)) ~tutorials))
 ~right_sidebar_html:(Some (Toc.render toc)) @@
-<h1 class="font-bold mb-8"><%s meta.question %></h1>
-<p class="mt-6 text-xl leading-8"><%s meta.answer %></p>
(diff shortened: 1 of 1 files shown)


cuihtlauac (approved):
Thanks @SquidDev, this looks good to me

stuckinforloop:
The same issue is there on the docs pages as well. For instance, look at the following example:

### Dark Mode
Screenshot 2024-03-14 at 17 50 42

### Light Mode
Screenshot 2024-03-14 at 17 52 08

Issue #2198, opened 2024-03-14
Turn exercices 013 to 022 into practice


cuihtlauac:
This task consists of doing 10 exercises instead of one, as described in issues #2167, #2168, #2170, #2172, #2173, #2176, #2177, #2178, #2179 and #2180. Create 10 folders (from 013 to 022), and populate them as the others. Makes sure `dune build --root .` and `dune test --root . answer` work.

* [X] 013 #2204
* [X] 014 #2206
* [X] 015 #2207
* [X] 016 #2208
* [X] 017 #2209
* [X] 018 #2210
* [X] 019 #2213
* [X] 020 #2214
* [X] 021 #2215
* [X] 022 #2216

divyankachaudhari:
I'd like to work on this. Thank you.

cuihtlauac:
Yes @divyankachaudhari, you may work on this one. Please do not use git push forcefully in your PR, adding commits is easier.

divyankachaudhari:
@cuihtlauac noted 🫡 I have a question, I'll create separate commits but do I include all the exercises in one PR itself?

cuihtlauac:
Let's create separate PRs it will be easier to review

divyankachaudhari:
Opened PRs for all 😊

cuihtlauac:
All PR merged, closing

PR #2206, opened 2024-03-14
Turned exercise 014 into practice

divyankachaudhari:
Addresses #2198. Turns [Exercise 14](https://github.com/divyankachaudhari/ocaml.org/blob/010/data/exercises/014_duplicate_elements.md) into Practice.

`dune build --root .` runs.

In the exercise, there's a remark `Remark: this function is not tail recursive. Can you modify it so it becomes so?`. What do I do of it?

@cuihtlauac Can you please review this? Thank you.


cuihtlauac (approved):

PR #2208, opened 2024-03-14
Turned exercise 016 into Practice.

divyankachaudhari:
Addresses #2198. Turns [Exercise 16](https://github.com/divyankachaudhari/ocaml.org/blob/010/data/exercises/016_drop_elements.md) into Practice.


practice/016/ex.ml
cuihtlauac, line 1: Add tests with drop at 0 and 1.

cuihtlauac (changes requested):

practice/016/ex.ml
divyankachaudhari: Done.

cuihtlauac (approved):

PR #2209, opened 2024-03-14
Turned exercise 017 into practice.

divyankachaudhari:
Addresses #2198. Turns [Exercise 017](https://github.com/divyankachaudhari/ocaml.org/blob/010/data/exercises/017_split_list.md) into Practice.


practice/017/ex.ml
cuihtlauac, line 1: Can you add a test a length equal to the list's length

cuihtlauac (changes requested):

practice/017/ex.ml
divyankachaudhari: Done.

cuihtlauac (approved):

PR #2210, opened 2024-03-14
Turned exercise 018 into practice

divyankachaudhari:
Addresses #2198. Turns [Exercise 018](https://github.com/divyankachaudhari/ocaml.org/blob/010/data/exercises/018_extract_slice.md) into practice.

Added 3 extra test cases.


practice/018/answer/impl.ml
cuihtlauac, line 1: Can you work out a solution using `List.filteri`? As an OCaml newbie, would you prefer the existing solution or the `List.filteri` one? Which one helps more in learning OCaml?

divyankachaudhari: Oh. Interesting. It would look like:

```
let slice list i k = List.filteri (fun index _ -> index >= i && index <= k) list

```
I think it really depends. As a newbie, I'd like simple solutions as they take less energy to understand. However, I think it really depends. 

- `fold_until` solution teaches tail recursion and pattern matching. 
- `List.filteri` solution is more simple and readable. 

Now that I'm exposed to this, I'd like to know about both methods 😅 I'm not leaning towards `List.filteri` because it doesn't seem like a clear winner to me because the goal of these exercises is to teach _functional programming_ in OCaml.

Is it possible that we could add the `List.filteri` solution in a comment line? See the commit below.

cuihtlauac: Thanks a lot, that's a very well thought answer.

Initially, I was going to tell you to dump the existing version and replace it by the `List.filteri` one. But you are right; the goal is to teach functional programming. In this perspective the `List.filteri` answer should be a comment, as you proposed.

But I still believe the existing solutions do more harm than good. It is too complex. I believe a nicer implementation without using an iterator is possible. Take a look at the way `List.filteri` is [implemented](https://github.com/ocaml/ocaml/blob/main/stdlib/list.ml#L269). Can you write something like that?

cuihtlauac (changes requested):
Can you create a folder `answer2` with the same structure as `answer` but containing the `List.filteri` example?

cuihtlauac:
@divyankachaudhari: Have you forgot about this one ;-)

practice/018/answer/impl.ml
```diff
@@ -0,0 +1,15 @@
...
+let slice list i k =
+  let _, list = fold_until (fun _ _ -> []) [] i list in
+  let taken, _ = fold_until (fun acc h -> h :: acc) [] (k - i + 1) list in
+  List.rev taken
+(* 
+An elegant way to do this would be to use List.filteri function:
+
+let slice list i k = 
+  List.filteri (fun index _ -> index >= i && index <= k) list
+*)
```
cuihtlauac:
```suggestion
let slice list i k = 
  List.filteri (fun index _ -> index >= i && index <= k) list
```

cuihtlauac (changes requested):

divyankachaudhari:
@cuihtlauac Hahah, I was deliberately procrastinating a bit. 

So, this is the `List.filteri` solution:
```
let slice list i k = List.filteri (fun index _ -> index >= i && index <= k) list
```

This is a solution inspired by `List.filteri`'s implementation without relying on iterators:
```
let slice list start stop =
  let rec aux i = function
    | [] -> []
    | x :: xs when i >= start && i <= stop -> x :: (aux (i + 1) xs)
    | _ :: xs -> aux (i + 1) xs
  in aux 0 list
```
I wasn't sure which one to retain in the `answer` folder, so I've left it unchanged (except for making the variable names a bit better/readable) and added this new solution in `answer2` folder.

------

I was just writing the comment above when I saw the suggested changes.

cuihtlauac (approved):
Thanks @divyankachaudhari. I didn't liked the original solution, I picked yours and mine only

PR #2212, opened 2024-03-14
Turn exercise 023 to practice

Ozyugoo:


Ozyugoo:
Hi, @cuihtlauac Kindly check this. I had to put include a zero so that it will pass the nil test case.

practice/023/ex.ml
cuihtlauac, line 1: We need a more clever way to test this.
- The length of the list returned should always be equal to the number passed to rand_select
- Each element of the returned list appears in the argument list (use `List.mem`)

cuihtlauac (changes requested):

practice/023/answer/impl.ml
```diff
@@ -0,0 +1,16 @@
+let rand_select list n =
+  Random.init 0;
```
cuihtlauac, line 2:
```suggestion
```

We can have the real thing and do the tests.

cuihtlauac (changes requested):

PR #2213, opened 2024-03-14
Turned exercise 019 into practice

divyankachaudhari:
Addresses #2198. Turns [Exercise 019](https://github.com/ocaml/ocaml.org/blob/010/data/exercises/019_rotate_list.md) into practice.

Added four extra test cases:
- Rotate with negative n
- Rotate more than the list length
- Rotate zero places
- Rotate Empty List


practice/019/ex.ml
cuihtlauac, line 1: The test with negative length fails when running `dune test --root .`

cuihtlauac (changes requested):

practice/019/ex.ml
divyankachaudhari: Fixed.

divyankachaudhari:
@cuihtlauac Could you please review this PR? Thank you.

cuihtlauac (approved):
Thnaks @divyankachaudhari

PR #2216, opened 2024-03-14
Turned exercise 022 into practice

divyankachaudhari:
Addresses #2198. Turns [Exercise 022](https://github.com/ocaml/ocaml.org/blob/main/data/exercises/022_create_list.md) into practice.

Added two extra test cases:
- Descending order
- Same number

The solution in the markdown has two solutions: a non-tail recursive and a tail recursive. So, with #2210 as precedent, I added one solution normally and I added the other one in comments.


practice/022/answer/impl.ml
cuihtlauac, line 1: Turn the tail recursive implementation into an `answer2` folder.

cuihtlauac (changes requested):

practice/022/answer/impl.ml
divyankachaudhari: Done.

divyankachaudhari:
Added two solutions, one in `answer` and tail-recursive one in `answer2`.

Both `dune test --root . answer` and `dune test --root . answer2` work.

divyankachaudhari:
@cuihtlauac Could you please review this again? Thank you.

cuihtlauac (approved):
Thanks @divyankachaudhari, sorry for the delay in reviewing

cuihtlauac (approved):

Issue #2200, opened 2024-03-14
Turn exercices 033 to 042 into practice


cuihtlauac:
This task consists of doing 9 exercises instead of one, as described in issues #2167, #2168, #2170, #2172, #2173, #2176, #2177, #2178, #2179 and #2180. Create 10 folders (from 033 to 042), and populate them as the others. Makes sure `dune build --root .` and `dune test --root . answer` work. 

- [X] 033 #2220 
- [X] 034 #2221 
- [X] 035 #2222 
- [X] 036 #2223 
- [X] 037 #2245
- [X] 038 #2248 
- [X] 039 #2281
- [X] 040 #2283
- [ ] 041

MissJae:
Can I do this? @cuihtlauac

maha-sachin:
@cuihtlauac, can you help me to assign this issue?

cuihtlauac:
@divyankachaudhari

cuihtlauac:
@MissJae 
> Can I do this? @cuihtlauac

Complete #2192 first

komutumu:
@cuihtlauac perhaps I can work on some of the issues here that haven't been worked on yet here?

PR #2220, opened 2024-03-15
Turned exercise 033 into practice

divyankachaudhari:
Addresses #2200. Turns [Exercise 033](https://github.com/ocaml/ocaml.org/blob/main/data/exercises/033_is_coprime.md) into practice.


cuihtlauac (approved):
Thanks @divyankachaudhari, this is nice

PR #2221, opened 2024-03-15
Turned exercise 034 into practice

divyankachaudhari:
Addresses #2200. Turns [Exercise 034](https://github.com/ocaml/ocaml.org/blob/main/data/exercises/034_euler_totient.md) into practice.

Added two extra test cases:
- φ(1) = 1 (Edged test case)
- φ(13) = 12 (Prime number)


cuihtlauac (approved):
Nice @divyankachaudhari. Thanks for adding the test cases

PR #2222, opened 2024-03-15
Turned exercise 035 into practice

divyankachaudhari:
Addresses #2200. Turns [Exercise 035](https://github.com/ocaml/ocaml.org/blob/main/data/exercises/035_prime_factor.md) into practice.

Added two extra test cases:
- Prime factors of 1
- Prime factor of a prime number


cuihtlauac (approved):
Thanks @divyankachaudhari

PR #2223, opened 2024-03-15
Turned exercise 036 into practice

divyankachaudhari:
Addresses #2200. Turns [Exercise 036](https://github.com/ocaml/ocaml.org/blob/main/data/exercises/036_prime_factor_2.md) into practice.

Added two extra test cases:
- Prime factors with multiplicities of 1
- Prime factors with multiplicities of a prime number


cuihtlauac (approved):
Tanks @divyankachaudhari

PR #2226, opened 2024-03-15
Turn exercise 032 to practice

Ozyugoo:


cuihtlauac (changes requested):
Thanks @Ozyugoo. Please remove files related to exercises 023 from this PR

PR #2227, opened 2024-03-16
Sort exercises by slug before emitting template

### LINKED ISSUE #2149: Exercises are not listed in order
Hey all, I recently noticed that the problems in the [Exercises](https://ocaml.org/exercises) page are not being displayed in "ascending" order, i.e. following the slug numbering [here](https://github.com/ocaml/ocaml.org/tree/main/data/exercises). Here's what the first few items currently look like for me;  I've tried both Chrome and Firefox in a couple devices just to be sure, but let me know if you can't reproduce it.

<details>
  <summary>~Large picture.</summary>
  
  
</details>

I've also observed the same partial (unsorted) order when filtering exercises by difficulty.
I'm under the impression that they used to be in ascending order before. I'm not sure if that's the intended behavior, but I think the ascending order gives a nice progression to the problems, and it can be confusing for beginners to see some less trivial exercises right away within the first 4-5 in the list.

I'm pretty new to OCaml and not super familiar with Alpine or Dream+eml templating, but I tried to take a look anyways. Do feel free to ignore the following if it's too off-base.

As far as I can tell, [the Dream handler for the `exercises` route](https://github.com/ocaml/ocaml.org/blob/29b37c3e28ad1f09120a2b0de30916aa0f170f32/src/ocamlorg_web/lib/handler.ml#L458) gets the exercise list from the `Data.Exercise.all` value, which comes from calling [this `all` function](https://github.com/ocaml/ocaml.org/blob/29b37c3e28ad1f09120a2b0de30916aa0f170f32/tool/ood-gen/lib/exercise.ml#L76), which in turn uses some helpers from [this `Utils` module](https://github.com/ocaml/ocaml.org/blob/main/tool/ood-gen/lib/utils.ml) to read and decode the markdown files. In particular, I'd guess that [this](https://github.com/ocaml/ocaml.org/blob/29b37c3e28ad1f09120a2b0de30916aa0f170f32/tool/ood-gen/lib/utils.ml#L38) `Bos.OS.Path.matches` call is where the actual exercise order first materializes, since the rest looks like (presumably) order-preserving maps.

If that call cannot guarantee the order in which the markdown files/filepaths are enlisted (at least the [docs](https://erratique.ch/software/bos/doc/Bos/OS/Path/index.html#val-matches) don't seem to claim so), then would it be too unwieldy to sort the exercises explicitly inside the route handler? After all, the exercise list is filtered by difficulty [inside the handler](https://github.com/ocaml/ocaml.org/blob/29b37c3e28ad1f09120a2b0de30916aa0f170f32/src/ocamlorg_web/lib/handler.ml#L472) as well, right before passing the list to the template's main render function (`Ocamlorg_frontend.exercises`). Though, I imagine it would be ideal (if possible) for this sorting to happen at build time/statically instead of every time a request is handled. Not sure what the best way to do that would be. At least it should be a quite small sort.

csaltachin:
Should fix #2149.

Note that comparing slugs lexicographically doesn't work because it yields `"10" < "2"`, etc. Here we parse each slug into its numeric part (defaulting to 0) and its rightmost letter (if present) before comparing them as `int * string` tuples.
Maybe the letter could be instead parsed as a `char option` or directly as an `int`, but I hope it's not too convoluted overall.

### DIFF
--- a/tool/ood-gen/lib/exercise.ml
+++ b/tool/ood-gen/lib/exercise.ml
@@ -74,16 +74,8 @@ let decode (fpath, (head, body)) : (t, [> `Msg of string ]) result =
   Ok (metadata |> of_metadata ~statement ~solution)
 
 let compare_by_slug =
-  let parse_int s = s |> int_of_string_opt |> Option.value ~default:0 in
   let key exercise : int * string =
-    let slug = exercise.slug in
-    let len = String.length slug in
-    if len = 0 then (0, "")
-    else
-      match slug.[len - 1] with
-      | 'A' .. 'Z' as c ->
-          (parse_int (String.sub slug 0 (len - 1)), String.make 1 c)
-      | _ -> (parse_int slug, "")
+    Scanf.sscanf exercise.slug "%d%[A-Z]" (fun s c -> (s, c))
   in
   fun (x : t) (y : t) -> compare (key x) (key y)
 
--- a/tool/ood-gen/lib/exercise.ml
+++ b/tool/ood-gen/lib/exercise.ml
@@ -73,7 +73,22 @@ let decode (fpath, (head, body)) : (t, [> `Msg of string ]) result =
   in
   Ok (metadata |> of_metadata ~statement ~solution)
 
-let all () = Utils.map_files decode "exercises/*.md"
+let compare_by_slug =
+  let parse_int s = s |> int_of_string_opt |> Option.value ~default:0 in
+  let key exercise : int * string =
+    let slug = exercise.slug in
+    let len = String.length slug in
+    if len = 0 then (0, "")
+    else
+      match slug.[len - 1] with
+      | 'A' .. 'Z' as c ->
+          (parse_int (String.sub slug 0 (len - 1)), String.make 1 c)
+      | _ -> (parse_int slug, "")
+  in
+  fun (x : t) (y : t) -> compare (key x) (key y)
+
+let all () =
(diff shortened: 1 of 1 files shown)


tool/ood-gen/lib/exercise.ml
```diff
@@ -73,7 +73,22 @@ let decode (fpath, (head, body)) : (t, [> `Msg of string ]) result =
...
+  let parse_int s = s |> int_of_string_opt |> Option.value ~default:0 in
+  let key exercise : int * string =
+    let slug = exercise.slug in
+    let len = String.length slug in
+    if len = 0 then (0, "")
+    else
+      match slug.[len - 1] with
+      | 'A' .. 'Z' as c ->
+          (parse_int (String.sub slug 0 (len - 1)), String.make 1 c)
+      | _ -> (parse_int slug, "")
```
cuihtlauac:
```suggestion
    Scanf.sscanf exercise.slug "%d%[A-Z]" (fun s c -> (s, c))
```

What do you think of something like that?

csaltachin: Wow, I didn't know sscanf was so versatile -- thanks for showing me. I'll change this and remove parse_int.
This could throw at build time if sscanf fails, but I don't imagine the slug format will change drastically in the future. We can always just update the format string + receiver function if so.

cuihtlauac (approved):
Thanks @csaltachin, this looks good to me

PR #2231, opened 2024-03-17
Turn exercise 024 to practice

Ozyugoo:


Ozyugoo:
Hi, @cuihtlauac kindly check it out.

practice/024/answer/impl.ml
```diff
@@ -0,0 +1,25 @@
...
+
+let range a b =
+  let rec aux acc high low =
+    if high >= low then
+      aux (high :: acc) (high - 1) low
+    else acc
+  in
+    if a < b then aux [] b a else List.rev (aux [] a b)
+
+  let lotto_select n m = rand_select (range 1 m) n
```
cuihtlauac:
```suggestion
  let lotto_select n m = rand_select (range 1 m) n

```

cuihtlauac (changes requested):

practice/024/answer/impl.ml
```diff
@@ -0,0 +1,26 @@
...
+
```
cuihtlauac: You need to fix this end of file

cuihtlauac (changes requested):

practice/024/answer/impl.ml
Ozyugoo: Hey @cuihtlauac sorry about this end of file, I will be more careful. I was viewing with the wrong editor as you said.

cuihtlauac (approved):
Thanks @Ozyugoo, this looks good to me

PR #2236, opened 2024-03-17
Turn exercise 031 to practice

Ozyugoo:


Ozyugoo:
@cuihtlauac kindly check this out, also I didn't see exercises 029 and 030.

practice/031/ex.ml
```diff
@@ -0,0 +1,15 @@
+open OUnit2
+
+module type Testable = sig
+  val is_prime : int -> bool
+end
+
+module Make(Tested: Testable) : sig val v : test end = struct
+  let v = "is_prime" >::: [
+    "nil" >:: (fun _ -> assert_equal false (Tested.(not (is_prime 7))));
+    "cons" >:: (fun _ -> assert_equal true (Tested.is_prime 7));
```
cuihtlauac: * Remove test `nil`.
* Define a list of primes such as: 2, 3, 5, 7, 11, 13, 17, 19, 23, 29, 31, 37, 41, 43, 47, 53, 59, 61, 67, 71, 73, 79, 83, 89, 97
* Compute the list of tests for all values from 0 to 100 from the list above

cuihtlauac (changes requested):

practice/031/ex.ml
Ozyugoo: Kindly check if what I have done if what expect me to do, if not, please point me in the right direction, thanks.

practice/031/answer/impl.ml
```diff
@@ -0,0 +1,7 @@
+let is_prime n =
+  let n = abs n in
+  let rec is_not_divisor d =
+    d * d > n || (n mod d <> 0 && is_not_divisor (d + 1)) in
+  n <> 1 && is_not_divisor 2
+
+  let prime_0_to_100 = [ 2; 3; 5; 7; 11; 13; 17; 19; 23; 29; 31; 37; 41; 43; 47; 53; 59; 61; 67; 71; 73; 79; 83; 89; 97]
```
cuihtlauac:
```suggestion
  let prime_0_to_100 = [2; 3; 5; 7; 11; 13; 17; 19; 23; 29; 31; 37; 41; 43; 47; 53; 59; 61; 67; 71; 73; 79; 83; 89; 97]

```

practice/031/ex.ml
cuihtlauac, line 1: You should write a recursive function `flag_prime` which has type `int -> int list -> (int * bool) list`

Writing `flag_prime 0 [2;3;5;7;11]` returns `[(0, false); (1, false); (2, true); (3, true); (4, false); (5, true); (6; false); (7, true); (8, false); (9, false); (10, falsse); (11, true)]`

Using this function, you can don't have to enumerate all the cases when write `let v = ` You can just call the function

cuihtlauac (changes requested):

practice/031/ex.ml
Ozyugoo: I wonder if this is what you expect.

practice/031/answer/impl.ml
```diff
@@ -0,0 +1,13 @@
...
+    d * d > n || (n mod d <> 0 && is_not_divisor (d + 1)) in
+  n <> 1 && is_not_divisor 2
+
+let rec flag_prime n lst_n = 
+  match lst_n with 
+  | [] -> []
+  | hd :: tl -> (hd, is_prime hd) :: flag_prime n tl
+
+
+  let prime_0_to_100 = [2; 3; 5; 7; 11; 13; 17; 19; 23; 29; 31; 37; 41; 43; 47; 53; 59; 61; 67; 71; 73; 79; 83; 89; 97]
```
cuihtlauac:
```suggestion
  let prime_0_to_100 = [2; 3; 5; 7; 11; 13; 17; 19; 23; 29; 31; 37; 41; 43; 47; 53; 59; 61; 67; 71; 73; 79; 83; 89; 97]

```

cuihtlauac: Function `flag_prime` and list `prime_0_to_100` (that should be renamed `primes`) should be defined in file `ex.ml`, not in `impl.ml` they should be used to test `is_prime`. Also, because of `flag_prime` should be used to test `is_prime`, it must not use it.

`flag_prime` is basically iterating from 0 to the latest value in `primes`. Each value is compared to the head of the primes list
- if equal then, the result pair is marked `true` and the head of `primes` is removed
- if no equal, the result pair is marked `false` an computation continues with `primes` unchanged

practice/031/ex.ml
cuihtlauac: One you have a valid `flag_prime` version you'll be able to have something like that:
```ocaml
"flag_prime" >::: List.map (fun (n, is_prime) -> fun _ assert_equal is_prime (Tested.is_prime n))
```

This way, the implementation of `is_prime` will be tested for all the values from `0` to `97`, with a single line of code :-)

cuihtlauac (changes requested):
Thanks @Ozyugoo. This is a hard case, take the time to have a look at the ocaml documentation, the Lists tutorial may be helpful

Ozyugoo:
Hi @cuihtlauac, please take a look at this, I don't know if this is what you expect me to do.

practice/031/ex.ml
cuihtlauac: This is better, but it does not compile because `primes` has type `int list` and the anonymous function you pass to `List.map` has type `(int * bool) list`. However, this is indeed what I want.

You need to build the `(int * bool) list` from the `primes` using this function:
```ocaml
  let rec flag_primes n = function
  | [] -> []
  | p :: u -> if n = p then
                (n, true) :: flag_primes (n + 1) u
              else
                (n, false) :: flag_primes (n + 1) (p :: u)
```

practice/031/answer/impl.ml
```diff
@@ -0,0 +1,6 @@
+let is_prime n =
+  let n = abs n in
+  let rec is_not_divisor d =
+    d * d > n || (n mod d <> 0 && is_not_divisor (d + 1)) in
+  n <> 1 && is_not_divisor 2
+
```
cuihtlauac:
```suggestion
  n <> 1 && is_not_divisor 2

```

cuihtlauac (changes requested):

practice/031/ex.ml
Ozyugoo: Hi @cuihtlauac thank you for your feedback. This is what I have been able to come up with, is this what you expect?

practice/031/ex.ml
cuihtlauac: Make sure `dune build --root .` is successful

practice/031/ex.ml
```diff
@@ -0,0 +1,33 @@
...
+  | [] -> []
+  | p :: u ->
+    if n = p then 
+      (p, Tested.is_prime p) :: flag_primes (n + 1) u
+    else
+      (n, false) :: flag_primes (n + 1) (p :: u)
+
+  let primes = [2; 3; 5; 7; 11; 13; 17; 19; 23; 29; 31; 37; 41; 43; 47; 53; 59; 61; 67; 71; 73; 79; 83; 89; 97]
+
+  let prime_flags = flag_primes 2 primes
```
cuihtlauac:
```suggestion
  let prime_flags = flag_primes 0 primes
```

practice/031/ex.ml
```diff
@@ -0,0 +1,33 @@
...
+
+module type Testable = sig
+  val is_prime : int -> bool
+end
+
+module Make(Tested: Testable) : sig val v : test end = struct
+  let rec flag_primes n = function
+  | [] -> []
+  | p :: u ->
+    if n = p then
```
cuihtlauac:
```suggestion
    if n = p then
```

practice/031/ex.ml
```diff
@@ -0,0 +1,33 @@
...
+module type Testable = sig
+  val is_prime : int -> bool
+end
+
+module Make(Tested: Testable) : sig val v : test end = struct
+  let rec flag_primes n = function
+  | [] -> []
+  | p :: u ->
+    if n = p then 
+      (p, Tested.is_prime p) :: flag_primes (n + 1) u
```
cuihtlauac:
```suggestion
      (p, true) :: flag_primes (n + 1) u
```

Since `flag_prime` is used to test `is_prime` you can't use it.

cuihtlauac (changes requested):

practice/031/ex.ml
Ozyugoo: Thank you @cuihtlauac for your feedback and guidance.

cuihtlauac (changes requested):
`dune build --root .` fails :-(

Ozyugoo:
Hey @cuihtlauac I think I nailed it this time, kindly look at it, thank you.

practice/031/ex.ml
```diff
@@ -0,0 +1,33 @@
...
+    let is_prime = Tested.is_prime n in
+    let test_name = Printf.sprintf "Testing if %d is prime" n in
+    test_name >:: (fun _ -> assert_equal is_prime (Tested.is_prime n))
+  ) primes
+
+  let v = "is_prime Tests" >::: [
+    tests
+  ]
+
+  let _ = flag_prime 0 primes
```
cuihtlauac:
```suggestion
```

practice/031/answer/impl.ml
```diff
@@ -0,0 +1,5 @@
+let is_prime n =
+  let n = abs n in
+  let rec is_not_divisor d =
+    d * d > n || (n mod d <> 0 && is_not_divisor (d + 1)) in
+  n <> 1 && is_not_divisor 2
```
cuihtlauac:
```suggestion
  n > 1 && is_not_divisor 2
```
This was bugged. Can you fix in markdown?

practice/031/ex.ml
```diff
@@ -0,0 +1,33 @@
...
+      if n = p then
+        (p, true) :: flag_prime (n + 1) u
+      else
+        (n, false) :: flag_prime (n + 1) (p :: u)
+
+  let primes = [2; 3; 5; 7; 11; 13; 17; 19; 23; 29; 31; 37; 41; 43; 47; 53; 59; 61; 67; 71; 73; 79; 83; 89; 97]
+
+  let tests =
+  "flag_prime" >::: List.map (fun n ->
+    let is_prime = Tested.is_prime n in
```
cuihtlauac:
```suggestion
  "flag_prime" >::: List.map (fun (n, p) ->
```

practice/031/ex.ml
```diff
@@ -0,0 +1,33 @@
...
+    let test_name = Printf.sprintf "Testing if %d is prime" n in
+    test_name >:: (fun _ -> assert_equal is_prime (Tested.is_prime n))
+  ) primes
```
cuihtlauac:
```suggestion
    test_name >:: (fun _ -> assert_equal p (Tested.is_prime n))
  ) (flag_prime 0 primes)
```

cuihtlauac (changes requested):
Here is how it was meant to be

Ozyugoo:
> Here is how it was meant to be

Thank you for this, it was quite challenging to crack.

data/exercises/031_is_prime.md
```diff
@@ -2,23 +2,23 @@
...
 ---
 
 # Solution
 
-  Recall that `d` divides `n` iff `n mod d = 0`.  This is a naive
-  solution.  See the [Sieve of
-  Eratosthenes](http://en.wikipedia.org/wiki/Sieve_of_Eratosthenes) for a
-  more clever one.
-  
+Recall that `d` divides `n` iff `n mod d = 0`. This is a naive
```
christinerose: Should "iff" be "if" or does it stand for "if and only if"? If the latter, it should be IFF in all caps to as no to cause confusion.

christinerose (commented):
Just the one question about if/iff/IFF. After that, I'll have a look at the entire thing again.

data/exercises/031_is_prime.md
Ozyugoo: > Should "iff" be "if" or does it stand for "if and only if"? If the latter, it should be IFF in all caps to as no to cause confusion.

I think it be should IFF since it is explaining that the condition for d to divide n is that the remainder of n divided by d is zero. Let me change it in caps then.

data/exercises/031_is_prime.md
```diff
@@ -8,7 +8,7 @@ description: "Check if a given integer is a prime number."
 
 # Solution
 
-Recall that `d` divides `n` iff `n mod d = 0`. This is a naive
+Recall that `d` divides `n` IFF `n mod d = 0`. This is a naive
```
cuihtlauac:
```suggestion
Recall that `d` divides `n` if and only if `n mod d = 0`. This is a naive
```

cuihtlauac (changes requested):

cuihtlauac (approved):

cuihtlauac:
Thanks @Ozyugoo, this was not an easy one

Ozyugoo:
> Thanks @Ozyugoo, this was not an easy one

Thank you @cuihtlauac it means a lot coming from you.

PR #2245, opened 2024-03-18
Turned exercise 037 into practice

divyankachaudhari:
Addresses #2200. Turns [Exercise 037](https://github.com/ocaml/ocaml.org/blob/main/data/exercises/037_euler_totent_improved.md) into practice.

Added one extra test case:
- φ(1) = 1 (Edge test case)


cuihtlauac (approved):
Thanks @divyankachaudhari

PR #2247, opened 2024-03-18
(docs) fix example in 'Libraries With Dune'

0xRamsi:
Just a small fix, so the example actually compiles.


christinerose (approved):

sabine (approved):
Thanks @0xRamsi ! :sparkles:

PR #2248, opened 2024-03-18
Merge 038 into 037

divyankachaudhari:
Addresses #2278. Merges [Exercise 038](https://github.com/ocaml/ocaml.org/blob/main/data/exercises/038_compare_totient.md) into [Exercise 037](https://github.com/ocaml/ocaml.org/blob/main/data/exercises/037_euler_totent_improved.md).


practice/038/answer/impl.ml
```diff
@@ -0,0 +1,31 @@
...
+  let rec aux acc = function
+    | [] -> acc
+    | (p, m) :: t -> aux ((p - 1) * pow p (m - 1) * acc) t
+  in aux 1 (factors n)
+
+let timeit f a =
+  let t0 = Unix.gettimeofday () in
+  ignore (f a);
+  let t1 = Unix.gettimeofday () in
+  t1 -. t0
```
cuihtlauac: This should be part of file `ex.ml` not this one, otherwise we ask the system under test to provide its one time measuring procedure, which allows cheating ;-)

practice/038/ex.ml
```diff
@@ -0,0 +1,27 @@
+open OUnit2
+
+module type Testable = sig
+  val phi : int -> int
+  val phi_improved : int -> int
+  val timeit : (int -> int) -> int -> float
```
cuihtlauac:
```suggestion
```

practice/038/work/impl.ml
```diff
@@ -0,0 +1,6 @@
+let rec gcd a b = if b = 0 then a else gcd b (a mod b)
+let coprime a b = gcd a b = 1
+let phi _ = failwith "Not yet implemented"
+let factors _ = failwith "Not yet implemented"
+let phi_improved _ = failwith "Not yet implemented"
+let timeit _ _ = failwith "Not yet implemented"
```
cuihtlauac:
```suggestion
```

cuihtlauac (changes requested):
I'm thinking we should merge this exercice with the 37. The goal of 037 is to improve the performance of 34. The tests for 37 should be : “the results are the same, but always faster that 34”

practice/038/answer/impl.ml
divyankachaudhari: I understand what you're saying: testing framework should be independent of the system under test. 

But, the [`solution`](https://github.com/ocaml/ocaml.org/blob/main/data/exercises/038_compare_totient.md#solution) is the `timeit` function. That's assuming they're _already_ given `phi` and `phi_improved` (which is not the case in the current commit) and they have to just implement `timeit` since they've already implemented `phi` and `phi_improved` before.

So should we
- **case 1:** put `timeit` in `ex.ml` and then just ask the user to implement `phi` and `phi_improved` again and then just run the tests?
- or, **case 2:** implement `phi` and `phi_improved` in `work/impl.ml` (and put them in `ex.ml` just in case the learner modifies their copy of `phi` and `phi_improved` in `work/impl.ml` ) and ask the learner to just implement `timeit`?

I think the current question is expecting case 2.

cuihtlauac: These exercises are derived from a Lisp implementation but were initially designed for Prolog. I believe exercise 38 is just not a good idea. It should be part of 37 to check the results are the same as in 34 and the performance is better. Asking the users to write the `timeit` function does not have a great value

divyankachaudhari: @cuihtlauac That's true. So should we not do anything for `038`? It's trying to demonstrate that `phi_improved` is faster and not just that it gives the same answer.

cuihtlauac: Yes, 38 should be deleted. I've created this: https://github.com/ocaml/ocaml.org/issues/2278

cuihtlauac: You can change the title of this issue and do the changes on 37 in this PR. Fixing the markdown in `/data` should be another PR

divyankachaudhari: @cuihtlauac Okay, on it 🫡

cuihtlauac (changes requested):
Let's have this subsumed by 37

practice/037/ex.ml
```diff
@@ -1,17 +1,34 @@
 open OUnit2
 
 module type Testable = sig
+  val phi : int -> int
   val phi_improved : int -> int
```
cuihtlauac:
```suggestion
```

The API should be the same as 34, it's only an optimization

practice/037/ex.ml
cuihtlauac, line 1: The naive implementation of phi (from `34/answer.ml`) should go into `ex.ml`

The tests should be a dynamically generated list results from `Tested.phi` and `naive_phi` are compared : result same, speed improved. We only need a couple of values, such as 10090, 1090, 190, 19 (only examples)

So the performance improvement is a comment (you probably don't need the printfs), is a fail it solution for 37 does no go faster than 34

practice/037/work/impl.ml
cuihtlauac, line 1: The goal is from 37, so the task is not different. No changes are needed in `37/work/impl.ml`

cuihtlauac (changes requested):

practice/037/ex.ml
divyankachaudhari: It may or may not be faster for small values of `n` because it depends on the execution environment, should we still fail the tests if `037` isn't faster than `034`? It will be definitely faster for large values of `n` such as 10090.

Right now, I'm asserting that output of `034` == output of `037` but only printing the time taken to run as a comment (and not using the timing to assert anything). 

```
~/ocaml.org/practice/037 (038 ✘)✭ ᐅ dune test --root . answer              
.For n=10090, phi_naive took 0.001097 seconds, phi took 0.000004 seconds
For n=1090, phi_naive took 0.000080 seconds, phi took 0.000000 seconds
For n=190, phi_naive took 0.000007 seconds, phi took 0.000000 seconds
For n=19, phi_naive took 0.000001 seconds, phi took 0.000000 seconds

Ran: 1 tests in: 0.11 seconds.
OK
```


If I modify it to fail if `phi` (037) is not faster than `phi_naive` (034). It fails on small value, `n=19`:

```
~/ocaml.org/practice/037 (038 ✘)✹✭ ᐅ dune test --root . answer
File "answer/test/dune", line 2, characters 12-15:
2 | (test (name run) (libraries ounit2 ex))
                ^^^
F
==============================================================================
Error: Euler's Totient Function Performance Tests:0:compare phi_naive and phi.

File "/Users/divyanka/ocaml.org/practice/037/_build/default/answer/test/oUnit-Euler's Totient Function Performance Tests-divyanka.local#01.log", line 2, characters 1-1:
Error: Euler's Totient Function Performance Tests:0:compare phi_naive and phi (in the log).

Raised at OUnitAssert.assert_failure in file "src/lib/ounit2/advanced/oUnitAssert.ml", line 45, characters 2-27
Called from Stdlib__List.iter in file "list.ml", line 110, characters 12-15
Called from OUnitRunner.run_one_test.(fun) in file "src/lib/ounit2/advanced/oUnitRunner.ml", line 83, characters 13-26

For n=19, phi should be faster than phi_naive. phi_naive took 0.000000 seconds, phi took 0.000000 seconds
------------------------------------------------------------------------------
Ran: 1 tests in: 0.11 seconds.
FAILED: Cases: 1 Tried: 1 Errors: 0 Failures: 1 Skip:  0 Todo: 0 Timeouts: 0.
```

cuihtlauac: You are right, testing on small values does not make sense. I overlooked this. Thanks for pointing this.

We should test we three, four or a couple more values, starting in the thousands and growing exponentially.

Regarding the tests, I was thinking using `assert_bool true` and have a function which calls `phi`, `phi_naive` and returns `true` if results are the same and `phi` improved is indeed faster than `phi_naive`

cuihtlauac (changes requested):

practice/037/ex.ml
```diff
@@ -1,23 +1,48 @@
...
+        Printf.printf "For n=%d, phi_naive took %f seconds, phi took %f seconds\n" n time_phi time_phi_improved;
+        let msg = Printf.sprintf "For n=%d, phi should be faster than phi_naive. phi_naive took %f seconds, phi took %f seconds" n time_phi time_phi_improved in
+        assert_bool msg (time_phi_improved < time_phi)
+      ) test_values
+  
+    let v = "Euler's Totient Function Performance Tests" >::: [
+      "compare phi_naive and phi" >:: (fun _ -> compare_phi_functions ());
+    ]
+  end
+
```
cuihtlauac:
```suggestion
    let test_values = [10090; 100090; 1000090]

    let compare_phi_functions () =
      List.iter (fun n ->
        let result_phi = phi_naive n in
        let result_phi_improved = Tested.phi n in
        let time_phi = timeit phi_naive n in
        let time_phi_improved = timeit Tested.phi n in

        (* Ensuring the results are the same *)
        assert_equal ~msg:(Printf.sprintf "phi_naive and phi should return the same result for n=%d" n) result_phi result_phi_improved;

        (* Failing the test if phi_improved is not faster than phi_naive *)
        Printf.printf "For n=%d, phi_naive took %f seconds, phi took %f seconds\n" n time_phi time_phi_improved;
        let msg = Printf.sprintf "For n=%d, phi should be faster than phi_naive. phi_naive took %f seconds, phi took %f seconds" n time_phi time_phi_improved in
        assert_bool msg (time_phi_improved < time_phi)
      ) test_values

    let v = "Euler's Totient Function Performance Tests" >::: [
      "compare phi_naive and phi" >:: (fun _ -> compare_phi_functions ());
    ]
  end

```

cuihtlauac (changes requested):

cuihtlauac (approved):
Thanks @divyankachaudhari, this is really nice. Sorry for the delayed review.

Issue #1962, opened 2024-01-17
Using Dream as demonstrated in the tutorial is failing on Windows


heathhenley:
Trying to follow along with https://ocaml.org/docs/your-first-program#installing-and-using-modules-from-a-package - when I got to the install / run dream part, I ran into some problems on windows. 

Here's the versions: 
```
C:\dev\ocaml\hello2>opam --version
2.2.0~alpha0~20221228
```
and OCaml 4.14.1 on Windows 11 with dune 3.6.2. 

I'm getting an issue with pkg-config if I try to install dream with opam: 

```cmd
#=== ERROR while compiling conf-pkg-config.3 ==================================#
"pkg-config": command not found.


<><> Error report <><><><><><><><><><><><><><><><><><><><><><><><><><><><><><><>
┌─ The following actions failed
│ ▼ fetch camlp-streams                5.0.1
│ ▼ fetch caqti.1.9.0, caqti-lwt.1.9.0
│ ▼ fetch cppo                         1.6.9
│ λ build conf-pkg-config              3
└─
```

I believe the same issue is reported on the Dream repo here: https://github.com/aantron/dream/issues/313 and also on opam:
https://github.com/ocaml/opam-repository/issues/24851

I'm just going to switch to using WSL to work through the tutorials if this doesn't turn out to be an issue between the keyboard and chair. If it is really just a Windows support thing, maybe a note should be added where the tutorials use Dream so that others newbies don't hit this snag?  It's a bummer to run into this while trying to learn OCaml, and I would be happy to PR it if it's something you think is useful. 

Confirmed the hello world dream tutorials example is working using WSL on Windows instead.

cuihtlauac:
Hi @heathhenley, sorry to hear about the bad experience in Windows. Honestly, this was not tested under Windows; you're the first to try running this in Windows. Thanks for letting us know about this issue.

Yes, a disclaimer may help. If the second part of this tutorial works in Windows, the others will too. "Your first OCaml program" is the only tutorial that requires installing Opam packages; the other only needs vanilla OCaml.

sabine:
We have to change to a different code example anyways, since the "Getting Started" and "Language" sections are not meant to use packages from the OCaml ecosyste. That should reduce the dependencies and risk for such breakage a lot.

heathhenley:
[...]

PR #2249, opened 2024-03-18
(docs) Use DkML 2.1.0

jonahbeckford:
- Correct DkML capitalization
- Remove outdated name "Diskuv OCaml" and replace with DkML
- Update WSL2 with 4.14.2
- Replace usages of DkML installer with DkML distribution since winget is now the controlling installer on Windows
- Remove `opam env` variations for DkML
- Restore `ppx_show` example for DkML since it works (obviously without the earlier Dream example)

### DIFF
--- a/data/tutorials/getting-started/1_00_install_OCaml.md
+++ b/data/tutorials/getting-started/1_00_install_OCaml.md
@@ -99,7 +99,7 @@ winget install Git.Git
 winget install Diskuv.OCaml
 ```
 
-And then in a **NEW** terminal:
+And then in a new terminal:
 
 ```powershell
 dkml init --system
--- a/data/tutorials/getting-started/1_00_install_OCaml.md
+++ b/data/tutorials/getting-started/1_00_install_OCaml.md
@@ -9,7 +9,7 @@ category: "First Steps"
 
 This guide will walk you through a minimum installation of OCaml. That includes installing a package manager and [the compiler](#installation-on-unix-and-macos) itself. We'll also install some platform tools like a build system, support for your editor, and a few other important ones.
 
-On this page, you'll find installation instructions for Linux, macOS, and &ast;BSD for recent OCaml versions. For Docker, Linux instructions apply, except when setting up opam. For Windows, we recommend using WSL but also provide instructions for installing OCaml 4.14.0 via the [Diskuv OCaml](https://github.com/diskuv/dkml-installer-ocaml#readme) Installer. If you are setting up OCaml on Windows and are unsure which installation method to use, you might be interested in reading [OCaml on Windows](/docs/ocaml-on-windows) first.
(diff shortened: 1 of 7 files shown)


christinerose (approved):

data/tutorials/getting-started/1_02_your_first_ocaml_program.md
cuihtlauac, line 1: This is not clear to me.
- Are the examples working with DkML 2.1.0?
- Are the examples working with WSL?
- Are the examples working with Cygwin?

We should be explicit about what is working or not.

data/tutorials/getting-started/1_00_install_OCaml.md
```diff
@@ -83,38 +83,30 @@ Opam is now installed and configured! You can now move to [installing the OCaml
...
-  -         >>  A transient failure occurred. <<<
-  ```
+```powershell
+winget install Microsoft.VisualStudio.2019.BuildTools --override "--wait --passive --installPath C:\VS --addProductLang En-us --add Microsoft.VisualStudio.Workload.VCTools --includeRecommended"
+winget install Git.Git
+winget install Diskuv.OCaml
+```
 
-* You may be asked to accept a certificate from `Open Source Developer, Gerardo Grignoli` for the `gsudo` executable that was issued by `Certum Code Signing CA SHA2`.
+And then in a **NEW** terminal:
```
cuihtlauac:
```suggestion
And then in a new terminal:
```

cuihtlauac (changes requested):
Thanks @jonahbeckford. It's great to see this material updated.

I'm still confused about the compatibility status of the examples.

data/tutorials/getting-started/1_02_your_first_ocaml_program.md
jonahbeckford: I tested it with DkML 2.1.0 and it works. So, I removed the "OCaml DKML installer due to a dependency issue" DkML-specific comment.

This is a PR for DkML 2.1.0. I don't think I should be on the hook to test WSL2 and Cygwin. Hopefully ocaml.org uses MDX so the examples can be tested in CI. If you do have CI, it is straightforward to add DkML to CI.

Side-note: I did nothing to fix the issue because I was completely unaware of that comment. It is weird that is works, and that makes me highly suspicious of the DkML-specific comment. I suspect that the comment was really for Dream (next side-note). Hopefully we can file issues in the future if there really was a problem with a simple PPX.

Side-note 2: The earlier Dream example does not work with DkML. But we've already discussed in https://github.com/ocaml/ocaml.org/issues/1962 that Dream has a huge dependency cone and to come up with alternatives.

cuihtlauac: Ok, that's clear, thanks a lot. Maybe we could say something like this:
> **Note:** This example was successfully tested on Windows using DkML 2.1.0

jonahbeckford: That sounds confusing to the user ... in particular, it begs the following questions:
1. How do I tell what version I am on?
2. Why doesn't every example tell me which versions, operating systems and distributions it was tested on?

It seems like a good addition as an HTML comment for ocaml.org maintainers. Is that what you meant?

EDIT: In the interests of time, I added it in with some language on how to check the version. Since I know of at least one user today who was messed up by not having this PR, if you have more minor suggestions can you just add them and submit? Thx.

cuihtlauac (approved):
Thanks @jonahbeckford, let's merge, you are right, this is needed by the community

PR #2250, opened 2024-03-18
Turned exercise 046 into practice

maha-sachin:
Changed exercise 046 into practice


practice/046/answer/impl.ml
```diff
@@ -0,0 +1,21 @@
+
```
cuihtlauac:
```suggestion
```

practice/046/ex.ml
cuihtlauac, line 1: Add tests for each case of the true tables of each logic operator

cuihtlauac (changes requested):

practice/046/work/impl.ml
```diff
@@ -0,0 +1,7 @@
+type bool_expr =
+  | Var of string
+  | Not of bool_expr
+  | And of bool_expr * bool_expr
+  | Or of bool_expr * bool_expr
+let table2 _ = failwith "Not yet implemented"
+
```
divyankachaudhari: Add newline at end of file.

cuihtlauac (approved):

PR #2257, opened 2024-03-18
Turned exercise 059 into practice 059

maha-sachin:


maha-sachin:
@cuihtlauac please review.
Thanks

maha-sachin:
Committed the wrong way, will re-create PR

PR #2258, opened 2024-03-18
Turned exercise 060 into practice 060

maha-sachin:
dune build --root . 
dune test --root . answer


maha-sachin:
@cuihtlauac please review.
Thanks

maha-sachin:
Committed the wrong way, will re-create PR

Issue #2246, opened 2024-03-18
Turn exercises 061A to 068 into practice


cuihtlauac:
This task consists of doing 10 exercises instead of one, as described in issues #2167, #2168, #2170, #2172, #2173, #2176, #2177, #2178, #2179 and #2180. Create 10 folders, and populate them as the others. Makes sure `dune build --root .` and `dune test --root . answer` work.

- [X] 061A #2259
- [X] 061B #2260
- [X] 062A #2261
- [X] 062B #2262
- [X] 063 #2263
- [X] 064 #2264
- [X] 065 #2265
- [X] 066 #2266
- [X] 067 #2267
- [X] 068 #2268

komutumu:
Hello @cuihtlauac, can I take this on?
Assign it to me please.
Let me get started on this sir.
Thanks

komutumu:
As I get started @cuihtlauac  
Can I pick your mind on what the error ```No config found for file practice/001/answer/impl.ml. Try calling 'dune build'.ocamllsp``` signifies? And I did call dune build but perhaps not in the right manner or folder because the issue persists.
Thanks

jahielkomu:
> Hello @cuihtlauac, can I take this on?  
Assign it to me please.
Let me get started on this sir. 
Thanks

Appleeyes:
@cuihtlauac I have made different PRs for different folders, please kindly review

komutumu:
> @cuihtlauac I have made different PRs for different folders, please kindly review

@Appleeyes I was already working on it as shown in the above comments but let me find another one.

Appleeyes:
> > @cuihtlauac I have made different PRs for different folders, please kindly review
> 
> @Appleeyes I was already working on it as shown in the above comments but let me find another one.

ohh so sorry, he pointed me to it @jahielkomu

PR #2260, opened 2024-03-18
Turn-exercise-061B-into-practice

Appleeyes:
This PR is referencing the issue https://github.com/ocaml/ocaml.org/issues/2246

@cuihtlauac please review


practice/061B/answer/impl.ml
cuihtlauac, line 1: - Fix compilation
- Fix eof

cuihtlauac (changes requested):

practice/061B/answer/impl.ml
Appleeyes: I always add new line to the end of line. just does not show on github @cuihtlauac

cuihtlauac (changes requested):
Use the tree from exercice 62B to build a larger test case

cuihtlauac (approved):

PR #2285, opened 2024-03-21
Fix error in practice 005 ex.ml file

MissJae:
This should fix the error on line 9 in the Practice/005/ex.ml.
Changed None to []. 
@cuihtlauac please can you check this?

### DIFF
--- a/practice/037/answer/impl.ml
+++ b/practice/037/answer/impl.ml
@@ -11,7 +11,7 @@ let factors n =
 (* Naive power function. *)
 let rec pow n p = if p < 1 then 1 else n * pow n (p - 1)
 
-let phi_improved n =
+let phi n =
   let rec aux acc = function
     | [] -> acc
     | (p, m) :: t -> aux ((p - 1) * pow p (m - 1) * acc) t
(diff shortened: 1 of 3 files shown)


practice/005/ex.ml
```diff
@@ -0,0 +1,15 @@
+open OUnit2
+
+module type Testable = sig
+  val rev : 'a list -> 'a list
+end
+
+module Make(Tested: Testable) : sig val v : test end = struct
+  let v = "rev" >::: [
+    "nil" >:: (fun _ -> assert_equal [] (Tested.rev []));
```
cuihtlauac: Nice, I had overlooked that. Thanks @MissJae

practice/005/dune
```diff
@@ -0,0 +1,5 @@
+(include_subdirs qualified)
+
+(library
+(name ex) 
+(libraries ounit2))
```
cuihtlauac:
```suggestion
(library (name ex) (libraries ounit2))
```

cuihtlauac (changes requested):
Thanks @MissJae. Only list file `ex.ml` in PR, please.

PR #2288, opened 2024-03-25
Add Platform changelogs for February 2024

tmattio:


cuihtlauac (approved):

tmattio:
cc @emillon and @pitag-ha this is adding changelogs for both Utop and Merlin.
The Dune release process has been working great to have the Dune changelogs in OCaml.org for the releases. I wonder if that's something we can replicate in other projects, as we seem to have to catch up on the changelogs every month. What do you think, is there something else we can do to help make sure we publish changelogs alongside the releases of the other tools?

cuihtlauac:
Do you think this is something we could approach with a GitHub action? Possibly using this:

https://github.com/marketplace/actions/track-and-record-changes-in-other-repos

emillon:
You can fetch that using the graphql API:

```graphql
query { 
  repository(owner:"ocaml", name:"dune"){
    releases(last:3) {
      nodes {
        description
      }
    }
  }
}
```

(that's filled by `dune-release` using the changelog file)

PR #2323, opened 2024-04-07
Implement dark mode switch 

oyenuga17:
Hi @sabine  so i played around with this over the weekend and I managed to pull it off.
Here, is a [video](https://www.loom.com/share/13a41ad2bd16445b87969621bc5e6f51?sid=e7ae45c3-44ee-4eaf-81cc-5d72f58f1100) of the implementation. Please let me know what you think 😊


sabine (commented):
Hey @oyenuga17! Thanks for the contribution! This looks very nice.

Here's a few issues:
- ESC key cannot be used to close the modal
- clicking outside the modal does not close the modal

I think that having a HTML/JS/CSS modal here is probably too complex for what is needed here. It will be much simpler to use the browser's built in `confirm()` method (see https://developer.mozilla.org/en-US/docs/Web/API/Window/confirm?retiredLocale=de) to obtain confirmation from the user.

Alternatively, we can update the privacy policy to list that using the dark/light mode switch will use LocalStorage to remember the user's setting.

oyenuga17:
hmm... you're right @sabine, might be hard to achieve a fully accessible modal using this my approach. I'll try the window.confirm method since that's the closest thing to what we're trying to achieve.

oyenuga17:
@sabine I just mad the change. Please help review again

sabine:
Thanks @oyenuga17, I did some refactoring with the intent to improve code clarity. :heart: Let's merge this now!

sabine (approved):

oyenuga17:
Thank you @sabine ❤️ 🚀

PR #2343, opened 2024-04-15
OCaml 5.2.0-beta2 changelog entry

Octachron:
This PR adds the changelog entry for the upcoming beta2 release of OCaml 5.2.0 (which should happen tomorrow).


Octachron:
The opam packages has been merged to the repo, so the PR can be merged now.

PR #2345, opened 2024-04-15
fix: Serve manual under /lts and /latest URLs

### LINKED ISSUE #2344: OCaml.org Set Documentation Missing
https://ocaml.org/api/Set.S.html The set documentation on this page is down.

sabine:
Resolves https://github.com/ocaml/ocaml.org/issues/2344

- make documentation links point to /manual/lts


Octachron:
I am not sure if we want to link to `lts` rather than `latest`: my issue is that the existence of a `lts` release distinct from the `lastest` release is temporary : it will last until OCaml 5 has restored full features and OS supports compared to OCaml 4. So hopefully, not beyond 2026 .

Another issue is that for the `Atomic` module, we definitively want to link a multicore enabled version, which means *not* the lts version which is a sequential dummy implementation of the interface.

PR #2352, opened 2024-04-16
Add /api/** redirection

mtelvers:
Doing a Google search of `ocaml map` (for example) returns these two URLs as the top results:

* https://ocaml.org/api/Map.Make.html
* https://ocaml.org/api/Map.html

With the current code base, both of these links are broken.  I propose this PR as a quick fix.

### DIFF
--- a/src/ocamlorg_web/lib/redirection.ml
+++ b/src/ocamlorg_web/lib/redirection.ml
@@ -486,6 +486,13 @@ let default_index_html =
     ("/manual/5.3/api", Url.api_with_version "5.3");
   ]
 
+let redirect_to_latest pattern =
+  let handler req =
+    let target = Dream.target req in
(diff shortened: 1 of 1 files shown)


cuihtlauac (approved):
That makes sense

.ocamlformat
cuihtlauac, line 1: This should not be part of this PR

cuihtlauac (dismissed):

.ocamlformat
mtelvers: The perils of `git commit -a`.  I removed the `version = 0.25.1` to allow me to format the project with the current version of `ocamlformat`, which was the version that I had installed.  Perhaps we should also upgrade the `ocamlformat` version requirement to `0.26.1`?

sabine (approved):

PR #2358, opened 2024-04-17
Handle lts, default and missing version in middleware

### LINKED ISSUE #2357: Manual pages without a version in the URL return an empty body and no redirect
[Example](https://ocaml.org/manual/intfc.html)

Google currently links to the manual in this format. It should probably redirect to /manual/5.1

cuihtlauac:
Fix Issue #2357

New redirects also introduced
* X into X.max
* X.Y.Z into X.Y

### DIFF
--- a/src/ocamlorg_web/lib/middleware.ml
+++ b/src/ocamlorg_web/lib/middleware.ml
@@ -9,7 +9,7 @@ let no_trailing_slash next_handler request =
   if Dream.target request = target then next_handler request
   else Dream.redirect request target
 
-let versioning next_handler request =
+let language_manual next_handler request =
   let init_path = request |> Dream.target |> String.split_on_char '/' in
   let path =
     let minor (release : Data.Release.t) = Ocamlorg.Url.minor release.version in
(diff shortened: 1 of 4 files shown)


src/ocamlorg_web/lib/middleware.ml
```diff
@@ -9,6 +9,24 @@ let no_trailing_slash next_handler request =
   if Dream.target request = target then next_handler request
   else Dream.redirect request target
 
+let versioning next_handler request =
```
sabine: would use a clearer name because this relates only to the ocaml compiler, not just any versioning. This is more of a `language_manual_middleware`, right?

It seems wrong to me to apply this middleware to all routes, when it should only be applied on the manual routes.

sabine (commented):
Seems reasonable overall. Just one point in the comment below.

src/ocamlorg_web/lib/middleware.ml
cuihtlauac: `Middleware.language_manual_middleware` or  `Middleware.language_manual` ?

cuihtlauac: Or maybe `Middleware.language_manual_version` ?

Issue #2361, opened 2024-04-18
(Data model) Add "event_type" Field to Events


sabine:
An event can either be
* a meetup
* a conference
* a seminar
* a hackathon
* a retreat

For this, extend the data model in [`event.ml`](https://github.com/ocaml/ocaml.org/blob/main/tool/ood-gen/lib/event.ml) to have a new field `event_type` that records what type the event is. For inspiration on how to do this, look at the "difficulty" field in [`book.ml`](https://github.com/ocaml/ocaml.org/blob/main/tool/ood-gen/lib/book.ml).

Both the recurring and upcoming events need to be extended. For an upcoming event that has a linked recurring event, this field will be populated from the linked recurring event.

Events in `data/event` will need to be updated.

Also, we need to render the new information as part of the Event card (as a tag below the event title).

csaltachin:
Hi! I'd like to take a shot at this. A few questions:

- Do we want to require that the event type is always specified both in `data/event/*.md` files _and_ in `data/event/recurring.yml` entries? Or do we want to allow e.g. an upcoming event `2024-foo.md` to omit its event type if it has a linked recurring event that does specify an event type?
- Looking at the [`decode`](https://github.com/ocaml/ocaml.org/blob/1cb5e57c81f9311930de53f92cdb7f41e72ef056/tool/ood-gen/lib/event.ml#L64) function in `event.ml`, should it be an error (and should we throw) if an upcoming event and its linked recurring event have different event types? Or should we always use the type from the recurring event if any?
- I'm not super familiar with the site's design system, but as for rendering, it looks like we just need to add a couple divs in the `events.eml` template, with a similar Tailwind class as [the other Event card tags](https://github.com/ocaml/ocaml.org/blob/1cb5e57c81f9311930de53f92cdb7f41e72ef056/src/ocamlorg_frontend/pages/events.eml#L23), and hydrated via `event.event_type` and `recurring_event.event_type`. Does this sound roughly correct? Do we want to render a different icon (if any) for each event type?

sabine:
> Do we want to require that the event type is always specified both in data/event/*.md files and in data/event/recurring.yml entries? Or do we want to allow e.g. an upcoming event 2024-foo.md to omit its event type if it has a linked recurring event that does specify an event type?

Yes, `event_type` can be `None` for events that have a recurring event linked.

[...]

PR #2366, opened 2024-04-20
Add event_type field to Events, and render tag in Event cards

csaltachin:
Addresses #2361. This PR:

- adds a new `event_type` field to the types `t` and `RecurringEvent.t` (and their respective metadata types) in `tool/ood-gen/lib/event.ml`,
- updates the `Event` module signature in `src/ocamlorg_data/data.mli`,
- adds an `event_type` value to the entries in `data/events/recurring.yml` and the `data/events/*.md` files, and
- makes a first attempt at rendering a tag in Event cards (for both upcoming and recurring events) on the Events page:
  <details><summary>(expand image)</summary>
  
  </details>

Some notes:

- Entries in `recurring.yml` must always specify an `event_type` value. `*.md` files can omit them, but only if they have a linked recurring event.
- An upcoming event's `event_type` value is populated from either its `*.md` metadata, or its linked recurring event (if there is one). If these two sources specify different event types, or if neither specifies one, it is an error (and an exception is raised in `tool/ood-gen/lib/event.ml` when that happens).
- The event type tags in `src/ocamlorg_frontend/pages/events.eml` use the tag colors from [the Community pages design draft](https://www.figma.com/file/7hmoWkQP9PgLTfZCqiZMWa/OCaml-Community-Pages?type=design&node-id=973-5423&mode=design&t=NSEmv1UycNyd52rF-4).
- Do we also want to render these tags in other pages, e.g. the Upcoming Events section of the Community landing page?

### DIFF
--- a/data/events/2024-02-09-oups.md
+++ b/data/events/2024-02-09-oups.md
@@ -1,5 +1,6 @@
 ---
 title: "OCaml Users in Paris (OUPS)"
+event_type: meetup
 textual_location: Paris, France
 location:
   lat: 48.8566
--- a/data/events/2024-03-08-s-repls.md
+++ b/data/events/2024-03-08-s-repls.md
@@ -1,5 +1,6 @@
 ---
 title: "S-REPLS 14"
+event_type: seminar
 textual_location: "London, United Kingdom"
 url: https://www.cl.cam.ac.uk/events/s-repls14/
 recurring_event_slug: s-repls-programming-languages-seminar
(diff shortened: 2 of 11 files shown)


christinerose (approved):

sabine (approved):
Hey @csaltachin, this is awesome! Literally no suggestions for improvements here. :sparkles: :heart:

On the question of whether we want to render the tags on other pages as well that display events: Yes, I think we do want to consolidate the event card to be a component that is used in different places. It can be a different PR because I am merging this here just now!

PR #2376, opened 2024-04-25
Cookbook database recipes

cuihtlauac:
Adapted from @F-Loyer recipes


christinerose (commented):
Please review my changes to ensure they're all still accurate. 

Also, in li 49, the sentence "Then when this query is called, `()` is returned (`Ok ()` to be accurate)" is confusing. 

"`()` is returned" — that's clear, but "`Ok ()` to be accurate" confuses me. If `Ok()` is more accurate, would it be better just to say "Then when this query is called, `Ok ()` is returned"?

F-Loyer:
> Please review my changes to ensure they're all still accurate.
> 
> Also, in li 49, the sentence "Then when this query is called, `()` is returned (`Ok ()` to be accurate)" is confusing.
> 
> "`()` is returned" — that's clear, but "`Ok ()` to be accurate" confuses me. If `Ok()` is more accurate, would it be better just to say "Then when this query is called, `Ok ()` is returned"?

Yes... "Then when this query is called, `Ok ()` is returned"? would be better.

christinerose:
> Yes... "Then when this query is called, `Ok ()` is returned"? would be better.

Done!

christinerose (approved):

PR #2414, opened 2024-05-07
(data) jobs: add a XenServer position again

edwintorok:
Follow up to https://github.com/ocaml/ocaml.org/pull/2387 where I removed the links to the positions that got filled.
We have a new job opening again, with a new link.


cuihtlauac (approved):
Thanks @edwintorok, this is nice

PR #2431, opened 2024-05-15
Support building with Dune pkg

tmattio:


cuihtlauac:
Thanks @tmattio, how is Dune pkg supposed to work?

tmattio:
> Thanks @tmattio, how is Dune pkg supposed to work?

It is still experimental, but it is supposed to work with the configuration in this PR. The changes here shouldn't have an impact on existing developer workflows, so I'm going to merge this so we can maintain the configuration while the Dune team continues to work on Dune package management.

PR #2432, opened 2024-05-15
Blog filters

cuihtlauac:
Filter blog by source id e.g.

https://ocaml.org/blog/emelle-tv
https://ocaml.org/blog/cwn
https://ocaml.org/blog/opam (local)

P.S. Includes PR #2430


christinerose (approved):

sabine (commented):
Still don't think this is a good idea because it conflates the concept of the blog hosted on ocaml.org and a feed that we pull in from outside sources for the OCaml Planet.

People can already see the individual feeds at their original location, we're not giving them anything useful here.

If we could filter the OCaml Planet feed posts by topic, however, that would be super useful.

PR #2442, opened 2024-05-21
Case insensitive substring test function

cuihtlauac:


src/ocamlorg_web/lib/handler.ml
```diff
@@ -172,9 +172,7 @@ let academic_users req =
     let open Data.Academic_institution in
     let pattern = String.lowercase_ascii pattern in
     let name_is_s { name; _ } = String.lowercase_ascii name = pattern in
-    let name_contains_s { name; _ } =
-      String.contains_s (String.lowercase_ascii name) pattern
-    in
+    let name_contains_s { name; _ } = String.is_sub_ignore_case pattern name in
```
Octachron, line 175: It seems a bit perilous to assume that names fit in the ascii subset of English?

cuihtlauac: You are right; it is not. Not sure how to handle this nicely here. This commit is just a refactoring. Any clues?

Octachron: If pulling `uucp` as dependency is ok: https://erratique.ch/software/uucp/doc/Uucp/Case/index.html#caselesseq looks like a good alternative to the `lowercase_ascii` normalization (even if does not cover "lookalike" search).

cuihtlauac: This is a good idea. Uucp is already part of ocamlorg dependencies (by transitivity) so there's no downside to this.

cuihtlauac: Issue on this: https://github.com/ocaml/ocaml.org/issues/2444

PR #2443, opened 2024-05-21
Filter OCaml Planet Blog posts for "caml" keyword

cuihtlauac:
Only signal-&-threads has been filtered, as a proof of concept. If we decide to go for this, everything has to be scraped from scratch once.

For historical reasons, we search using "caml" and not "ocaml" because some feeds still use the old name


christinerose (approved):

data/changelog/merlin/2024-06-03-merlin-54.15.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Merlin 4.15 for OCaml 4.14 and 5.1
+tags: [merlin, platform]
+changelog: |
+  + merlin binary
+    - destruct: Removal of residual patterns (ocaml/merlin#1737, fixes ocaml/merlin#1560)
+    - Do not erase fields' names when destructing punned record fields (ocaml/merlin#1734,
```
christinerose: Just checking, is "punned" correct here or should it be "pinned"?

christinerose (approved):

sabine (approved):

PR #2449, opened 2024-05-22
Handle translation redirection in middleware

cuihtlauac:


sabine (commented):
Don't think this is a good change.

These are legacy URLs that are unlikely to be revived, even if i18n is implemented at some point.

Feels to me like we're obfuscating the fact that there's just a small set of legacy URLs we keep redirecting because of possible backlinks.

PR #2452, opened 2024-05-23
Update is_ocaml_yet/gui.md: Plotting

lukstafi:
While procrastinating I noticed there were only 2 plotting libraries mentioned, while we have an abundance: https://discuss.ocaml.org/t/ocaml-plotting-libraries/13384/4


christinerose (approved):

sabine:
Thank you @lukstafi!

PR #2454, opened 2024-05-24
web.md: jsonchema->atd exists

Khady:
[jsonschema2atd](https://github.com/ahrefs/jsonschema2atd) has been published on opam and is in a relatively usable state with support for json schema and openapi definitions.

The world of json schema and openapi is a huge mess with different versions of the specifications and most document actually have a mix of multiple ones at the same time. So there will probably never be a fully working solution. This should give a head start a least.


cuihtlauac (dismissed):
Hi @Khady, thanks for this. Shouldn't you link to your tool too?
[jsonschema2atd](https://github.com/ahrefs/jsonschema2atd)

Khady:
I don't really know in which section it is supposed to appear.

sabine:
I added it under "External Web APIs"

sabine (approved):

Issue #2469, opened 2024-05-30
compilerlibref api is 404


cuihtlauac:
https://ocaml.org/manual/5.2/api/compilerlibref is 404

sabine:
Where does this URL occur?



Here's another one that's coming from Google search results:

Bug:

https://ocaml.org/manual/attributes.html

redirects to

https://ocaml.org/manual/5.2/attributes.html/index.html

sabine:
The open PR doesn't yet address the https://ocaml.org/manual/5.2/api/compilerlibref URL / redirect

PR #2470, opened 2024-05-30
fix: language manual redirect, remove unnecessary append of index.html

sabine:
Related:  #2469

Bug:

https://ocaml.org/manual/attributes.html

redirects to

https://ocaml.org/manual/5.2/attributes.html/index.html

This patch should fix this.


cuihtlauac:
Some 404 remains, e.g.:

http://localhost:8080/manual/api/compilerlibref
http://localhost:8080/manual/latest/

This branch fixes them: https://github.com/ocaml/ocaml.org/tree/fix-more-redirect (and makes the code more readable)

PR #2540, opened 2024-06-23
Cookbook Measures of Central Tendency

ggsmith842:
Add Measures of Central Tendency title with tasks for calculating the mean and median


data/calculate-the-median/00-stdlib.ml
```diff
@@ -0,0 +1,35 @@
+---
+packages: []
+discussion: |
+  - **Description:** The [Median](https://statistics.laerd.com/statistical-guides/measures-central-tendency-mean-mode-median.php) measure of central tendency
+  is the middle value in an ordered set of data. It is useful for statistical analysis as it isn't easily influenced by extreme values (outliers)
+  - **Data Assumption:** This code assumes the data is of type float. While we can calculate the median for any numeric value, it makes sense to use floats in OCaml since all
+  integers i.e 1 can be represented as floats 1. but not all floats can be represented by integer data types i.e. 1.5 does not equal 1 or 2
```
christinerose:
```suggestion
  - **Data Assumption:** This code assumes the data is of type float. While we can calculate the median for any numeric value, it makes sense to use floats in OCaml since all
  integers can be represented as floats (i.e., 1 can be written 1.0), but not all floats can be represented by integer data types. For example, 1.5 does not equal 1 or 2.
```

christinerose: Does my suggestion clarify your sentence, @ggsmith842?

ggsmith842: @christinerose yes it's much clearer now! Thank you

christinerose (approved):

cuihtlauac (changes requested):
I'm not sure we should have recipes that merely translate simple math formulas into code. Nowadays, this faster handled with LLM-generated code.

ggsmith842:
@cuihtlauac that's fair. I'm happy to close this PR and start on something else. I wanted to add something related to Statistics or Probability in the math section. Do you have any recommendations?

RCHG:
Hello.
 I suggest to add a simple linear regression in the recipes which is very easy to implement and it is an standard statistical method. I can do a PR with it.

sabine (commented):
Looking at this, I'm coming to the same conclusion as @cuihtlauac: there is no task here that
(a) provides meaningful value to a developer looking to solve a production task (LLMs can generate such code pretty reliably)
(b) the problem is not complex enough to warrant comparison of packages.

That's why I'm closing this PR. But indeed we are very much happy to receive PRs for other statistics tasks. (It is good to open a draft PR just proposing a new tasks with a clear description of what the task entails, if unsure about viability.)

PR #2568, opened 2024-07-02
Explain folders bin, lib and _build

### LINKED ISSUE #2561: Explain `bin` and `lib` folder naming convention
In Unix folders named:
* `bin` contains _compiled_ executable files (e.g. /bin or /usr/bin)
* `lib` contains _compiled_ library files (e.g. /lib or /usr/lib)

In OCaml/Dune projects those folders often contain _sources_ corresponding to executables or libraries.

Tutorials on executables, libraries, modules and Dune should clarify this.

From: https://discuss.ocaml.org/t/sorry-not-getting-it-with-mli/14883

Also, tutorials use Dune without explaining the role of the `_build` folder, which is problematic.

cuihtlauac:


christinerose (approved):

MisterDA:
The Unix standard doesn't say anything about user `lib` or `bin` directories, but rather about specific directories. I think we should reference the [Filesystem Hierarchy Standard](https://en.wikipedia.org/wiki/Filesystem_Hierarchy_Standard) instead, and just say that by convention OCaml project sources mimic this hierarchy: source files in the `bin` directory usually produce executables that may go in `/usr/bin` and source files in the `lib` directory usually produce libraries that may go in `/usr/lib` (but note that if opam is used all the artifacts will end up under the opam root!).
I don't know if the tutorial is aimed at unix systems only, but for instance `eval $(opam env)` wouldn't work on cmd or PowerShell. It's also using _folder_ and _directory_, in all these cases _directory_ [is more appropriate](https://stackoverflow.com/questions/5078676/what-is-the-difference-between-a-directory-and-a-folder).

christinerose (approved):

PR #2573, opened 2024-07-02
Cookbook: recipes for parse-command-line-arguments

richardhuxton:
One example with Cmdliner, and two from the standard library (Sys.argv and Arg),

Happy to rework them if I've missed some details.

### DIFF
--- a/data/cookbook/parse-command-line-arguments/00-stdlib.ml
+++ b/data/cookbook/parse-command-line-arguments/00-stdlib.ml
@@ -4,10 +4,10 @@ discussion: |
   For a longer introduction, see the [Command-line arguments](https://ocaml.org/docs/cli-arguments) tutorial in the docs.
   The standard library documents the [Arg module](https://ocaml.org/manual/latest/api/Arg.html).
 ---
-(* At the lowest-level we can interact with `Sys.argv` and access arguments by index.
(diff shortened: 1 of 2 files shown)


christinerose (approved):

christinerose (approved):

sabine (commented):
These recipes have two problems:

* cmdliner example focuses on one oddly-specific task and doesn't cover enough API surface (for comparison, look at https://rust-lang-nursery.github.io/rust-cookbook/cli/arguments.html#parse-command-line-arguments)
* stdlib example does not implement the same task as cmdliner example. A point of the cookbook is comparing the exact same task between different libraries / stdlib, so all the recipes for a task need to implement the same task. To facilitate this, it can help to add a description for the task, when it is unclear.

ETA: also, as a general remark, avoiding to `open` the modules in question does a lot to help people see where values are coming from (we do not have a LSP running here that would allow to inspect the code).

cuihtlauac:
This needs to be rebased against a fresh `main`

sabine:
Rebased, and we need to integrate with https://github.com/ocaml/ocaml.org/pull/3038

sabine (approved):
Merging this instead of https://github.com/ocaml/ocaml.org/pull/3038/files because this is better for the intended copy-paste-modification workflow. Thanks @richardhuxton!

PR #2552, opened 2024-06-26
Cookbook Extract Links from HTML

ggsmith842:
Adds an example for the extract-links-from-html task using the `Re` library. A sample html string is provided and the usage example shows how to read an HTML and then print the links found.

PR #2581, opened 2024-07-05
Cookbook Check a Webpage for Broken Links

ggsmith842:
Adds recipe for validating links on a webpage. For this example the webpage is represented as an html string. This code makes use of the cohttp and lambdasoup libraries.

### DIFF
--- a/data/cookbook/check-webpage-for-broken-links/00-cohttp_lambdasoup.ml
+++ b/data/cookbook/check-webpage-for-broken-links/00-cohttp_lambdasoup.ml
@@ -1,3 +1,4 @@
+---
 packages:
   - name: "lambdasoup"
     tested_version: "1.1.0"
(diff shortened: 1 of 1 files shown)


christinerose:
It shows the library as Lambda Soup here:
https://opam.ocaml.org/packages/lambdasoup/

Otherwise, perhaps put `lambdasoup` in monospace.  ??

yawaramin:
I think this makes #2552 redundant? They both show how to parse a web page and grab all the hyperlinks inside it.

sabine:
Reevaluating the usefulness of the tasks in terms of evaluating packages, I've come to the conclusion that it's best to remove the task (since it will always require a combination of different libraries for making the HTTP requests and for the HTML parsing). Thus, I removed the task in https://github.com/ocaml/ocaml.org/pull/2851.

With the lambdasoup recipe merged in https://github.com/ocaml/ocaml.org/pull/2552, I think we got the useful parts covered. :heart:

PR #2584, opened 2024-07-07
Cookbook: "Timeout with Promises": Lwt, Async

richardhuxton:
Note: I did try to add a recipe with Eio too but it didn't seem to like that. I will raise an issue with more details.


christinerose (approved):

sabine (approved):
@richardhuxton thank you for this recipe. I edited it to hopefully make it more reasonable to copy paste.

Since it is doing an interesting real world task that goes beyond just creating and awaiting a promise (which is probably not even a very interesting task, as the earlier "creating and awaiting promises" recipe shows), I'm adding this as an entirely new task.

We try not to use PPXes (unless unavoidable) in cookbook recipes because that makes using the code example in utop more complicated and requires people to set up a dune config to run the ppx. However, making people aware of the possibility to use `ppx_let` is great.

@christinerose ready for line editing

Issue #2583, opened 2024-07-07
Link to specific platform installation instructions


dra27:
https://ocaml.org/install automatically displays the Windows page when viewed from a Windows machine, but it would be nice if it updated the URL with an anchor (i.e. such that https://ocaml.org/install#windows activates the Windows instructions anywhere and https://ocaml.org/install#{unix,linux,macos,bsd,etc.} activates the Linux instructions) so that it can be referenced directly.

sabine:
Yes, this would be nice, and also it's a small task that could be contributed by a newcomer.

Gbury:
I tried using the new anchors, but on my machine (a linux one), if I try to go explicitly to the windows instructions from a link (i.e. https://ocaml.org/install#windows or even by refreshing the webpage after clicking on the windows instructions), the page automatically redirects to the linux installation instructions, which is really counter-intuitive and annoying, since the link was explicitly to the windows instructions.

I guess that's why this issue is not closed yet ?

piotr-m-jurek:
What would be the ideal scenario to resolve this issue?
1. when someone sends me the link `https://ocaml.org/install` I am presented with the `install` page and none of the options are chosen
2. when someone sends me the link `https://ocaml.org/install#windows` and it matches my system I'm presented with the `install#windows` page (same for the other option).
3. when someone sends me the link `https://ocaml.org/install#windows` and it doesn't match my system, I'm being redirected to `install#linux_mac_bsd` page.

Gbury:
I'd say the ideal scenario would be to have all the following points be true:
- when someone follows a link to `https://ocaml.org/install`, redirect to the page that matches the user's system
- when someone follows a link to `https://ocaml.org/install#foo`, present the page for `foo` (i.e. either `windows` or `linux_mac_bsd`) regardless of the system. If the anchor is part of the url being clicked on/followed, then it means that there was some intent to link to these specific instructions, and overriding it seems not great.

PR #2600, opened 2024-07-17
Issue #2583: Added OS Anchor Tags to ocaml.org/install

SisyphianLiger:
Hi Ocaml Team,

I decided to solve issue #2583 today, to get into the Ocaml ecosystem (because I spend to much time on twitter talking about how great Ocaml is).

That said, there are two modifications I have added to the url in the install.eml file. The first is that when a user selects the ocaml.org/install tag, they url shown is ocaml.org/install#<OS>

There is already a detect_os() function so I added onto that. 

When you arrive on the page, the function anchor_os_loc() is called, and the anchor is set to ocaml.org/install#<anchor from detect_os>

Then if, say, you are showing your friends that it's not just for mac/linux/bsd, it's for windows, you click the windows button, and the anchor tag will change. You can toggle then back and forth between the two, so that you know you are looking at the correct OS related information from the url.


cuihtlauac (dismissed):
Thanks @SisyphianLiger, that sounds nice. A couple of remarks/questions?

1. Why do you copy the contents of `url.ml` into `import.ml`?
2. Could you limit the changes in to `install.eml` to what's really new in it? Changing the indentation in all the lines makes it harder to review.
3. Also, CI tests are failing

SisyphianLiger:
Hi @cuihtlauac,

1. That is a good question... most likely happened when I tinkered with the std.lib os to try to figure out where I needed to implement these changes and copied over a file that I should not have. I thought I had fixed it before making the PR but that is definitely not the case. I have just gone and reverted the ocaml.org current `url.ml` and `import.ml` from github to fix that.  Also had a url 2.ml...that has been deleted.

2. I can see what you mean, unfortunately the PR says changed all of the `install.eml`, in the next request I have checked with git diff and can see that now the changes for`install.eml` should only be visible. 

3. From the failed CI test, I can see that the command `dune build --auto-promote @fmt` is what has failed specifically for the Ubuntu version, I have since ran it locally and found an extra `end` keyword that has been deleted, most likely from problem 1. After deletion I do not get any formatting errors, so I hope the problem has been resolved!

sabine (approved):
Thanks @SisyphianLiger, this is very useful and looks good to me now! :orange_heart:

PR #2606, opened 2024-07-19
Fix Example referencing Type not yet Defined

avlec:
Fix an example that used a type that was not yet defined.

This caused me some headache when going through the tutorials thinking I had missed something but it turned out the example was using a type that was defined in a below section.

### DIFF
--- a/data/tutorials/language/0it_00_values_functions.md
+++ b/data/tutorials/language/0it_00_values_functions.md
@@ -162,8 +162,8 @@ Single-case pattern matching can also be used for parameter declaration.
 
 Here is an example with tuples:
 ```ocaml
-# let get_country ((country, { first; last }) : citizen) = country;;
-val get_country : citizen -> string = <fun>
+# let get_balance ((balance, { first; last }) : int * name) = balance;;
+val get_balance : int * string -> int = <fun>
 ```
 
 Here is an example with the `name` record:
--- a/data/tutorials/language/0it_00_values_functions.md
+++ b/data/tutorials/language/0it_00_values_functions.md
@@ -162,8 +162,8 @@ Single-case pattern matching can also be used for parameter declaration.
 
 Here is an example with tuples:
 ```ocaml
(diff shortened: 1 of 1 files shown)


data/tutorials/language/0it_00_values_functions.md
```diff
@@ -162,8 +162,8 @@ Single-case pattern matching can also be used for parameter declaration.
 
 Here is an example with tuples:
 ```ocaml
-# let get_country ((country, { first; last }) : citizen) = country;;
-val get_country : citizen -> string = <fun>
+# let get_balance ((balance, { first; last }) : int * name) = balance;;
+val get_balance : int * string -> int = <fun>
```
cuihtlauac:
```suggestion
# let get_country ((country, { first; last }) : string * name) = country;;
val get_country : string * name -> string = <fun>
```

Let's only replace `citizen` by `string * name`.

cuihtlauac (changes requested):
Thanks @avlec. You are right; the type was not yet defined. 

I suggest we keep the old example, only fix the type.

avlec:
Thanks for the quick feedback @cuihtlauac that sounds quite reasonable.

I was wondering what would be thought about reducing the amount of context shared between these examples. Simplifying them in that way I think makes it easier for the reader to understand the syntax by having less things they need to keep track of.

cuihtlauac:
> Thanks for the quick feedback @cuihtlauac that sounds quite reasonable.
> 
> I was wondering what would be thought about reducing the amount of context shared between these examples. Simplifying them in that way I think makes it easier for the reader to understand the syntax by having less things they need to keep track of.

Yes, that's a good idea. 
* Sections understandable on their own
* Changes easier, less coupling

It would have to be done in separate PR

cuihtlauac (approved):

avlec:
> It would have to be done in separate PR

Thanks for being receptive to that. I'll see if I can put something together as I work through the tutorials

christinerose (approved):

PR #2630, opened 2024-08-02
Events Page Redesign

oyenuga17:
@sabine  please help me put on staging.


sabine (commented):
image

It would make sense to list only those locations that are actually there in the upcoming events.

Would also consider giving the URL query parameters more descriptive names.

oyenuga17:
> image
> 
> It would make sense to list only those locations that are actually there in the upcoming events.
> True that makes sense i will look into this.

> Would also consider giving the URL query parameters more descriptive names.
Okay. This was what I did initially but I thought the parameters were too long so I shortened to include just the letters

sabine (approved):
Merging, thanks @oyenuga17! :sparkles:

PR #2631, opened 2024-08-02
implement jobs feed/redesign

oyenuga17:
@sabine please can you help look into the job feed here : https://github.com/oyenuga17/ocaml.org/blob/a5ed9b99b8e268b61a4eddc7fd2b78d05e223d75/tool/ood-gen/lib/job.ml#L8-L29. I used the same implementation for the changelog feed but for some weird reason only 3 jobs are being populated in the job.xml file.

### DIFF
--- a/src/ocamlorg_frontend/pages/jobs.eml
+++ b/src/ocamlorg_frontend/pages/jobs.eml
@@ -27,7 +27,7 @@ OCaml community."
             </h5>
             <div class="ml-4 flex items-center">
                 <%s! Icons.rss "h-6 w-6 md:h-8 text-primary dark:text-dark-title pb-1" %>
-              <a href="/job.xml" class="text-primary dark:text-dark-white hover:dark:text-dark-primary hover:underline dark:underline ml-2">Subscribe by RSS to receive all Jobs
+              <a href="/jobs.xml" class="text-primary dark:text-dark-white hover:dark:text-dark-primary hover:underline dark:underline ml-2">Subscribe by RSS to receive all Jobs
                 <span class="sr-only">Job RSS</span>
               </a>
             </div>
--- a/src/ocamlorg_frontend/pages/jobs.eml
+++ b/src/ocamlorg_frontend/pages/jobs.eml
@@ -11,26 +11,37 @@ OCaml community."
             <h1 class="text-title dark:text-dark-title font-bold mb-2">OCaml Job Board</h1>
             <p class="text-content dark:text-dark-content text-xl lg:max-w-[80%]">This is a space where groups, companies, and organisations can advertise their projects directly to the
                 OCaml community.</p>
+               <div class="flex lg:flex-row mt-8 mr-4 max-w-[230px]">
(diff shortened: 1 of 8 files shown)


tool/ood-gen/lib/job.ml
```diff
@@ -5,6 +5,29 @@ let all () =
...
+module JobFeed = struct
+  let create_entry (job : t) =
+    let content = 
+      Syndic.Atom.Text (Printf.sprintf "Company: %s\nLocations: %s\n" job.company (String.concat ", " job.locations)) 
+    in
+    let id = Uri.of_string (job.link) in
+    let authors = (Syndic.Atom.author "Ocaml.org", []) in
+    let updated = match job.publication_date with
+      | Some date_str -> Syndic.Date.of_rfc3339 (date_str ^ "T00:00:00-00:00")
+      | None -> Syndic.Date.of_rfc3339 "1970-01-01T00:00:00Z" in
```
sabine: I think the jobs are being filtered out because the default date is more than 365 days in the past. Since we don't have so many jobs that we'd need to limit, probably best to not have a cutoff, and also assume that any job that doesn't have a date set is very recent

tool/ood-gen/lib/job.ml
```diff
@@ -5,6 +5,29 @@ let all () =
...
+    Syndic.Atom.entry ~content ~id ~authors ~title:(Syndic.Atom.Text job.title)
+      ~updated
+      ~links:[ Syndic.Atom.link id ]
+      ()
+
+  let create_feed () =
+    let open Rss in
+    () |> all
+    |> create_entries ~create_entry ~days:365
```
sabine:
```suggestion
    |> create_entries ~create_entry
```

tool/ood-gen/lib/job.ml
oyenuga17: Nice, wasn't sure what that was doing. Thank you ❤️

src/ocamlorg_frontend/pages/jobs.eml
```diff
@@ -11,26 +11,37 @@ OCaml community."
             <h1 class="text-title dark:text-dark-title font-bold mb-2">OCaml Job Board</h1>
             <p class="text-content dark:text-dark-content text-xl lg:max-w-[80%]">This is a space where groups, companies, and organisations can advertise their projects directly to the
                 OCaml community.</p>
+               <div class="flex lg:flex-row mt-8 mr-4 max-w-[230px]">
+                <%s!  Hero_section.hero_button ~left_icon:(Icons.edit "w-6 h-6") ~right_icon:(Icons.link "w-5 h-5") ~text:("Add a Job Offer") ~href:("https://github.com/ocaml/ocaml.org/blob/main/CONTRIBUTING.md") "" %>
```
sabine: Please make sure that the link for contributing always goes to the correct section in the CONTRIBUTING.md document (this is true for all the links that invite people to contribute).

Here, it would be https://github.com/ocaml/ocaml.org/blob/main/CONTRIBUTING.md#content-job

sabine (commented):
Awesome, merging and setting up the RSS feed to automatically post to Twitter, LinkedIn and Mastodon! :sparkles:

PR #2632, opened 2024-08-04
Cookbook for invert a matrix using owl library

RCHG:
Hello. I have created a cookbook recipe for "invert a matrix" using the Owl library. My 1st cookbook, so happy to introduce changes/improvements where needed.


christinerose:
Hi @RCHG! Thanks for requesting my review. Forgive the delay, as I was on holiday for the past two weeks. Just mostly minor grammar/formatting things. I think I caught a missing word `close to zero` rather than `close zero`; however, if I'm mistaken, let's work together to determine what you meant there.

RCHG:
Thanks for the update.  You are right: the statement should be "close to zero". In pure math they have to be zero but, I guess, in computer algorithms we can only guarantee a tiny float number. Everything seems to be fine now.

christinerose (approved):

data/cookbook/invert-matrix/01-owl.ml
```diff
@@ -0,0 +1,35 @@
+---
+packages: [owl]
+---
+
+(* Creates an Owl `float` matrix using random numbers drawn from a uniform distribution with `Owl.Mat.uniform`. It is very likely that it will be invertible but we ensure it by checking that the determinat is not zero with `Owl.Linalg.D.dat`. The dimensions of the matrix are defined by `m`. *)
```
christinerose, line 5: Should the word be "determinant" or is "determinat" correct?

christinerose:
```suggestion
(* Creates an Owl `float` matrix using random numbers drawn from a uniform distribution with `Owl.Mat.uniform`. It is very likely that it will be invertible, but we ensure it by checking that the determinat is not zero with `Owl.Linalg.D.dat`. The dimensions of the matrix are defined by `m`. *)
```
Should the word be "determinant" or is "determinat" correct?

christinerose (commented):
Minor grammar correction and a possible typo

RCHG:
> Minor grammar correction and a possible typo

Thank you, yes it should be determinant.  Also it should be `Owl.Linalg.D.det` instead of `Owl.Linalg.D.dat`

christinerose (approved):

RCHG:
This seems to be finished and ready, but their integration seems frozen since Aug. Do I need to do something more? Thanks.

sabine:
Hey @RCHG, I am very sorry, I just merged https://github.com/ocaml/ocaml.org/pull/2406 which already contained a matrix inversion recipe using the owl library.

Thus, closing this one. But would be happy to review future PRs much quicker.

PR #2666, opened 2024-08-25
Update Academic Institutions Data and Extend Data Model

### LINKED ISSUE #2328: (Data) Extend the Data Model of Academic Institution to Record Information about Course Materials
In the [Wireframes for the improved "Teaching and Research" page](https://www.figma.com/file/eLNSdvayxqvvfBsRsdbJXN/OCaml-Home-Page?type=design&node-id=5%3A2500&mode=design&t=I6cJuQypSS6xjBag-1) (formerly "Academic Users"), we list course materials from lectures.

To make this truly useful, we need to improve the data model in [`academic_institution.ml`](https://github.com/ocaml/ocaml.org/blob/main/tool/ood-gen/lib/academic_institution.ml):
1. Add a field `year : int option` to `course`, so that we can sort the courses by most recent.
2. Add a field `lecture_notes_available: bool option` to represent whether lecture notes can be found at the URL pointed to by the field `online_resource`
3. Similarly, add a field `exercises_available: exercise_availability option` where `type exercise_availability = Yes | Interactive` (where `Yes` maps to the string `"yes"`, and `Interactive` maps to the string `"interactive"`) to represent whether there are exercises provided at the URL pointed to by the field `online_resource`.

To make this data available in the templates, you need to change [`data.mli`](https://github.com/ocaml/ocaml.org/blob/main/src/ocamlorg_data/data.mli) and render the new fields to the generated module in `academic_institution.ml`.

The task here is only to change the data model and maybe update one data entry (as to keep the PR small).

sabine:
Resolves #2328.


data/academic_institutions/pennsylvania.md
```diff
@@ -7,17 +7,22 @@ logo: academic_institution/penn.png
...
-      professor: Steve Zdancewic
+      teacher: Steve Zdancewic
       last_check: "2024-03-06"
       acronym: CIS 3410
       enrollment: "70"
-      online_resource: "https://www.seas.upenn.edu/~cis3410/current/"
+      url: "https://www.seas.upenn.edu/~cis3410/current/"
+      exercises: true
+      description: |
+        The provided homework projects of this compilers class progress from basic OCaml to building an X86lite simulator and an LLVMlite backend. Later assignments focus on developing and extending the Oat language compiler, as well as program analysis and optimizations.
```
christinerose: I'm not sure what to do with these @sabine. Words like "optimization" has an "s" in UK spelling, but this is a US class. My opinion is to stay consistent with the UK spelling in descriptions, but if Optimization is in the title of the course, keep the US spelling. What do you think?

christinerose (commented):
Some suggestions for title case as well as a question about UK spelling in places.

data/academic_institutions/pennsylvania.md
sabine: makes sense to me

christinerose (approved):

Issue #2015, opened 2024-01-31
Ready-to-paste install & dune dependency in package overview should not use version numbers


cuihtlauac:
In the general case, one wants to do:
```shell
$ opam install foo
```
Not 
```
$ opan install foo.0.9.9
```

The latter isn't _nominal_ it is exceptional. The same reasoning applies to Dune dependency. One lists a package name without a version. Specifying a version number is exceptional, specifying a fixed version number (with an equal sign) is even more.

Version numbers should be removed from both texts.

sabine:
It makes sense to remove the version on the latest version URL.

I think if people are on a specific version we can assume they probably have a reason for why they aren't on the latest version URL. :thinking:

The-Amoghavarsha:
So what's the middle ground here? Should we just remove the version number for the latest one and also let users use a specific version number (if they have a specific reason)?

sabine:
Yeah, we should change it on the latest version URL only to have no version in the clipboard-copy boxes.

The-Amoghavarsha:
Where to start from here? Need some ideas.

sabine:
When on the latest version URL, have the copy to clipboard snippets without a version. On pages with a version URL, keep things as they are.

shonfeder:
Fixing this for the `Dune Dependency` field will reduce friction to publishing packages, since the current behavior will lead newer contributors to put fixed bounds on deps that will then need correction before publishing to opam.

sabine:
Resolved by #2668

PR #2668, opened 2024-08-26
update install and dune dependency in package overview

akankshat05:
For #2015 
cc @sabine

### DIFF
--- a/src/ocamlorg_frontend/pages/package_overview.eml
+++ b/src/ocamlorg_frontend/pages/package_overview.eml
@@ -54,14 +54,14 @@ in
       Install
   </h2>
 
-  <%s! Copy_to_clipboard.small_code_snippet ~id:"install-package" ~class_:"whitespace-nowrap" ("opam install " ^ package.name) %>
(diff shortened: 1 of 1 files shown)


src/ocamlorg_frontend/pages/package_overview.eml
```diff
@@ -54,14 +54,14 @@ in
       Install
   </h2>
 
-  <%s! Copy_to_clipboard.small_code_snippet ~id:"install-package" ~class_:"whitespace-nowrap" ("opam install " ^ package.name ^ "." ^ specific_version) %>
+  <%s! Copy_to_clipboard.small_code_snippet ~id:"install-package" ~class_:"whitespace-nowrap" ("opam install " ^ package.name) %>
```
sabine: removing the version entirely only makes sense when we are on the latest version of the package, i.e. on any URL like this: `https://ocaml.org/p/PACKAGE_NAME/latest`.

Also, I think we don't need to remove the version from the `opam install` command because being precise here is actually helpful so people get an error message if they haven't updated their opam.

akankshat05: It should now be omitted for only the dune dependency for latest version of the package. Am i correct?

sabine: I don't know, there's no new commit on this branch yet.

sabine (approved):
updated to omit the version on the /latest URL

PR #2671, opened 2024-08-27
Cookbook recipe for matrix addition and multiplication

RCHG:
Hello!

I have prepared a recipe for the matrix addition and multiplication. In this case using the Array module of the standard library and implemented using integer elements for the arrays.

Note that it relies on vector addition and dot product of vectors to define the matrix operations in a kind of functional style (rather than using loops). Also in the process the recipe defines additional useful functions:
- Transpose of a matrix
- Vector addition
- Vector dot product 

It can be useful for those cases where Owl library can't be used, otherwise these operations of the matrices are defined in (a more performant way) in Owl. 

This example use integer vector and matrix, but I am happy to implement for floats if preferred. 

Thanks and kind regards,
Ramiro.

### DIFF
--- a/data/cookbook/matrix-addition-multiplication/01-array.ml
+++ b/data/cookbook/matrix-addition-multiplication/01-array.ml
@@ -1,0 +1,65 @@
+---
+packages:
+---
+
+(* Note: For production the Owl library has these matrix operations implemented in a more performant approach. *)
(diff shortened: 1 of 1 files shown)


RCHG:
Hello. I realised that some recipes have a kind of discussion at the beginning. I don't know exactly the syntax for that but the initial comments in the recipe are aimed to be a kind of discussion. I can work on it.

christinerose (approved):

RCHG:
This seems to be finished and ready, but their integration seems frozen since 29-Aug. Do I need to do something more? Thanks.

sabine:
Hi @RCHG, looking at this, I come to a similar conclusion as for https://github.com/ocaml/ocaml.org/pull/2540, which aimed to add computation of means and median:
(a) LLMs should be able to easily generate this kind of code when a developer wants to avoid using packages
(b) the cookbook aims to help people discover the good parts of the OCaml ecosystem by showcasing common tasks that occur in programming applications.

Thus, I would refrain from merging a standard library example for matrix operations and close this PR.

PR #2691, opened 2024-09-10
Adding Reason OCaml India to list of recurring meetups

ManasJayanth:


data/events/recurring.yml
```diff
@@ -74,3 +74,9 @@ recurring:
     country: "Sweden"
     city: "Stockholm"
     url: https://www.meetup.com/ReasonSTHLM/
+  - title: "Reason OCaml India meetups"
```
sabine:
```suggestion
  - title: "Reason OCaml India Meetup"
```

christinerose (approved):

ManasJayanth:
Thank you, for the review comments :)

Should we raise PRs here every week once the agenda is decided? I mean to ask, is this the expected usage of the `ocaml.org` platform for OCaml events?

sabine:
> Should we raise PRs here every week once the agenda is decided?

Would be happy to have those as early as you can make them!

In the long run we'd love to scrape the events so we'd automatically pick them up as they go live on your site, but right now we don't have such a setup, so it's individual PRs adding events.

PR #2693, opened 2024-09-11
Fix link

wpdevelopment11:
The link points to `discuss.ocaml.org`, but contains the text "Watch OCaml Videos" and uses PeerTube logo.


cuihtlauac:
Thanks, @wpdevelopment11; you are correct; this needed to be fixed.

PR #2701, opened 2024-09-13
Biome

smorimoto:
This PR introduces Biome to CI. This allows you to perform format and lint checks on several file types. Actually, it found some issues.

https://biomejs.dev

Remember to use hide white space mode to make the review easier.


cuihtlauac:
Thanks @smorimoto. This is interesting; however, I have mixed feelings about it.

I'm a strong supporter of automated checks. However, the proposed changes only include whitespace, line breaks, and optional semicolumns. Did I miss something? Tidying this way is acceptable, but having a tool to do that feels overkill. Do those files change often enough to use an automated tool?

Additionally, if we were to use such a tool, we should be able to run it locally, like we do with `ocamlformat`. However, if I'm correct, that would add Node.js to our list of development dependencies. Isn't it a lot for that kind of change?

smorimoto:
Biome does not require the Node.js ecosystem: https://biomejs.dev/guides/manual-installation

tmattio:
Without a strong reason to add this, I'd rather not add new steps (aka complexity and latency) to our CI.

smorimoto:
https://repology.org/project/direnv/versions

PR #2703, opened 2024-09-17
Fixed typo in zip archive example

ancantus:
Fixing a small typo I noticed while using the creating zip archive cookbook. The recursive call to the function `traverse_fs` was missing the `_fs`.


cuihtlauac (approved):

cuihtlauac (approved):
Thanks @ancantus. You are correct, this is a typo

PR #1004, opened 2023-03-22
remove top banner, but keep structure intact for future use

sabine:
* change banner color to dark blue, so the next person adding a banner doesn't have to think about that
* adjust `dark-blue` color to better resemble the dark blue background sections
* hide banner, but keep code around for future use

I imagine in a few months or weeks someone will want to add a banner again.

|before|after| after with banner|
|-|-|-|
|Screenshot 2023-03-22 at 17-53-20 Welcome to a World of OCaml|Screenshot 2023-03-22 at 17-52-51 Welcome to a World of OCaml|Screenshot 2023-03-22 at 17-53-24 Welcome to a World of OCaml|
|Screenshot 2023-03-22 at 17-53-30 Welcome to a World of OCaml|-|Screenshot 2023-03-22 at 17-53-33 Welcome to a World of OCaml|
|Screenshot 2023-03-22 at 17-53-41 Welcome to a World of OCaml alignment messed up|-|Screenshot 2023-03-22 at 17-53-44 Welcome to a World of OCaml|

PR #709, opened 2022-12-15
OCaml 5.0 release: announce and release page

Octachron:
This is the OCaml 5.0.0 announce slightly in advance of tomorrow release.

PR #2706, opened 2024-09-18
Introduce the dune developer preview

maiste:
Hi :wave: 
This PR introduces an explanation about the Dune Dev Preview we are putting in place to provide early access to Dune Package Management. It's still a draft as we need to finish some business on our side before publishing this and rewrite some part of it. I made it public so we can iterate over it.

@sabine, @cuihtlauac do you think it would be possible to add a header for a limited time to promote this? I can do it, but I want to have your green light on this before investing some time.


cuihtlauac:
Do you mean a banner like the one we had for OCaml 5.0.0? As in PR #709? (Also look at PR #1004)

maiste:
Yes exactly this! :pray:

cuihtlauac:
Why not? Dune is part of the OCaml platform. Let's find color and text.

maiste:
Thanks! I'll update the PR with the top bar, following the instruction in the provided PRs.

For the color, I would suggest #B0A687: it's the grey / cream color from the Dune logo :+1:

maiste:
Should put the link in `url.ml` instead of hardcoding it?

data/tool_pages/platform/2du_00_dune_developer_preview.md
```diff
@@ -0,0 +1,56 @@
...
+---
+
+## Context of the Dune Developer Preview
+
+Since 2012, [opam](https://opam.ocaml.org) has been the default system to
+manage package in the OCaml ecosystem. It allows interacting with a central
+package repository called `opam-repository`. It contains around 5000 packages of
+the OCaml ecosystem.
+
+However, with time, the OCaml Community has raised concerned about having a
```
samoht: I'd lead this with positive things about the new workflow instead of negative things about the existing one.

maiste: Would it sound better if I replace the "However, with time, [...] have raised concerned" by "With time, the OCaml Comunity have suggested to have [...]"? This way, the sentence does not contain anything negative.

My point here is just to emphasise that this is a community wanted feature.

cuihtlauac:
Here is how it looks like
Screenshot_2024-09-20_13-36-52
Screenshot_20240920-133858

maiste:
Some updates here. After discussing the subject with @leostera, we will only keep the banner as most of the information would be displayed on the `dune.ci.dev` (or `dune.ocaml.org`). The page will be moved to the website in the FAQ section.

maiste:
Closing in favour of #2766 (using the announcement in the changelog instead of the banner).

PR #2709, opened 2024-09-19
Fix opam switch activation instruction on Windows

### LINKED ISSUE #2707: The commands to activate the opam switch on windows are wrong
On the page "install ocaml on windows", the first steps are fine, but the two commands to activate the switch are wrong.
The one for cmd reads `for /f"tokens=*" %i in ('opam env') do @%i`, and it misses a space between `/f` and `"tokens=*"`.
The one for PoweShell reads `(& opam env) -split '?' | ForEach-Object { Invoke-Expression $_ }`,  but the `?` should be replaced with the endline code (maybe hard to put on a web page). This makes the installation confusing.

Another issue I had is that windows 10 appears to spend most of its running the antivirus. A comment about how to stop it would be welcom.

cuihtlauac:
Fix #2707 
* [X] Missing space (typo)
* [X] Endline code (escape sequence)
* [X] Antivirus (add "takes minutes" text)

### DIFF
--- a/data/tutorials/getting-started/1_00_install_OCaml.md
+++ b/data/tutorials/getting-started/1_00_install_OCaml.md
@@ -108,7 +108,7 @@ or from PowerShell:
 (& opam env) -split '\r?\n' | ForEach-Object { Invoke-Expression $_ }
 ```
 
-Opam is now installed and configured!
+Opam initialisation may take several minutes. While waiting its installation and configuration to complete, start reading [A Tour of OCaml](tour-of-ocaml).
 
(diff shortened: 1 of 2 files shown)


christinerose (approved):

cuihtlauac:
I've updated the eml code. I see `\r?\n` on both pages in 
* [X] Linux/Chrome
* [X] Linux/Firefox
* [X] Android/Chrome
* [X] iOS/Safari
* [X] Windows 11/Edge

On ChromeOS, I can't display `/install#windows`. That's an unrelated issue.

dra27 (approved):
LGTM, thank you!

christinerose (approved):

PR #2712, opened 2024-09-20
Update 1_02_your_first_ocaml_program.md

Durbatuluk1701:
I could be incorrect, but I believe that `dune proj init` should instead be `dune init proj`


cuihtlauac (approved):
Hantanyë @Durbatuluk1701.

PR #2713, opened 2024-09-21
Cookbook: added matrix multiplication and addition

masonhgn:
I added a recipe for the task "matrix addition and multiplication"


christinerose (approved):

RCHG:
Hello! There is another pull request for this recipe with identical dependencies since August (it is explained also [here](https://rchg.srht.site/articles/computing/2024-08/array_cookbook-en.html)), but not yet integrated in the webpage ocaml.org. Maybe this one is better or the idea is to merge/publish both? Thanks!

sabine:
Hi @masonhgn, looking at this, I come to a similar conclusion as for https://github.com/ocaml/ocaml.org/pull/2540, which aimed to add computation of means and median:
(a) LLMs should be able to easily generate this kind of code when a developer wants to avoid using packages
(b) the cookbook aims to help people discover the good parts of the OCaml ecosystem by showcasing common tasks that occur in programming applications.

Thus, I would refrain from merging a standard library example for matrix operations and close this PR.

PR #2718, opened 2024-09-26
Download and install tailwindcss binary without opam

cuihtlauac:
- Remove tailwindcss as a dependence in `dune-project` and `ocamlorg.opam`
- Remove Opam package pin in `dune-workspace` and `ocamlorg.opam.template`
- Add `tools/tailwindcss/dune` that
  - Download `tailwindcss` binary from `tailwindlabs/tailwindcss`
  - Installs it in local switch

Remark: files `tools/tailwindcss/URL` and `tools/tailwindcss/RELEASE` must not have end of line characters.


tmattio:
Generally in favor, but can you expand on the benefits you see here?
Two comments:
- I believe we now have an action in Dune to download things, so we can avoid using curl, which won't work on Windows
- We should only download the binary that's relevant to the user's system, the PR currently downloads all of the binaries.

cuihtlauac:
> Generally in favor, but can you expand on the benefits you see here?

One potential benefit is outside ocaml.org, not storing binaries in https://github.com/tmattio/opam-tailwindcss. @MisterDA shared his concern about it. 

But mostly, this is proof-of-concept code, I needed it whilst developing branch `bb` to make it possible to downgrade to `ocaml.4.14` and `dune.3.9`.

> I believe we now have an action in Dune to download things, so we can avoid using curl, which won't work on Windows

I wasn't aware this existed. You're right; it's better.

> We should only download the binary that's relevant to the user's system, the PR currently downloads all of the binaries.

~~Is this what happens? Rules with `action (run curl` are triggered by rules with `action (copy`, which have `enabled_if` stanzas made to only trigger a single download per system/hardware.~~

Update: Fixed

MisterDA:
> I believe we now have an action in Dune to download things, so we can avoid using curl, which won't work on Windows

There's a curl shipped by Microsoft since Windows 10.
https://curl.se/windows/microsoft.html

cuihtlauac:
BTW, has anybody ever tried to compile ocaml.org in Windows? This PR's `dune` stanzas do not allow to have a `tailwindcss` binary for Windows.

samoht:
Could these Dune rules be packaged in the `tailwindcss` package? I know a few other projects that could re-use this (including [mirage](https://github.com/mirage/mirage-www/blob/master/mirageio.opam.template#L6)).

cuihtlauac:
There's already an unused `dune` file in https://github.com/tmattio/opam-tailwindcss. It could be patched to include those rules (or look like ones). The `.opam` file could call dune in `install:` (or elsewhere). But wouldn't it be a useless level of indirection? opam calling dune calling curl instead of dune calling curl. I'd rather vendor a project with a single directory such as `tool/tailwindcss` that only contains the `dune` file I need to download the binaries.

Leonidas-from-XIV:
> I believe we now have an action in Dune to download things, so we can avoid using curl, which won't work on Windows

Dune has support for fetching files but it is not part of the action language and under the hood it also uses and requires `curl`.

tmattio:
Ok, I'm convinced with the approach, thanks for the PR @cuihtlauac! I've pushed a commit to simplify things slightly (I don't think we need to put the URL and version in separate files), but I'll merge now.

For some reason Dune package management was failing when installing opam-tailwindcss (cc @Leonidas-from-XIV might be an unknown issue?), so this brings back support for dune pkg.

cuihtlauac:
Happy to have convinced you. Did you intentionally or accidentally remove the `default` alias from `tool/tailwind/dune`? It was meant to prevent all four downloads from taking place.

tmattio:
It was intentional, I commented on the other PR: I'm not sure how the alias helps to download only one binary. Can you expand on this?

PR #2729, opened 2024-09-29
Update 1_01_a_tour_of_ocaml.md

221V:
fix lost ;; for newbie not be lost on the first command in repl ..


cuihtlauac (approved):
Thanks @221V, you're right, we've missed that one.

PR #2756, opened 2024-10-08
Serve V2 assets from filesystem

### LINKED ISSUE #2754: Existing links to meeting resources no longer resolve.
Looking through an old thread here https://discuss.ocaml.org/t/idea-standard-ocaml-runtime-type-representation/12051 I noticed the following links no longer resolve correctly on the new site. They get into a redirect loop and eventually give up with too many redirects error. It would be appreciated if existing links could be made to resolve again. Thank you.

https://v2.ocaml.org/meetings/ocaml/2013/proposals/runtime-types.pdf
https://v2.ocaml.org/meetings/ocaml/2013/slides/henry.pdf

cuihtlauac:
Fix #2754


christinerose (approved):

tmattio:
Why do we put the data in a special v2 directory and not with the other conferences?

sabine:
> Why do we put the data in a special v2 directory and not with the other conferences?

Indeed, if we are hosting files for conference materials, then we should just host them, this has nothing to do with v2 anymore after the migration.

PR #2758, opened 2024-10-09
Fix #2583 - Auto detect os on Install page & persist selected os

### LINKED ISSUE #2583: Link to specific platform installation instructions

piotr-m-jurek:
Fixes #2583.
- when user enters link `/install` without hash, gets redirected to detected os specific instructions
- when user enters link with hashed os name, stays on the hashed page, instead of switching to os

### DIFF
--- a/src/ocamlorg_frontend/pages/install.eml
+++ b/src/ocamlorg_frontend/pages/install.eml
@@ -5,14 +5,18 @@ Layout.render
 ~description:"Quick setup instructions to install OCaml on your system."
 ~canonical:"" @@
 <div class="lg:py-8 py-6">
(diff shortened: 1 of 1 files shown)


sabine (approved):
Thanks @piotr-m-jurek, this improves the behavior of linking to specific install instructions!

Merging when CI finishes. :sparkles:

PR #2760, opened 2024-10-09
Fixup constraints

Leonidas-from-XIV:
While updating an older OPAM switch I realized that the versions that are installed are too old to successfully compile, so I had to add a few constraints on versions and add dependencies that were not declared before.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -38,6 +38,7 @@
   bos
   crunch
   mirage-kv-mem
+  (mirage-crypto-rng (>= 0.11.0))
   dream-accept
   dream-encoding
   (graphql
--- a/dune-project
+++ b/dune-project
@@ -70,7 +70,7 @@
   ppx_stable
   ezjsonm
   lambdasoup
-  ptime
+  (ptime (>= 1.1.0))
   (cmdliner
    (>= 1.1.0))
   xmlm
--- a/dune-project
+++ b/dune-project
@@ -66,7 +66,7 @@
     (>= 1.10.0)))
   olinkcheck
   ; tools/ood-gen
-  ppx_deriving_yaml
+  (ppx_deriving_yaml (>= 0.3))
   ppx_stable
   ezjsonm
   lambdasoup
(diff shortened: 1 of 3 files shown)


sabine:
Since we're managing dependency versions via a pin on opam-repository, can we close this now? @cuihtlauac

cuihtlauac:
I'm tempted to merge this one. I find it nice to have a .opam file that allows building without a switch.

sabine:
> nice to have a .opam file that allows building without a switch

this is not enabled by this patch, since it's all just lower bounds and the upper bound is provided by the opam-repository pin

cuihtlauac:
@Leonidas-from-XIV told me he successfully built without creating a local switch and without a repo pin, only adding this PR

sabine:
I think this is trying to solve a problem that shouldn't exist in the first place, since everyone should be developing against the set of dependencies we're deploying with - in order to avoid wasting time debugging inconsistencies or bugs with dependencies in different people's switches

Leonidas-from-XIV:
> I think this is trying to solve a problem that shouldn't exist in the first place, since everyone should be developing against the set of dependencies we're deploying with

I still think that the dependencies are wrongly described in that case. If the versions are fixed then the dependencies should either be removed from the `dune-project` or fixed with `(= version)` constraints, but the state of having dependencies that can't build the project is not good.

Futhermore, pinning the opam-repository only fixes the upper bound but this PR fixes the lower bound, so the problem I ran into is still possible even when using a pin.

cuihtlauac:
I’ve been discussing this matter with @mtelvers and @shonfeder. Thanks to them, I’m now convinced we should move away from the repo pin and pin the docker image instead. That will enable caching opam switch building stage, which should speed up the ocaml-ci builds.

When building in Docker, either with pinned repo or pinned image, we don’t need any versionning information, what’s under the pin's cap does the job.

What’s needed is a way to have the same switch as in Docker when building outside Docker, for day-to-day development. Maybe, a GitHub action could derive an Opam switch dump when there's a pin change and create a PR out of that. Something automatic would be handy. Maintaining consistency on this is a weak spot in our set up.

sabine:
Obsolete since the issue has been resolved by upgrading the opam pin.

PR #2761, opened 2024-10-09
`dune` file cleanups

Leonidas-from-XIV:
I was looking for some improvements to the `dune` files to make them better and at the moment I have found

1. `workspace_root` vs `project_root` confusion. A lot of paths use `workspace_root` where they probably should point to the location relative to ocaml.org's `dune-project`, not where the workspace lies. `dream_eml` has a `--workspace` argument, I suspect it is thus smart enough to deal with workspaces properly.
2. Wrapping makes no difference to libraries whose names match single-file modules, thus the `wrapped` field can be removed.
3. Probably a lot of `ocamlorg_frontend/dune` could be auto-generated, but I haven't done this yet since I'd like to know if people are interested in that.

### DIFF
--- a/dune
+++ b/dune
@@ -3,15 +3,15 @@
  (rule
   (target main.css)
   (deps
-   %{workspace_root}/tool/tailwind/tailwindcss
-   (:config %{workspace_root}/tailwind.config.js)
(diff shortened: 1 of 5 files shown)


sabine (approved):
I rebased on main and I'm merging this now.

PR #2764, opened 2024-10-10
Generate `dream_eml` rules automatically

Leonidas-from-XIV:
Instead of having all the rules defined by hand, this uses a generator to write the rules to an include file.


src/rulegen/dune
cuihtlauac, line 1: Should probably go in `tools`.

src/rulegen/rulegen.ml
cuihtlauac, line 1: Should probably go in `tools`.

PR #2765, opened 2024-10-10
Only download a single tailwindcss binary

cuihtlauac:
Only download a single tailwindcss binary

Bring back stuff removed when merging #2718

### DIFF
--- a/dune
+++ b/dune
@@ -10,7 +10,15 @@
   (action
    (chdir
     %{workspace_root}
-    (run %{workspace_root}/tool/tailwind/tailwindcss -m -c %{config} -i %{input} -o %{target})))))
+    (run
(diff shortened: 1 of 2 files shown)


tmattio:
I am not sure how adding a default alias tells dune to download a single binary?
The alias wasn't used anywhere, we're using the binary directly.

cuihtlauac:
If `default` is not restricted to `taildwindcss`, it also contains the 4 platform specific binaries. Without this restriction, just doing `make clean` and `make build` downloads the 4 binaries.

Incidentally, having the binary copied to the local switch allowed it to survive `dune clean`

tmattio:
That's a behaviour of Dune I didn't know about, thanks!

PR #2767, opened 2024-10-10
Update the OCaml Changelog guidelines

tmattio:
cc @shonfeder - do you find the updated guidelines clearer? Would you add or change anything?


sabine (approved):

shonfeder (commented):
This is great. Thanks!

One other idea: if there is an exemplary post that you think gives newer authors a good indicator how the desired tone/length/content etc., you could reference that here.

PR #2770, opened 2024-10-13
Resolves  #2739 Use "/" to focus search input in the Header

### LINKED ISSUE #2739: '/' Shortcut for search
'/' is a common shortcut for accessing search bars. For instance, youtube and github use it. I think adding it to ocaml.org would be quite useful.

piotr-m-jurek:
So I didn't know what would be the best way to structure the code and what other things there are to consider, but I think I can handle the topic, so here goes nothing :)

Resolves #2739


sabine (approved):
Hey @piotr-m-jurek, thanks for this, I'm merging! :sparkles:

piotr-m-jurek:
Thanks @sabine, happy to contribute :)

EmileTrotignon:
Thanks a lot @piotr-m-jurek and @sabine

PR #2779, opened 2024-10-17
Weekly Reason OCaml India virtual event: OCaml with Buck2

ManasJayanth:
Event details: https://reason-ocaml.in/upcoming-meetups/


ManasJayanth:
Updated the PR to add October 27 meetup too, since the speaker has confirmed.

PR #2781, opened 2024-10-17
Use a Dune dialect to preprocess eml templates

aantron:
This spiritually resurrects #1181 by @gpetiot which stalled, as I understand it, due to lack of a Dream release that would make dialect support ergonomic. Because of the Dream release, the present PR doesn't need a wrapper around `dream_eml`.

It should also make #2764 unnecessary.


cuihtlauac:
Thanks @aantron 

In #1181, @emillon made this comment:
> Sidenote, I had a look at the diff in this PR. If I understand correctly you're replacing N rules with each 1 dependency and 1 target, by a single rule that has N dependencies and N targets. 

What's the status now? Are all the `.eml` going to be processed if a single one changes?

sabine:
Note: This makes https://github.com/ocaml/ocaml.org/pull/2764 obsolete (which is good IMO).

aantron:
AFAIK that comment referred to an approach that was being used in that PR (https://github.com/ocaml/ocaml.org/pull/1181/commits/12c9134ce9c03f1c7f01c3b5be9195ad87fa6a0a#diff-85098e336ea5f1f8dca91cda539ff5f46db068c37b959eb70a32d02bc591941fL6) that had all the template outputs depend on all the template inputs, before @gpetiot switched to using a Dune dialect in the commit from that PR that I am linking in this comment. I also had that concern for a moment before I realized that, but AFAIK there is no reason to suspect that Dune dialects have this downside.

sabine (approved):
LGTM thanks @aantron!

PR #2783, opened 2024-10-22
remove hyperlinks to nqsb dot io

hannesm:
The reason is that the domain is no longer providing the content we wanted to link to. It is now providing explicit content (pornography).

I thought ocaml.org planet would pick up from the updates of mirage.io feed etc.  -- but it looks like it doesn't.


hannesm:
anything I can do to push this PR forward?

cuihtlauac:
Sorry for the delay @hannesm

hannesm:
Thanks a lot @cuihtlauac :)

PR #2789, opened 2024-10-26
Cookbook recipes for the Cryptography section

xavierleroy:
Using the Cryptokit library.  Hope the amount of comments is about right.


sabine (approved):
Thanks @xavierleroy, much appreciated! The elegant and readable code made this very easy to review and edit.

PR #2818, opened 2024-11-18
correct uninstall script URL

BBboy01:


christinerose (approved):

BBboy01:
kind of confused, If uninstalled brew, in reinstall setp, which brew we should use

PR #2831, opened 2024-11-25
Update the CI and deployment to OCaml 5.2.1, Alpine 2.21, opam 2.3, and use BuildKit syntax in the Dockerfile

MisterDA:
Update the CI and deployment from OCaml 5.2.0 to 5.2.1. Update the deployment image to Alpine 2.21, and opam 2.3 (should be faster).

Updating to the BuildKit syntax should speed up builds a bit, by (among other changes) allowing caching of apk and opam packages. Enabling BuildKit in the deployer also allows parallelism in building the image stages, where possible.
- [x] https://github.com/ocurrent/ocurrent-deployer/pull/248

PR #2861, opened 2024-12-12
Dockerfile cleanup

cuihtlauac:
Extracted from PR #2831

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,32 +1,41 @@
-FROM ocaml/opam:alpine-3.20-ocaml-5.2 as build
+FROM ocaml/opam:alpine-3.20-ocaml-5.2 AS build
 
 # Install system dependencies
-RUN sudo apk update && sudo apk add --update libev-dev openssl-dev gmp-dev oniguruma-dev inotify-tools curl-dev autoconf
+RUN apk -U upgrade --no-cache && apk add --no-cache \
+    autoconf \
+    curl-dev \
+    gmp-dev \
+    inotify-tools \
+    libev-dev \
+    oniguruma-dev \
+    openssl-dev
 
 # Branch freeze was opam-repo HEAD at the time of commit
-RUN cd opam-repository && git pull origin c45f5bab71d3589f41f9603daca5acad14df0ab0 && opam update
+RUN cd ~/opam-repository && git fetch -q origin master && git reset --hard c45f5bab71d3589f41f9603daca5acad14df0ab0 && opam update
 
 WORKDIR /home/opam
 
-# Install Opam dependencies
-ADD ocamlorg.opam ocamlorg.opam
+# Install opam dependencies
+COPY --chown=opam --link ocamlorg.opam .
 RUN opam install . --deps-only
 
 # Build project
-COPY --chown=opam:opam . .
+COPY --chown=opam . .
 RUN opam exec -- dune build @install --profile=release
 
 # Launch project in order to generate the package state cache
-RUN cd opam-repository && git checkout master && git pull origin master && opam update
-ENV OCAMLORG_REPO_PATH opam-repository
-ENV OCAMLORG_PKG_STATE_PATH package.state
+RUN cd ~/opam-repository && git checkout master && opam update
+ENV OCAMLORG_PKG_STATE_PATH=package.state \
+    OCAMLORG_REPO_PATH=opam-repository
 RUN touch package.state && ./init-cache package.state
 
-FROM alpine:3.20 as run
 
-RUN apk update && apk add --update libev gmp git
+FROM alpine:3.20
 
-RUN chmod -R 755 /var
+RUN apk -U upgrade --no-cache && apk add --no-cache \
+    git \
+    gmp \
+    libev
 
 COPY --from=build /home/opam/package.state /var/package.state
 COPY --from=build /home/opam/opam-repository /var/opam-repository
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -1,32 +1,41 @@
...
 # Branch freeze was opam-repo HEAD at the time of commit
-RUN cd opam-repository && git pull origin c45f5bab71d3589f41f9603daca5acad14df0ab0 && opam update
+RUN cd ~/opam-repository && git fetch -q origin master && git reset --hard c45f5bab71d3589f41f9603daca5acad14df0ab0 && opam update
 
 WORKDIR /home/opam
 
-# Install Opam dependencies
-ADD ocamlorg.opam ocamlorg.opam
+# Install opam dependencies
+COPY --chown=opam --link ocamlorg.opam .
```
cuihtlauac: Can we keep the old thing here?
```suggest
ADD ocamlorg.opam ocamlorg.opam
```

Dockerfile
```diff
@@ -1,32 +1,41 @@
...
+    curl-dev \
+    gmp-dev \
+    inotify-tools \
+    libev-dev \
+    oniguruma-dev \
+    openssl-dev
 
 # Branch freeze was opam-repo HEAD at the time of commit
-RUN cd opam-repository && git pull origin c45f5bab71d3589f41f9603daca5acad14df0ab0 && opam update
+RUN cd ~/opam-repository && git fetch -q origin master && git reset --hard c45f5bab71d3589f41f9603daca5acad14df0ab0 && opam update
```
cuihtlauac, line 14: This has always confused me. I think we've changed this line a couple of times. Back and forth between what we have now and what you're suggesting. Your line downloads some more doesn't it? Is there another difference?

Dockerfile
```diff
@@ -1,32 +1,41 @@
...
 # Build project
-COPY --chown=opam:opam . .
+COPY --chown=opam . .
 RUN opam exec -- dune build @install --profile=release
 
 # Launch project in order to generate the package state cache
-RUN cd opam-repository && git checkout master && git pull origin master && opam update
-ENV OCAMLORG_REPO_PATH opam-repository
-ENV OCAMLORG_PKG_STATE_PATH package.state
+RUN cd ~/opam-repository && git checkout master && opam update
```
cuihtlauac, line 27: Agreed, if you pulled in the previous stage, you don't need to do it again.

cuihtlauac:
@MisterDA can you have a look at this?

Dockerfile
cuihtlauac: In my view, here, to avoid docker cache miss, we should just do `git checkout` to a hash that is known to be already in opam-repo history.

Dockerfile
MisterDA: Ha, my mistake, I shouldn't have introduced the `--link` here, but `COPY` is preferable to `ADD`: [Dockerfile best practice: `ADD` or `COPY`](https://docs.docker.com/build/building/best-practices/#add-or-copy).

Dockerfile
MisterDA: I don't think it downloads more, as per the docs:

> More precisely, `git pull` runs `git fetch` with the given parameters and then depending on configuration options or command line flags, will call either `git rebase` or `git merge` to reconcile diverging branches.

So here by splitting `git pull` into `git fetch` and `git reset`, we're avoiding conflict resolution, though it's not clear to me why there would be any conflicts in the first place. Corruption could happen.

The base image already ships the opam repository, I tend to set this hash to the one already present in the image, so that the `git reset` operation is mostly free. It can be needed from time to time to update to a newer opam-repository, or just set it explicitly for reproducibility/debugging purposes. The base images are updated weekly, I don't know how often the deployed image is.

cuihtlauac: The deployed image is updated fairly frequently. Several times a week, often more. If we can take advantage of this, it would be nice. That's why we are considering using a docker image pin instead of an opam repo pin

MisterDA: Both may be needed, as the docker image is updated weekly, you might want to rely on a dependency that's been released during the week but not yet available in the base image.

Dockerfile
MisterDA:
```suggestion
COPY --chown=opam ocamlorg.opam .
```

Dockerfile
cuihtlauac: No we don't want to do that :-) We're laggards. We want to use old stuff that has been around for some time, preferably a long time. We run away from anything new, fresh, fancy or cool :-)

MisterDA: There might be new, fresh, fancy, or cool bug fixes!

Dockerfile
```diff
@@ -1,32 +1,41 @@
-FROM ocaml/opam:alpine-3.20-ocaml-5.2 as build
+FROM ocaml/opam:alpine-3.20-ocaml-5.2 AS build
 
 # Install system dependencies
-RUN sudo apk update && sudo apk add --update libev-dev openssl-dev gmp-dev oniguruma-dev inotify-tools curl-dev autoconf
+RUN apk -U upgrade --no-cache && apk add --no-cache \
```
MisterDA:
```suggestion
RUN sudo apk -U upgrade --no-cache && sudo apk add --no-cache \
```

cuihtlauac:
Terrible mistake I shouldn't have merged this!

PR #2879, opened 2024-12-19
feat: allow filtering of OCaml Planet blogs by 'ocaml' keyword

sabine:


yawaramin:
Does this mean that if I add my blog's feed https://dev.to/feed/yawaramin this will find and ingest only posts tagged `ocaml`? Ie where `rss > channel > item > category` contains `ocaml`?

cuihtlauac:
We can make a try if you want. That would be the easiest. Just do 'make scrape', it should create a directory with your posts. You'll see what's in it. If you're not set up to build ocamlorg I can do the test for you

yawaramin:
Oh I just read the PR and saw that it just scans the post title and content for the word `ocaml`, that should work fine too.

PR #2882, opened 2024-12-19
Strengthen feed filtering freshness

cuihtlauac:
No longer look at the dates. Just make sure the list aren't too long.


sabine (approved):

sabine:
Probably a better way to deal with the problem of feeds growing too large, indeed.

PR #2909, opened 2025-01-07
Remove section about installing and using "tree" tool from get started tutorial

danielstocks:
Hello!

I recently ran through the get started and language guides on ocaml.org as a total beginner and they're very good, so thank you for that!

I have a small suggestion for an improvement: While going through [Your First OCaml Program](https://ocaml.org/docs/your-first-program#compiling-ocaml-programs) I felt that the _Compiling OCaml Programs_ section goes on a bit of a unnecessary detour on installing `tree` on MacOs with Homebrew, and was a bit hard to read due to all extra disclaimers (4) needed in case of issues with installing those tools

Don't get me wrong, I love `tree` and I use it myself, but I simply suggest removing this section because I don't think it adds value to the core topic of the tutorial itself. There's nothing wrong with showing the directory structure created by dune init though, so I suggest keeping that, but only for illustrative purposes.

In this PR you can se my suggested edits, please feel free to close this if you disagree!

Thanks again for all the great work!


data/tutorials/getting-started/1_02_your_first_ocaml_program.md
```diff
@@ -78,10 +72,10 @@ $ tree
 └── test
     ├── dune
     └── hello.ml
-
-4 directories, 8 files
 ```
 
+Unlike in Unix where they contain compiled binaries, directories `lib` and `bin` contain source code files, for libraries and programs, respectively. This is the a convention used in many OCaml projects, including those created by Dune. All the built artifacts, and a copy of the sources, are stored in the `_build` directory. You shall not edit anything in the `_build` directory.
```
sabine:
```suggestion
Unlike in Unix where they contain compiled binaries, directories `lib` and `bin` contain source code files, for libraries and programs, respectively. This is the a convention used in many OCaml projects, including those created by Dune. All the built artifacts, and a copy of the sources, are stored in the `_build` directory. Do not edit anything in the `_build` directory, since any manual edits will be overwritten during subsequent builds.
```

sabine (approved):
@danielstocks Love this because, indeed, it's assuming people are on MacOS which is not necessarily the case, and we should be safely able to assume that people will know how to inspect directories in their OS of choice!

yawaramin (approved):

data/tutorials/getting-started/1_02_your_first_ocaml_program.md
```diff
@@ -57,15 +57,9 @@ Success: initialized project component named hello
...
+The project is stored in a directory named `hello` with the following contents:
 
-**Note 3**: If you get an error in Homebrew from this in an Apple silicon macOS, it's likely an issue with the architecture switch from Intel to ARM. Please refer to the [ARM64 Fix](/docs/arm64-fix) to remedy the ARM64 error.
-
-**Note 4**: Unlike in Unix where they contain compiled binaries, directories `lib` and `bin` contain source code files, for libraries and programs, respectively. This is the a convention used in many OCaml projects, and the canned project created by Dune. All the built artifacts, and a copy of the sources, are stored in the `_build` directory. You shall not edit anything in the `_build` directory.
 ```shell
-$ cd hello
-$ tree
-.
 ├── bin
```
cuihtlauac, line 64:
```suggestion
hello
├── bin
```

cuihtlauac (changes requested):
Thanks @danielstocks, this a nice improvement.

cuihtlauac (approved):

PR #2912, opened 2025-01-09
Fix issue 2911

### LINKED ISSUE #2911: The installing-ocaml page ought to end with how to get out of utop
The page at https://ocaml.org/docs/installing-ocaml currently ends with a tutorial that gets us into utop, and lets us run a simple query to prove it is working, and then leaves us hanging out in utop. The usual methods of quitting a shell aren't obvious (exit, quit, exit;; etc all fail). It would be good to leave the user in the state they started.

cuihtlauac:
Fix issue #2911


sabine (approved):
Makes sense, thanks!

PR #2917, opened 2025-01-09
Cookbook: Debug Print a Value with ppx_deriving

mostafatouny:
Hello,

I am contributing to [Debug Print a Value](https://ocaml.org/cookbook/debug-print-a-value).

Moreover,
- I hope someone shows how to use `ppx_deriving` with `utop -ppx`.
- I nominate @yawaramin to add his post, as he commented [here](https://discuss.ocaml.org/t/contributing-to-ocaml-recipes-debug-print-a-value/15898/19?u=mostafatouny).
- I hope someone add printing recipes for maps (association table), Sets, Hash-tables, and Sequences data structures.

### DIFF
--- a/data/cookbook/debug-print-a-value/00-ppx_deriving.ml
+++ b/data/cookbook/debug-print-a-value/00-ppx_deriving.ml
@@ -5,22 +5,7 @@ packages:
   used_libraries:
   - ppx_deriving
 discussion: |
-    To create a minimal project:
-    ```
-    mkdir printDebug
-    cd printDebug
-    printf "\
-    (executable
-     (name minimo)
-       (preprocess (pps ppx_deriving.show))
-       )" > ./dune
-    printf "(lang dune 3.17)" > ./dune-project
-    printf "let () = print_endline ([%%show: string] \"hello world\")" > ./minimo.ml
-    opam switch create ./
-    opam install dune ppx_deriving
-    ```
-    Check your dune version by `opam exec -- dune --version` and modify `dune-project` file if needed. Build and execute the project by `opam exec -- dune exec ./minimo.exe`.
-    For a background, see [minimum setup](https://ocaml.org/docs/your-first-program#minimum-setup), [opam switch intro](https://ocaml.org/docs/opam-switch-introduction), and [ppx](https://ocaml.org/docs/metaprogramming).
+    It is required to have `(preprocess (pps ppx_deriving.show))` in `dune` file.
 ---
 
 (* string *)
@@ -31,14 +16,14 @@ let () = print_endline ([%show: string] "hello world")
 let show_int = [%show: int]
 let () = print_endline (show_int 7)
 
-(* tuple *)
+(* tuple of integer and string *)
 let show_pair : (int * string) -> string = [%show: (int * string)]
 let () = print_endline (show_pair (3, "hello"))
 
-(* list of tuples *)
+(* list of tuples, each is a boolean and character *)
 let () = print_endline @@ [%show: (bool * char) list] [ (true, 'a'); (false, 'b') ]
 
-(* user-defined type *)
+(* user-defined type; binary tree with weighted vertices *)
 type tree =
     | Leaf of float
     | Node of float * tree * tree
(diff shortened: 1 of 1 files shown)


cuihtlauac (approved):
Thanks @mostafatouny

data/cookbook/debug-print-a-value/00-ppx_deriving.ml
cuihtlauac, line 1: Thanks @mostafatouny. I just wanted to inform you that your comments don't provide helpful information. This needs to be fixed.
image

The discussion block does not need project creation instruction. This is covered elsewhere.

cuihtlauac (changes requested):

mostafatouny:
Thank you @cuihtlauac and @sabine for your time. I removed the project initialization and added more info to the comments.

cuihtlauac:
Thanks for the quick update, @mostafatouny. That's nice. I'm under the impression this recipe has more to do with using `ppx_deriving.show` than debugging. The former is useful in the latter context, but that's the only relationship I see. I propose we remove all references to debugging and change the title to produce value literals using preprocessing (or something akin).

yawaramin:
@cuihtlauac there is a connection to debugging. People often want to debug by just printing the values. Then they have to figure out how to print the values. And they see that they need to derive printers for custom types. So we should definitely mention this workflow.

If we don't mention a motivating use case like debugging, it might not be clear to people _why_ they would want to use this recipe. The recipe should justify itself by providing a use case. It's not enough to just show how to do it.

cuihtlauac:
Thanks, @yawaramin. You have a valid point. The key is discovering that you need ppx `show` when debugging. However, `show` is also very useful in multi-staged compilation. Maybe this could be mentioned in the discussion.

data/cookbook/debug-print-a-value/00-ppx_deriving.ml
```diff
@@ -0,0 +1,68 @@
...
+  tested_version: "6.0.3"
+  used_libraries:
+  - ppx_deriving
+discussion: |
+    It is required to have `(preprocess (pps ppx_deriving.show))` in `dune` file.
+---
+
+(* string *)
+let () = print_endline "hello world"
+let () = print_endline ([%show: string] "hello world")
```
yawaramin: This seems redundant. The string literal is already a string.

data/cookbook/debug-print-a-value/00-ppx_deriving.ml
```diff
@@ -0,0 +1,68 @@
...
+
+(* integer *)
+let show_int = [%show: int]
```
yawaramin: This is slightly more useful but no clear advantage over `string_of_int` imho.

data/cookbook/debug-print-a-value/00-ppx_deriving.ml
```diff
@@ -0,0 +1,68 @@
...
+let () = print_endline (show_pair (3, "hello"))
+
+(* list of tuples, each is a boolean and character *)
+let () = print_endline @@ [%show: (bool * char) list] [ (true, 'a'); (false, 'b') ]
+
+(* user-defined type; binary tree with weighted vertices *)
+type tree =
+    | Leaf of float
+    | Node of float * tree * tree
+    [@@deriving show]
```
yawaramin, line 25: We might want to explain why we are suddenly switching from `[%show: ...]` to `[@@deriving show]`.

mostafatouny:
> The key is discovering that you need ppx show when debugging

I agree. If something is well-documented but end users cannot figure it is the solution to their problem, then it is useless.

>  However, show is also very useful in multi-staged compilation

We can create
1 - a cookbook guide about ppx_deriving.
2 - small example on Debug Print, citing guide (1).
3 - small example on multi-staged compilation, citing guide (1).

---

Thanks @yawaramin for the code feedback. I removed redundant `string` and `int`. As a beginner myself, I cannot explain why we need `@@deriving`.

data/cookbook/debug-print-a-value/00-ppx_deriving.ml
sabine: It helps to take inspiration from the JSON parsing recipes, to give a better explanation of what `@@deriving` does.

And indeed, makes sense to add a single sentence that explains that `[%show: t]` generates the code to print values of type `t`

mostafatouny: @sabine would you recommend me background resources to check? maybe I can write a short explanation and enhance the tutorial's accessibility.

cuihtlauac:
LGTM. Thanks @mostafatouny @yawaramin @sabine

cuihtlauac (approved):

PR #2930, opened 2025-01-20
Use ppx import in ood

cuihtlauac:
Move `@@deriving` attributes from `Data_intf` to ood/lib files

Module `Data_intf` does not need to expose Yaml parsing or
pp/show functions at runtime. This is only used when crunching
the data. Types are duplicated in ood using `ppx_import` and
 `ppx_deriving` is applied to those definitions. 

Staged preprocessing is used.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -70,6 +70,7 @@
   olinkcheck
   ; tools/ood-gen
   ppx_deriving_yaml
+  ppx_import
   ppx_stable
   ezjsonm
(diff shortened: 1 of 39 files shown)


cuihtlauac:
@mattam82: this PR might be of interest for rocq-prover.org 

Using ppx-import allows removing yaml processing matters from runtime, which reduces coupling.

PR #2944, opened 2025-01-29
Cookbook recipe to salt and hash passwords with Argon2

boechat107:
Since [OWASP recommends](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html) using [Argon2id](https://github.com/P-H-C/phc-winner-argon2), I thought this recipe could be useful.


cuihtlauac (approved):
Thanks @boechat107, that looks good to me

PR #2946, opened 2025-01-31
Add changelog entry for OCamlformat 0.27.0

Julow:


Julow:
That was fast, thanks!

PR #2947, opened 2025-02-02
Change `#exit;;` to `#quit;;`

onsah:
Seems like there was a minor mistake. `exit` is a function but it's used like an interpreter directive. It can be confusing to beginners since it doesn't work.


cuihtlauac:
Thanks @onsah

PR #2951, opened 2025-02-11
Fix missing newline in Arrays

emerikaji:
Fixes a missing newline after the Markdown list that specifies the arguments for the `Arrays.sort` function.


cuihtlauac (approved):
Thanks @emerikaji

PR #2964, opened 2025-02-18
Fix some typos

bbatsov:
I've noticed a typo in the release notes and then when playing with tools like `codespell` I saw there are quite a few typos in the docs overall. (I added a commit fixing the most common typo)

Let me know if you're interested in such small fixes. Perhaps we can even automate a bit the spellchecking with a GHA.


bbatsov:
The whitespaces fixes were applied automatically by my editor and I think they are all good.

cuihtlauac (approved):
Thanks @bbatsov

cuihtlauac:
As said previously, fixing spaces is ok, but we prefer whitespace-only changes as separated PRs.

Yes, spelling as a GHA would be nice; we use British English. Bear in mind that news and planet content are automatically scraped by Action. I'm not sure we should alter syndicated content.

bbatsov:
> Bear in mind that news and planet content are automatically scraped by Action. I'm not sure we should alter syndicated content.

Got it. I didn't notice those were pulled with a job. I guess it doesn't make much sense to fix typos there, but I guess it won't hurt either. I'm not sure if the articles are re-fetched or not to check for updates, but either way - they'll either get updated and override the typos change or not, and both options seem fine to me.

PR #2968, opened 2025-02-18
[Fix #2965] Remove duplicate page

### LINKED ISSUE #2965: Duplicated Editor content
We've got some editor setup pages in two places:

- https://ocaml.org/docs/configuring-your-editor (Platform)
- https://ocaml.org/docs/set-up-editor (Getting Started)

I noticed this when I was looking to update the editor instructions. I think we should probably
moving everything tooling relate to getting started, especially if the content is mostly the same.

Might also be nice to have different URL paths for the main sections of the docs, as now it seems a bit arbitrary what ends up in which part of the site. (from the perspective of the main nav bar, at least)

Anyways, let me know if you'd like me to fold those pages in one and maybe move the https://ocaml.org/docs/formatting-your-code page to getting started as well.

bbatsov:
The duplicate page was removed and a redirect was added for it. I'll move the page for ocamlformat to the "Tooling" section in a separate PR (#2969)


cuihtlauac (approved):
Thanks @bbatsov, that looks good to me

PR #2969, opened 2025-02-18
Move ocamlformat page under "Tooling"

bbatsov:
As mentioned in #2965, I think that's the right place for this page in the menu. The link's the same, so there's nothing else that needs changing.

### DIFF
--- a/data/tutorials/platform/1_07_ocamlformat.md
+++ b/data/tutorials/platform/1_07_ocamlformat.md
@@ -4,7 +4,7 @@ title: "Formatting Your Code With OCamlFormat"
 short_title: "Formatting Your Code"
 description: |
   How to set up OCamlFormat to automatically format your code
-category: "Editor Support"
+category: "Tooling"
 ---
 
 Automatic formatting with OCamlFormat requires an `.ocamlformat` configuration file at the root of the project.
--- a/data/tutorials/platform/1_07_ocamlformat.md
+++ b/data/tutorials/platform/1_07_ocamlformat.md
@@ -4,7 +4,7 @@ title: "Formatting Your Code With OCamlFormat"
 short_title: "Formatting Your Code"
 description: |
   How to set up OCamlFormat to automatically format your code
-category: "Tooling"
+category: "Additional Tooling"
 ---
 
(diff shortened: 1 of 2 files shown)


cuihtlauac (commented):
~~@bbatsov Why this change? I don't understand the motivation~~

bbatsov:
Just to expand a bit on the reasoning - `ocamlformat` seems more related to the editor tooling than to the OCaml Platform (which I see as the compiler, standard library, opam and potentially dune).

In general I didn't see a clear definition of what's considered part of the Platform, so that's just my interpretation based on what's considered Platform in other programming communities.

cuihtlauac:
Since #2968 moves “Configuring Your Editor” from “Platform” to “Get Started” it leaves the “Platform / Editor Support” section with only one item: “Formatting Your Code”. And this renames “Editor” into “Tooling”.

Please provide that kind of information when creating a PR, that helps review. 

I agree “Editor Support” isn't a great section title. I don't think “Tooling” is an improvement, it remains vague. That doesn't help indexing by search and LLM engines.

For the sake of consistency, I'd rather stick with old name, it's not great, but we prefer to change only if an issue is genuinely fixed.

bbatsov:
> I agree “Editor Support” isn't a great section title. I don't think “Tooling” is an improvement, it remains vague. That doesn't help indexing by search and LLM engines.

The rename is only to group this together with the other similar articles (it's not a preferences for "Tooling" over "Editor Support"). I think this will improve the discoverability of this section a lot. If we keep the name of this section it will stay apart from the other related articles.

cuihtlauac:
I've merged PR #2965. Can you rebase this PR? I must admit I'm still missing the point of this one.

bbatsov:
Most to consolidate the editor-related docs in the same section. Now we have this page shown on `Learn -> Platform -> Editor Support` and I think it's better to be shown on `Learn -> Getting Started -> Tooling`. Alternative I can suggest folding the single-item sections on the Platform page under `Projects`:

image

The odoc and ocamlformat pages make sense in this context. Overall my point is that high-level sections and their subsections are not organized in the most logical way and I've been trying to improve this a bit.

sabine (approved):
I'm moving the odoc page under "Additional Tooling" as well.

PR #2949, opened 2025-02-06
data scraper for Changelogs

Julow:
This adds a tool that detects missing changelogs by looking at projects' release pages. My intention is to write a GA that will notify maintainers (perhaps by opening draft PRs) for new releases.
Does that sound good ?

This will not help catching new releases in its current form but it can be used to add changelogs that are already missing. Here's its output:

<details>

```
Downloading https://github.com/ocaml/dune/releases.atom ... done 
Downloading https://github.com/tarides/dune-release/releases.atom ... done 
Downloading https://github.com/realworldocaml/mdx/releases.atom ... done 
No changelog entry for "mdx" version "2.5.0"

No changelog entry for "mdx" version "Release 0.3.3"

No changelog entry for "mdx" version "Release 0.3.2"

No changelog entry for "mdx" version "Release 0.3.1"

Downloading https://github.com/ocaml/merlin/releases.atom ... done 
No changelog entry for "merlin" version "5.4.1-503"

No changelog entry for "merlin" version "5.4-503"

No changelog entry for "merlin" version "4.18-414"

No changelog entry for "merlin" version "5.3-502"

No changelog entry for "merlin" version "4.17.1-414"

No changelog entry for "merlin" version "4.17.1-501"

No changelog entry for "merlin" version "5.2.1-502"

No changelog entry for "merlin" version "4.17-414"

No changelog entry for "merlin" version "4.17-501"

No changelog entry for "merlin" version "5.2-502"

Downloading https://github.com/ocaml/ocaml/releases.atom ... done 
No changelog entry for "ocaml" version "5.2.1-rc1"

Downloading https://github.com/ocaml/ocaml-lsp/releases.atom ... done 
No changelog entry for "ocaml-lsp" version "1.22.0"

No changelog entry for "ocaml-lsp" version "1.21.0"

No changelog entry for "ocaml-lsp" version "1.20.0"

Downloading https://github.com/ocaml-ppx/ocamlformat/releases.atom ... done 
Downloading https://github.com/OCamlPro/ocp-indent/releases.atom ... done 
No changelog entry for "ocp-indent" version "nlfork-1.5.5"

No changelog entry for "ocp-indent" version "1.8.2"

No changelog entry for "ocp-indent" version "nlfork-1.5.4"

No changelog entry for "ocp-indent" version "nlfork-1.5.3"

Downloading https://github.com/ocaml/odoc/releases.atom ... done 
No changelog entry for "odoc" version "3.0.0~beta1"

[...]

PR #2971, opened 2025-02-18
data: Remove changelogs for alpha, beta, rc releases

Julow:
Continuing the discussion from https://github.com/ocaml/ocaml.org/pull/2949#discussion_r1944496714


cuihtlauac:
Thanks, @Julow; I support this. Prereleases are for the developers and early adopters willing to experiment, mainstream should not mention it. @sabine what do you think of this?

As a side remark, I can't find text in `CONTRIBUTING.md` that defines a policy regarding release announcements. We can announce prereleases or not, but in any case, the policy should be explicit.

PR #2979, opened 2025-02-24
Update loops_and_recursion.md with minor updates for clarity

jakubsvec001:
These are minor tweaks.

1) remove comma
-any use at all, except in conjunction with our old friend *references*.
+any use at all except in conjunction with our old friend *references*.

2) make language slightly more formal
-I'll only talk about the most useful ones here.
+we'll restrict ourselves to the most useful ones in this chapter.

3)  update flow of sentence
-The `map` and `filter` functions operate on individual list elements in
-isolation. **Fold** is a more unusual operation that is best
+While `map` and `filter` functions operate on individual list elements in
+isolation, **Fold** is a more unusual operation that is best
 thought about as "inserting an operator between each element of the
 list." Suppose I wanted to add all the numbers in my list together. In
 hand-waving terms, I want to insert a plus (+) sign between the
@@ -275,7 +275,7 @@ this:

4) add link to documentation
-`really_input` method. This is the simplest, but it might not work on
+[`really_input`](https://ocaml.org/manual/5.3/api/Stdlib.html#VALreally_input) method. This is the simplest, but it might not 

5) make side-comment parenthetical
-2, but it's not so easy to understand, for imperative programmers at least:
+2, but it's not so easy to understand (for imperative programmers at least):

### DIFF
--- a/data/tutorials/language/0it_02_loops_and_recursion.md
+++ b/data/tutorials/language/0it_02_loops_and_recursion.md
@@ -89,7 +89,7 @@ programmers like recursion, so `while` loops are second-class citizens
 in OCaml.
 
 If you stop to consider `while` loops, you may see that they aren't really
-any use at all, except in conjunction with our old friend *references*.
+any use at all except in conjunction with our old friend *references*.
 Let's imagine that OCaml didn't have references for a moment:
 
 <!-- $MDX skip -->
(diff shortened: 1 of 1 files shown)


data/tutorials/language/0it_02_loops_and_recursion.md
```diff
@@ -275,7 +275,7 @@ this:
 
 ### Approach 1
 Get the length of the file and read it all at once using the
-`really_input` method. This is the simplest, but it might not work on
+[`really_input`](https://ocaml.org/manual/5.3/api/Stdlib.html#VALreally_input) method. This is the simplest, but it might not work on
```
cuihtlauac:
```suggestion
[`really_input`](https://ocaml.org/manual/api/Stdlib.html#VALreally_input) method. This is the simplest, but it might not work on
```

cuihtlauac (approved):

cuihtlauac:
Thanks @jakubsvec001. I tweaked a link.

PR #2981, opened 2025-02-24
Update imperative md

jakubsvec001:
These are minor tweaks.

1) add remark noting distinction between operators and language constructs (to complement a similar remark about the sequencing 'operator' later in the document)
+**Remark** the left arrow symbol `<-` for mutating mutable record field values is not an operator like the assignment operator `( := )` is for `refs`. It is rather a _construct_ of the language.

2) explicitly state that 'assign' is an operator:
-The error came from assign `:=`, which associates stronger than a semicolon `;`. Here is what we want to do, in order:
+The error came from the assign operator `:=`, which associates stronger than a semicolon `;`. Here is what we want to do, in order:

3) Follow direction in `FIXME` and link to internal version of "Memoization" page and delete related comments:
-You can find a concrete example of memoization and a more in-depth explanation in the chapter on [Memoization](https://cs3110.github.io/textbook/chapters/ds/memoization.html) of "OCaml Programming: Correct + Efficient + Beautiful."
-
-
+You can find a concrete example of memoization and a more in-depth explanation in the chapter on [Memoization](https://ocaml.org/docs/memoization) of "OCaml Programming: Correct + Efficient + Beautiful."

4) delete indecisive "would" to improve flow
-Most existing modules provide an interface meant to be used in a functional way. Some would require the development and maintenance of [wrapper libraries](https://en.wikipedia.org/wiki/Wrapper_library) to be used in an imperative setting and such use would in many cases be inefficient.
+Most existing modules provide an interface meant to be used in a functional way. Some require the development and maintenance of [wrapper libraries](https://en.wikipedia.org/wiki/Wrapper_library) to be used in an imperative setting and such use results in inefficient code.

5) flip "only is" to "is only"
-1. It depends: only expose state initialisation, which implies there only is a single state
+1. It depends: only expose state initialisation, which implies there is only a single state
 1. Bad: mutable state with no explicit initialisation function or no name referring to the mutable state

6) Move **note** about `Array.truncate` from below the code snippet to above:
+**Note**: The following example code will not run in your REPL; the function `Array.truncate` is not defined. It is provided as an example to contemplate.
-**Note:** This example will not run in the REPL, since the function `Array.truncate` is not defined.

7) Move **note** about an `Analytics` module from below the code snippet to above:
+**Note**: The following example will not run in your REPL; there is no module `Analytics` defined. It is provided as an example to contemplate. [Analytics](https://en.wikipedia.org/wiki/Web_analytics) are remote monitoring libraries.
-**Note:** This code will not run because there is no module called `Analytics`. [Analytics](https://en.wikipedia.org/wiki/Web_analytics) are remote monitoring libraries.

8) inline relevant information from linked content, keeping the link for further reading:
-To ensure that evaluation takes place in a specific order, use the means to put expressions in sequences. Check the [Evaluating Expressions in Sequence](#evaluating-expressions-in-sequence) section.
+To ensure that evaluation takes place in a specific order, use the means to put expressions in sequences using either `let … in` expressions or the semi-colon sequence opertor (`;`). Check the [Evaluating Expressions in Sequence](#evaluating-expressions-in-sequence) section.

### DIFF
--- a/data/tutorials/language/0it_02_loops_and_recursion.md
+++ b/data/tutorials/language/0it_02_loops_and_recursion.md
@@ -89,7 +89,7 @@ programmers like recursion, so `while` loops are second-class citizens
 in OCaml.
 
 If you stop to consider `while` loops, you may see that they aren't really
-any use at all, except in conjunction with our old friend *references*.
+any use at all except in conjunction with our old friend *references*.
 Let's imagine that OCaml didn't have references for a moment:
 
 <!-- $MDX skip -->
@@ -129,7 +129,7 @@ If you want to loop over a list, don't be an imperative programmer and
 reach for your trusty six-shooter Mr. `For` Loop! OCaml has some better
 and faster ways to loop over lists, and they are all located in the
 `List` module. In fact, there are dozens of good functions in `List`, but
-I'll only talk about the most useful ones here.
+we'll restrict ourselves to the most useful ones in this chapter.
 
 First off, let's define a list for us to use:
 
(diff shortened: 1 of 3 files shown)


jakubsvec001:
Incorrectly constructed pull request. See corrected version at: https://github.com/ocaml/ocaml.org/pull/2983

PR #2983, opened 2025-02-25
update imperative.md with minor updates for clarity

jakubsvec001:
These are minor tweaks.

1) add remark noting distinction between operators and language constructs (to complement a similar remark about the sequencing 'operator' later in the document)
+Remark the left arrow symbol <- for mutating mutable record field values is not an operator like the assignment operator ( := ) is for refs. It is rather a construct of the language.

2) explicitly state that 'assign' is an operator:
-The error came from assign :=, which associates stronger than a semicolon ;. Here is what we want to do, in order:
+The error came from the assign operator :=, which associates stronger than a semicolon ;. Here is what we want to do, in order:

3) Follow direction in FIXME and link to internal version of "Memoization" page and delete related comments:
-You can find a concrete example of memoization and a more in-depth explanation in the chapter on [Memoization](https://cs3110.github.io/textbook/chapters/ds/memoization.html) of "OCaml Programming: Correct + Efficient + Beautiful."

-
+You can find a concrete example of memoization and a more in-depth explanation in the chapter on [Memoization](https://ocaml.org/docs/memoization) of "OCaml Programming: Correct + Efficient + Beautiful."

4) delete indecisive "would" to improve flow
-Most existing modules provide an interface meant to be used in a functional way. Some would require the development and maintenance of [wrapper libraries](https://en.wikipedia.org/wiki/Wrapper_library) to be used in an imperative setting and such use would in many cases be inefficient.
+Most existing modules provide an interface meant to be used in a functional way. Some require the development and maintenance of [wrapper libraries](https://en.wikipedia.org/wiki/Wrapper_library) to be used in an imperative setting and such use results in inefficient code.

5) flip "only is" to "is only"
-1. It depends: only expose state initialisation, which implies there only is a single state
+1. It depends: only expose state initialisation, which implies there is only a single state

6) Move note about Array.truncate from below the code snippet to above:
+Note: The following example code will not run in your REPL; the function Array.truncate is not defined. It is provided as an example to contemplate.
-Note: This example will not run in the REPL, since the function Array.truncate is not defined.

7) Move note about an Analytics module from below the code snippet to above:
+Note: The following example will not run in your REPL; there is no module Analytics defined. It is provided as an example to contemplate. [Analytics](https://en.wikipedia.org/wiki/Web_analytics) are remote monitoring libraries.
-Note: This code will not run because there is no module called Analytics. [Analytics](https://en.wikipedia.org/wiki/Web_analytics) are remote monitoring libraries.

8) inline relevant information from linked content, keeping the link for further reading:
-To ensure that evaluation takes place in a specific order, use the means to put expressions in sequences. Check the [Evaluating Expressions in Sequence](https://github.com/ocaml/ocaml.org/pull/2981#evaluating-expressions-in-sequence) section.
+To ensure that evaluation takes place in a specific order, use the means to put expressions in sequences using either let … in expressions or the semi-colon sequence opertor (;). Check the [Evaluating Expressions in Sequence](https://github.com/ocaml/ocaml.org/pull/2981#evaluating-expressions-in-sequence) section.

### DIFF
--- a/data/tutorials/language/0it_06_imperative.md
+++ b/data/tutorials/language/0it_06_imperative.md
@@ -734,7 +734,7 @@ On the other hand, a module may define mutable data internally impacting its beh
 
 ### Bad: Undocumented Mutation
 
-**Note**: The following example code will not run in your REPL; the function `Array.truncate` is not defined. It is provided as an example to contemplate.
+**Note**: The following example code will purposely not run in your REPL; the function `Array.truncate` is not defined. It is provided as an example to contemplate and avoid.
 
 Here's an example of bad code:
 
--- a/data/tutorials/language/0it_06_imperative.md
+++ b/data/tutorials/language/0it_06_imperative.md
@@ -147,6 +147,8 @@ Mutable record fields are updated using the left arrow symbol `<-`. In the expre
 
 In contrast to references, there is no special syntax to dereference a mutable record field.
 
+**Remark** the left arrow symbol `<-` for mutating mutable record field values is not an operator like the assignment operator `( := )` is for `refs`. It is rather a _construct_ of the language. 
+
 ### Remark: References Are Single Field Records
 
 In OCaml, references are records with a single mutable field:
@@ -366,7 +368,7 @@ But here is how it can be made to work:
 val f : int ref -> unit = <fun>
 ```
 
-The error came from assign `:=`, which associates stronger than a semicolon `;`. Here is what we want to do, in order:
+The error came from the assign operator `:=`, which associates stronger than a semicolon `;`. Here is what we want to do, in order:
 1. Increment `r`
 2. Compute `2 * !r`
 3. Assign into `r`
@@ -695,9 +697,7 @@ However, instead of precomputing everything, memoization uses a cache that is po
 * are found in the cache (it is a hit) and the stored result is returned, or they
 * are not found in the cache (it's a miss), and the result is computed, stored in the cache, and returned.
 
-You can find a concrete example of memoization and a more in-depth explanation in the chapter on [Memoization](https://cs3110.github.io/textbook/chapters/ds/memoization.html) of "OCaml Programming: Correct + Efficient + Beautiful."
-
-<!-- FIXME: reference CS3110 memoization documented on ocaml.org after it is merged -->
+You can find a concrete example of memoization and a more in-depth explanation in the chapter on [Memoization](https://ocaml.org/docs/memoization) of "OCaml Programming: Correct + Efficient + Beautiful."
 
 ### Good: Functional by Default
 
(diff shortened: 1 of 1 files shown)


data/tutorials/language/0it_06_imperative.md
```diff
@@ -147,6 +147,8 @@ Mutable record fields are updated using the left arrow symbol `<-`. In the expre
 
 In contrast to references, there is no special syntax to dereference a mutable record field.
 
+**Remark** the left arrow symbol `<-` for mutating mutable record field values is not an operator like the assignment operator `( := )` is for `refs`. It is rather a _construct_ of the language.
```
cuihtlauac:
```suggestion
**Remark**: the left arrow symbol `<-` for mutating mutable record field values is not an operator function, like the assignment operator `( := )` is for `refs`. It is rather a _construct_ of the language, it has no type. 
```

data/tutorials/language/0it_06_imperative.md
```diff
@@ -695,23 +697,21 @@ However, instead of precomputing everything, memoization uses a cache that is po
 * are found in the cache (it is a hit) and the stored result is returned, or they
 * are not found in the cache (it's a miss), and the result is computed, stored in the cache, and returned.
 
-You can find a concrete example of memoization and a more in-depth explanation in the chapter on [Memoization](https://cs3110.github.io/textbook/chapters/ds/memoization.html) of "OCaml Programming: Correct + Efficient + Beautiful."
-
-<!-- FIXME: reference CS3110 memoization documented on ocaml.org after it is merged -->
+You can find a concrete example of memoization and a more in-depth explanation in the chapter on [Memoization](https://ocaml.org/docs/memoization) of "OCaml Programming: Correct + Efficient + Beautiful."
```
cuihtlauac:
```suggestion
You can find a concrete example of memoization and an in-depth explanation in the chapter on [Memoization](https://ocaml.org/docs/memoization) of "OCaml Programming: Correct + Efficient + Beautiful."
```

data/tutorials/language/0it_06_imperative.md
```diff
@@ -734,6 +734,8 @@ On the other hand, a module may define mutable data internally impacting its beh
 
 ### Bad: Undocumented Mutation
 
+**Note**: The following example code will not run in your REPL; the function `Array.truncate` is not defined. It is provided as an example to contemplate.
```
cuihtlauac:
```suggestion
**Note**: The following example code will purposely not run in your REPL; the function `Array.truncate` is not defined. It is provided as an example to contemplate and avoid.
```

cuihtlauac (changes requested):
Thanks @jakubsvec001, your suggestions are great. I only have minor tweaks to add to them.

cuihtlauac (approved):

PR #2986, opened 2025-02-26
(docs) Update option.md with small tweaks and additional examples

jakubsvec001:
This is 1 of 2 pull requests on the option.md tutorial.

I broke these into two parts. I don't intend for part 2's suggestions to be pulled into the repo since they are currently in a separate file for your consideration. There is probably a better way for me to structure such a pair of pull requests such that they can be considered separately and am open to suggestions for the future. My priority was to keep them from stepping on each other during a potential merge.

Part 1:
Part one has some minor tweaks and additional examples. This includes an updated "## Conclusion" to be more encompassing of the lesson as a whole. The deleted "## Conclusion" is included in part 2.

Part 2:
Part two is an extrapolation on the Option.bind implementation example using Option.map and Option.join. I found this example fun to consider but thought it warranted a longer treatment since it is not immediately clear how Option.map and Option.join chain together.
Part 2 can be found here: https://github.com/ocaml/ocaml.org/pull/2987


jakubsvec001:
@cuihtlauac 

I updated this lesson with an implementation of Option.bind using map and join.

It has mentioned [here](https://github.com/ocaml/ocaml.org/pull/2987#issuecomment-2724044564),  there is a short presentation of `bind` and an shortened implementation of bind using `map` and `join`. It also includes a side note about how `bind` and `join` can be used to construct monads.

In general, I added a few examples to several of the functions covered in this lesson. If you wish wish for me to shorten any of it, let me know and I'll remove less informative examples.

sabine (approved):
Thanks @jakubsvec001, this gives more examples and explanations.

PR #2987, opened 2025-02-26
extrapolation on Option.bind example using Option.map and Option.join

jakubsvec001:
This is 1 of 2 pull requests on the option.md tutorial.

I broke these into two parts. I don't intend for part 2's suggestions to be pulled into the repo since they are currently in a separate file for your consideration. There is probably a better way for me to structure such a pair of pull requests such that they can be considered separately and am open to suggestions for the future. My priority was to keep them from stepping on each other during a potential merge.

Part 1:
Part one has some minor tweaks and additional examples. This includes an updated "## Conclusion" to be more encompassing of the lesson as a whole. The deleted "## Conclusion" is included in part 2. 
Part 1 can be found here: https://github.com/ocaml/ocaml.org/pull/2986

Part 2:
Part two is an extrapolation on the Option.bind implementation example using Option.map and Option.join. I found this example fun to consider but thought it warranted a longer treatment since it is not immediately clear how Option.map and Option.join chain together.

### DIFF
--- a/data/tutorials/language/suggested_extrapolation_on_Option.bind_example_using_Option.map_and_Option.join.md
+++ b/data/tutorials/language/suggested_extrapolation_on_Option.bind_example_using_Option.map_and_Option.join.md
@@ -1,0 +1,96 @@
+### A Clever Implementation of Bind
+
+<!-- Jakub's Note: I found the implementation of `bind` using `Fun.flip`, `Option.map`, and `Option.join` to be a bit confusing, so I'm suggesting the above implementation.
+     However, it is a clever utilization of the building blocks introduced by the prior lessons. I broke the clever implementation down in the following section, but don't know it it is too
+     long. I'm curious what your thoughts are. To advocate for the length of this section, it acts as a nice culmination of this lesson. To concede the counterpoint, it requires a lot of text
+     to fully explain step by step -->
+
+There is a more clever way we can define our `bind` function that uses many of the features of the OCaml standard library we've learned so far.
+
+Lets construct our clever version of `bind` slowly. For reference, below is what we will be building up to:
+
+```ocaml
+# let clever_bind o f = o |> Option.map f |> Option.join;;
+val clever_bind : 'a option -> ('a -> 'b option) -> 'b option = <fun>
+```
+
+We will be using `Option.map` as a building block.
+
+First, we must account for the fact that the parameters to `Option.bind` and `Option.map` are flipped:
+
+```ocaml
+# #show Option.bind;;
+val bind : 'a option -> ('a -> 'b option) -> 'b option
+
+# #show Option.map;;
+val map : ('a -> 'b) -> 'a option -> 'b option
+```
+
+There are different ways we can solve this problem.
+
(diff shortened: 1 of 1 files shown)


cuihtlauac:
I'd like this remark to remain in the tutorial on options:

> By the way, any type where `map` and `join` functions can be implemented, with similar behaviour, can be called a _monad_, and `option` is often used to introduce monads. But don't freak out! You don't need to know what a monad is to use the `option` type.

This outlines the view many share in the community. OCaml teaching materials shouldn't give the impression that category theory is a requirement or useful for learning or using OCaml. I believe it is the other way around. Once comfortable with `bind`, `map`, and others, for options, lists, and promises, it is much easier to learn some category theory and figure out what a monad is, from a practical perspective.

With that in mind, I believe we're missing a tutorial showing monad “fun facts” such as:
```ocaml
bind ≅ map ∧ join
extend ≅ map ∧ dupe
app ≅ map ∧ combine
bind -> app -> map
extend -> app -> map

(* with *)

val join : 'a t t -> 'a t
val dupe : 'a t -> 'a t t
val map : ('a -> 'b) -> 'a t -> 'b t
val app : ('a -> 'b) t -> 'a t -> 'b t
val bind : ('a -> 'b t) -> 'a t -> 'b t
val extend : ('a t -> 'b) -> 'a t -> 'b t
val combine : 'a t -> 'b t -> ('a * 'b) t
```

But again, this is merely about showing what are the possible implementation paths. In some sense, it would be a refactoring tutorial.

I wonder if it would also make sense to cover distributivity of pairs other `Either.t`. That would allow to understand why tuples types are products and variants are sums, which explains “algebraic datatypes” which builds sense for “generalized algebraic datatypes”. 

So many things to explain!

cuihtlauac:
P.S.

To create a new tutorial, the markdown file needs a Yaml header. That's why the tests are failing on this PR.

jakubsvec001:
> P.S.
> 
> To create a new tutorial, the markdown file needs a Yaml header. That's why the tests are failing on this PR.

I meant for this pull request to be ephemeral, with the content eventually going into the Option lesson. I didn't want to pollute the other pull request with this longer version, and the other [pull request](https://github.com/ocaml/ocaml.org/pull/2986) removed the interesting passage on using map and join to define bind.

If these two pull requests are combined into one document, the lesson on Option will contain the passage mentioning that there are in fact monads.

If instead you envision these being two separate lessons, I can move the passage on Options being monads into the primary lesson and mention this document as a continuation of the topic.

QUESTION: How would you like to handle these two pull requests? I can merge the documents or make them separate articles. In the latter case, I can add the passage on monads back into the first article.

> With that in mind, I believe we're missing a tutorial showing monad “fun facts” ...

I really like this idea. I'll add "Monad Fun Facts" to the list. 

> This outlines the view many share in the community. OCaml teaching materials shouldn't give the impression that category theory is a requirement or useful for learning or using OCaml. I believe it is the other way around. Once comfortable with bind, map, and others, for options, lists, and promises, it is much easier to learn some category theory and figure out what a monad is, from a practical perspective.

As someone consuming the material from the outside, my impression is that the present material handles both predispositions quite well. Keeping CT in the background and only mentioning it in passing is very effective at peeling back the curtain and showing the possibilities without making it a prerequisite. It makes sense that OCaml's community would want to avoid suggesting CT is a prerequisite.

cuihtlauac:
> QUESTION: How would you like to handle these two pull requests? I can merge the documents or make them separate articles. In the latter case, I can add the passage on monads back into the first article.

In retrospect, the last section [Bind and Option](https://ocaml.org/docs/options#bind-an-option) isn't that great.

We should first have a short presentation on `bind` as it is provided by the standard library and show how it is used. A bit like in the [Error Handling](https://ocaml.org/docs/error-handling#using-on-optionmap-and-optionbind) tutorial. We need a shorter example.

In the last section we make the comparison between `map` and `bind`. Their types and behaviour are very close. We conclude by showing `bind ≅ map ∧ join` as fun fact

This should remain short. 

What do you think?

jakubsvec001:
An updated version of this content has been added to another pull request [2986](https://github.com/ocaml/ocaml.org/pull/2986)

PR #2995, opened 2025-03-04
(data) add Platform Newsletter q4 24

sabine:


sabine:
Let's merge this in the afternoon / evening, in case there's corrections that need to happen before this goes out via social media.

PR #3010, opened 2025-03-09
docs: explain list destructoring syntax

gm0stache:
hi there!
just some small information that i would have loved to have while working through your awesome docs!
thanks for all your work!


cuihtlauac (approved):
Thanks, @gm0stache, this is useful

PR #3013, opened 2025-03-10
Add a GHA workflow for linting markdown files

bbatsov:
It will run only on markdown files that were changed, so we can gradually get to an uniform code style for all of them.


bbatsov:
@cuihtlauac friendly ping :-)

PR #3014, opened 2025-03-10
Improve the formatting of some markdown files

bbatsov:
Basically, everything in "getting started" and "guide".

In case someone is wonder about the stripping of `$` from shell blocks - having them is considered a bad practice, as it:

- it makes it harder to copy commands
- `$` is reserved in markdown for `console` blocks with some output (e.g. from the compiler or whatever)

### DIFF
--- a/data/tutorials/getting-started/1_00_install_OCaml.md
+++ b/data/tutorials/getting-started/1_00_install_OCaml.md
@@ -26,13 +26,13 @@ To install opam, you can [use your system package manager](https://opam.ocaml.or
 If you're installing with [Homebrew](https://brew.sh/):
 
 ```shell
-$ brew install opam
+brew install opam
 ```
 
 Or if you're using [MacPorts](https://www.macports.org/):
 
 ```shell
-$ port install opam
+port install opam
 ```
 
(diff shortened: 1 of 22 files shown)


cuihtlauac (approved):
Thanks, @bbatsov, this is neat.

A few questions:

1. how did you processed the files? Please document how you did aceived this
2. Are we sure, the GHA and the configurations in the others two PR are going to produce consistent results?
3. Is it possible to address very long lines and break them at 70 characters? What's the recommender practice for markdown. Long lines as paragraphs as in TeX or fixed length lines?

bbatsov:
> how did you processed the files? Please document how you did aceived this

I just ran `markdownlint-cli2` with the default configuration. (using the `--fix` param) Sadly, it can't automatically reflow long lines - such problems will have to fixed manually. Also, a few checks by default are set to "consistent" (e.g. list item markers, italics syntax, etc) - this means that `markdownlint` will enforce whatever syntax was first used in a file consistently in the file. This can be changed to always use the same syntax if preferred. 

> Are we sure, the GHA and the configurations in the others two PR are going to produce consistent results?

Yeah, it will be consistent with what's this in this PR. Keep in my however, there are still problems that were flagged, but can't be auto-fixed. (e.g. long lines)

> Is it possible to address very long lines and break them at 70 characters? What's the recommender practice for markdown. Long lines as paragraphs as in TeX or fixed length lines?

I'll have to look at some other options for this later, as `markdownlint` itself can't fix such problems automatically.

jakubsvec001:
@cuihtlauac If you are referring to my pull requests and there is an issue (it sounds like there may not be), let me know and I'll resubmit the pull requests with updates using the new formatting. Happy to do so.

PR #3015, opened 2025-03-11
Update Seq.md with small tweaks, one expanded explanation

jakubsvec001:
Besides small tweaks, I expanded on a description of a recursive function that originally described it as being understandable by "common sense". If every page like this has multiple recursive examples, it may be worth the extra lines to introduce the first (relatively elaborate) recursion example explicitly and leave the rest for the reader to discover on their own, since any web page can serve as an entry-point for a developer new to OCaml.

I hope this approach balances the tradeoff between brevity and clarity. That being said, if you have a clear preference brevity I'm happy to scale it back or leave it the way it was. Just let me know, I'm happy to tailor these pull requests to your needs.

Additionally, I left the last section of this document largely untouched although it contains many markdown comment suggestions for improvement. It is a larger task that may introduce structural changes and is therefore out of scope for this pull request. I'm willing to take a crack at it later, though!

### DIFF
--- a/data/tutorials/language/3ds_05_seq.md
+++ b/data/tutorials/language/3ds_05_seq.md
@@ -12,7 +12,7 @@ prerequisite_tutorials:
 ## Introduction
 
 Sequences are very much like lists. However, from a pragmatic perspective, one
-should imagine they may be infinite. That's the key intuition to understanding
+should imagine they are infinite. That's the key intuition to understanding
 and using sequences. To achieve this, sequence elements are computed on demand
 and not stored in memory. Perhaps more frequently, sequences also allow for
 reducing memory consumption from linear to constant space
@@ -23,7 +23,7 @@ Still in the intro: for people familiar with Python, I believe it would be very 
 
 One way to look at a value of type `'a Seq.t` is to consider it as a list, but it contains
 a twist when it's not empty: its tail is frozen. To understand this analogy,
-consider how sequences are defined in the standard library:
+consider how sequences are defined in the Standard Library:
 
 ```ocaml
 type 'a node =
@@ -68,7 +68,22 @@ writing `fun _ -> a` or `fun () -> a`. The latter function is called a
 [_thunk_](https://en.wikipedia.org/wiki/Thunk). Using this terminology, `Seq.t`
 values are thunks. With the analogy used earlier, `a` is frozen in its thunk.
 
-Here is how to build seemingly infinite sequences of integers:
+## Constructing Sequences
+
+With this understanding, we can manually construct a sequence like so:
+
+``` ocaml
+let my_seq  =
+  fun () ->
+  Seq.Cons (1, fun () -> Seq.Cons (2, fun () -> Seq.Cons (3, fun () -> Seq.Nil)))
+```
+
+**Note:** The second component of each `Seq.Con`'s tuple is a function. This has
+the effect of providing a means of acquiring a value rather than providing a
+value directly.
+
+We can also construct sequences using functions. Here is how to build seemingly
+infinite sequences of integers:
 
 ```ocaml
 # let rec ints n : int Seq.t = fun () -> Seq.Cons (n, ints (n + 1));;
@@ -77,12 +92,16 @@ val ints : int -> int Seq.t = <fun>
 
 The function `ints n` looks as if building the infinite sequence `(n; n + 1; n +
 2; n + 3;...)`. In reality, since machine integers have bounds, the sequence
-isn't indefinitely increasing. When reaching `max_int`, it will circle
+isn't indefinitely increasing. For technical reasons, when `max_int` is reached, it will circle
 down to `min_int`.
 
-The OCaml standard library contains a module on sequences called
-[`Seq`](/releases/5.0/api/Seq.html). It contains a `Seq.iter` function, which
-has the same behaviour as `List.iter`. Writing this:
+The OCaml Standard Library contains a module for sequences called
+[`Seq`](/releases/5.0/api/Seq.html). It contains `Seq.int`, which we implemented above.
+
+## Iterating Over Sequences
+
+The OCaml Standard Library also contains a `Seq.iter` function, which has the
+same behavior as `List.iter`. Writing this:
 
 ```ocaml
 # Seq.iter print_int (ints 0);;
@@ -96,7 +115,7 @@ loop without any output:
 # Seq.iter ignore (ints 0);;
 ```
 
-The key point is: it doesn't leak memory. This example is running in constant
+The key point is that it doesn't leak memory. This example runs in constant
 space. It is effectively nothing more than an infinite loop, which can be
 confirmed by monitoring the space consumption of the program and by noticing
 that it spins forever without crashing. Whereas a version of this with a list
@@ -104,9 +123,9 @@ that it spins forever without crashing. Whereas a version of this with a list
 proportional to the running time, and thus would crash by running out of memory
 pretty quickly.
 
-## Example
+## Taking Parts of a Sequence
 
-The `Seq` module of the OCaml standard library contains the definition of the
+The `Seq` module of the OCaml Standard Library contains the definition of the
 function `Seq.take`, which returns a specified number of elements from the
 beginning of a sequence. Here is a simplified implementation:
 
@@ -124,7 +143,7 @@ let rec take n seq () =
 `seq` contains less than `n` elements, an identical sequence is returned. In
 particular, if `seq` is empty, or `n` is negative, an empty sequence is returned.
 
-Observe the first line of `take`. It is the common pattern for recursive
+Observe the first line of our `take` function. It is the common pattern for recursive
 functions over sequences. The last two parameters are:
 * a sequence called `seq`
 * a `unit` value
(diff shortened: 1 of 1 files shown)


data/tutorials/language/3ds_05_seq.md
```diff
@@ -12,7 +12,7 @@ prerequisite_tutorials:
 ## Introduction
 
 Sequences are very much like lists. However, from a pragmatic perspective, one
-should imagine they may be infinite. That's the key intuition to understanding
+should imagine they are infinite. That's the key intuition to understanding
```
sabine: ah, the thing is sequences can be finite, so they are not always infinite. Worth to make sure that this is captured, and it would be fine to rewrite the entire paragraph to be more clear.

data/tutorials/language/3ds_05_seq.md
```diff
@@ -68,7 +68,22 @@ writing `fun _ -> a` or `fun () -> a`. The latter function is called a
...
+  fun () ->
+  Seq.Cons (1, fun () -> Seq.Cons (2, fun () -> Seq.Cons (3, fun () -> Seq.Nil)))
+```
+
+**Note:** The second component of each `Seq.Con`'s tuple is a function. This has
+the effect of providing a means of acquiring a value rather than providing a
+value directly.
+
+We can also construct sequences using functions. Here is how to build seemingly
+infinite sequences of integers:
```
sabine:
````suggestion
## Constructing Sequences

With this understanding, we can manually construct a sequence like so:

``` ocaml
let my_seq  =
  fun () ->
  Seq.Cons (1, fun () -> Seq.Cons (2, fun () -> Seq.Cons (3, fun () -> Seq.Nil)))
```

**Note:** The second component of each `Seq.Con`'s tuple is a function. This has
the effect of providing a means of acquiring a value rather than providing a
value directly.

We can also construct sequences using functions. Here is how to build an
infinite sequence of integers:
````

Love the addition of showing how to define a finite sequence because it gives more intuition how sequence works.

File: data/tutorials/language/3ds_05_seq.md [hunk 3]
```diff
@@ -351,8 +421,8 @@ the former definition, the application is complete because `fibs` is provided
 with all the arguments it expects. In the latter definition, the application is
 partial because the `()` argument is missing. Since evaluation is
 [eager](https://en.wikipedia.org/wiki/Evaluation_strategy#Eager_evaluation) in
-OCaml, in the former case, evaluation of the recursive call is triggered and a
-non-terminating looping occurs. In contrast, in the latter case, the partially
+OCaml, in the former case evaluation of the recursive call is triggered and
+a non-terminating looping occurs. By contrast in the latter case, the partially
```
sabine:
```suggestion
OCaml, in the former case evaluation of the recursive call is triggered again and again, without ever terminating (this is what "looping recursion" in the error message refers to). In the latter case, the partially
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -77,12 +92,16 @@ val ints : int -> int Seq.t = <fun>
...
 2; n + 3;...)`. In reality, since machine integers have bounds, the sequence
-isn't indefinitely increasing. When reaching `max_int`, it will circle
+isn't indefinitely increasing. For technical reasons, when `max_int` is reached, it will circle
 down to `min_int`.
 
-The OCaml standard library contains a module on sequences called
-[`Seq`](/releases/5.0/api/Seq.html). It contains a `Seq.iter` function, which
-has the same behaviour as `List.iter`. Writing this:
+The OCaml Standard Library contains a module for sequences called
+[`Seq`](/releases/5.0/api/Seq.html). It contains `Seq.int`, which we implemented above.
```
cuihtlauac:
```suggestion
[`Seq`](/releases/api/Seq.html). It contains `Seq.int`, which we implemented above.
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -179,11 +200,32 @@ For instance, here is the list of 100 first prime numbers:
...
-using the `rec` keyword and calls itself. However, some call that kind of
-function [corecursive](https://en.wikipedia.org/wiki/Corecursion). This word is
-used to emphasise that, although it may not terminate, it can indefinitely
-produce valid output.
+The function `trial_div` is recursive in OCaml and can be understood if we break
+it down into its constituent parts. It is defined using the `rec` keyword, allowing the
+function to call itself. For each loop in the recursive call, it pattern-matches
+on either `Seq.Cons (m, seq)` or the end of the sequence, `Seq.Nil`.
+
+If it matches on the first branch (`Seq.Cons (m, seq)`), we filter the
```
cuihtlauac:
```suggestion
If it matches on the first branch `Seq.Cons (m, seq)`, we filter the
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -332,7 +396,13 @@ At that point, Seq.cons has not been introduced yet, so the reader only knows fu
 Stack overflow during evaluation (looping recursion?).
 ```
 
-This definition is behaving as expected (spot the differences, there are four): <!-- How do you count four? -->
+This definition is behaving as expected (spot the differences, there are four): <!-- How do you count four?
+
+1. Seq.Cons vs Seq.cons
```
cuihtlauac:
```suggestion
1. `Seq.Cons` vs `Seq.cons`
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -247,20 +291,40 @@ Here is a quick check:
...
+``` shell
+cat > README.md <<EOF
+This is the first line.
+This is the second line.
+EOF
 ```
 
-It is possible to read a file using `Seq.unfold`:
+Finally, lets define a variable that holds the file's contents and is
+constructed using `Seq.unfold`:
```
cuihtlauac:
```suggestion
Finally, let's read the file's contents using `Seq.unfold`. Mind that `cin` is a local definition.
```

Mind that _variable_ is in our list of banned words, see [`DOCUMENTATION_WRITING.md`](../../blob/main/DOCUMENTATION_WRITING.md)

data/tutorials/language/3ds_05_seq.md
cuihtlauac: Because `Seq.Node` has **two** constructors: `Nil` and `Cons`, a sequence _may_ be finite. This defines a type of strictly infinite sequences:
```ocaml
module Stream : sig
  type 'a t = unit -> 'a node
  and 'a node = Cons of 'a * 'a t 
end
```
We have three cases:
* `list`: strictly finite
* `Stream.t`: strictly infinite
* `Seq.t`: finite or infinite

This tutorial would be easier to write if `Stream` was part of the standard library, unfortunately, it isn't. Arguably there are few use cases for `Stream` (except teaching).

data/tutorials/language/3ds_05_seq.md
```diff
@@ -179,11 +200,32 @@ For instance, here is the list of 100 first prime numbers:
...
+colloquially be called a recursive, is an example of a kind of recursion called
+[corecursive](https://en.wikipedia.org/wiki/Corecursion). Corecursion differs
+from recursion in that it constructs values incrementally rather than computing
+a final result all at once. Unlike traditional recursion, which works towards a
+base case, corecursive functions can produce an infinite stream of values,
+generating only as much as needed. The `trial_div` function is corecursive
+because it does not compute the full sequence of primes immediately. Instead, it
+produces prime numbers lazily, filtering and deferring further computation until
+more elements are requested. This allows the sequence to be processed
+incrementally rather than requiring a complete traversal upfront.
```
cuihtlauac:
```suggestion
**Side Note**: It may be interesting to learn that `trial_div`, while it can
colloquially be called a recursive, is an example of a kind of recursion called
[corecursion](https://en.wikipedia.org/wiki/Corecursion). Corecursion differs
from recursion in that it constructs results incrementally rather than consuming
it's input incrementally. Unlike traditional recursion, which works towards a
base case, corecursive functions must indefinitely produce values as a stream. The `trial_div` function is corecursive
because it does not immediately compute the complete sequence of primes. Instead, it
produces prime numbers on-demand, filtering and deferring further computation until
more elements are requested. This allows the sequence to be processed
incrementally rather than requiring a complete traversal upfront.
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -247,20 +291,40 @@ Here is a quick check:
...
 empty. Otherwise, it returns `None`.
 
-Using this function:
+### Reading a File with `Seq.Unfold`
+
+For the next example, we will demonstrate the versatility of `Seq.unfold` by
+using it to read a file.
+
+Before doing so, lets define a function that processes a file's contents from a
+provided channel:
```
cuihtlauac:
```suggestion
Before doing so, let's define a function that reads a file's line from a
provided channel, with the type signature needed by `Seq.unfold`.
```

cuihtlauac (changes requested):
Thanks, @jakubsvec001, you're doing great

data/tutorials/language/3ds_05_seq.md
jakubsvec001: Thats an interesting way to look at it. I'll put Stream on the list of potential lessons

cuihtlauac: We have this: https://ocaml.org/exercises#100

jakubsvec001: Good catch, thank you :)

cuihtlauac (approved):

cuihtlauac:
Thanks @jakubsvec001

PR #3020, opened 2025-03-13
Removing the status on "is ocaml yet" pages

balat:
Put them back if we manage to define objective criteria.

Previous version only reflects one person opinion and is very counterproductive on a marketing point of view.
OCaml is probably better than any other language for building Web applications on many criteria (typing multitier and multiframework features save a huge amount of time, wasm support, etc.) but has also some missing features and libraries.

How can we compare with just a colored bullet-point? 
Let's not make people run-away immediately.

### DIFF
--- a/src/ocamlorg_frontend/pages/is_ocaml_yet.eml
+++ b/src/ocamlorg_frontend/pages/is_ocaml_yet.eml
@@ -27,36 +27,11 @@ Learn_layout.three_column_layout
 
  <%s! meta.body_html %>
 
-  <ul class="mt-8 px-0 font-light leading-8 text-neutral-800 text-lg flex flex-wrap gap-x-4 gap-y-1">
-  <% meta.categories |> List.iter (fun (category : Data.Is_ocaml_yet.category) -> %>
-    <li class="inline-flex m-0 whitespace-nowrap">
-      <a href="#<%s category.slug %>" class="leading-8"><span class="text-sm"><%s category.status %></span> <%s category.name %></a>
-    </li>
-  <% ); %>
-  </ul>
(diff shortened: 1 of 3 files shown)


cuihtlauac:
Indeed, both:

* agreement on objective maturity criteria and 
* their expression in accurate and understandable terms

are daunting tasks.

But don't forget about maintenance!

How do we check progress of all those projects? What about new ones? How do we keep this information up to date?

balat:
All the more reason to remove this.
Even maintaining this list is a nightmare.

src/ocamlorg_frontend/pages/is_ocaml_yet.eml
cuihtlauac, line 1: You should remove everything from line 30 to 36 as well as the block you commented, lines 37 to 53

cuihtlauac:
You all need to amend `data/is_ocaml_yet/gui.md`

cuihtlauac (changes requested):

src/ocamlorg_frontend/pages/is_ocaml_yet.eml
balat: I removed all the commented code and updated gui. I don't think we should remove lines 30 to 36. This is a kind of table of contents. WDYT?

cuihtlauac: Yes it is a table of contents. But it doesn't add much value. I'd remove it.

The text below “Want to help?” comes from the markdown source

image

balat: Ok! done :)

cuihtlauac (approved):
That looks good to me.

Let's see why the tests are failing

PR #3026, opened 2025-03-17
Update second half of seq.md, structural changes

jakubsvec001:
Some of these updates are to convert code blocks into Utop format.

The remaining update is for the second half of this document. This second half included several suggestions for improvement that required structural changes. 

This pull request may be difficult to read using diff comparison for this reason.

The updated headings read:

```
...
## Consumers vs Producers
### Sequence Consumers: Partially Applied Functions as Parameters
### Sequence Producers: Partially Applied Functions as Results
## Beware of Seq.Cons vs Seq.cons
### Understanding the Difference
## Sequences for Conversions
## Miscellaneous Considerations

- In "Understanding the Difference", I make the generalization that Seq.Cons is particularly useful for generating recursive producers, while Seq.cons is particularly useful for prepending a cons cell to a sequence. I hope this is accurate, happy to be told otherwise if it is overly generic

- After reading some of the suggestions, I kept the `fibs` examples over using a suggested simpler example for `ints`, but I left comments where the simpler `ints` version could be placed. These would require minor updates to the surrounding text, which I can make if you prefer the simpler example

- I expanded a bit on the section describing "producers" and "consumers"

As always, I'm happy to make updates or leave it the way it currently is.

### DIFF
--- a/data/tutorials/language/3ds_05_seq.md
+++ b/data/tutorials/language/3ds_05_seq.md
@@ -73,10 +73,12 @@ values are thunks. With the analogy used earlier, `a` is frozen in its thunk.
 With this understanding, we can manually construct a sequence like so:
 
 ``` ocaml
-# let my_seq  =
-    fun () ->
-      Seq.Cons (1, fun () -> Seq.Cons (2, fun () -> Seq.Cons (3, fun () -> Seq.Nil)))
-val my_seq : unit -> int Seq.node = <fun>
+# let seq_123  =
+    fun () -> Seq.Cons (1,
+      fun () -> Seq.Cons (2,
+        fun () -> Seq.Cons (3,
+          fun () -> Seq.Nil)))
+val seq_123 : unit -> int Seq.node = <fun>
 ```
 
 **Note:** The second component of each `Seq.Con`'s tuple is a function. This has
@@ -324,39 +326,33 @@ This is the second line.
 
 ## Consumers vs Producers
 
-Sequences can be viewed as functions that are either consumers or
-producers. This distinction refers to how functions interact with sequences in
-terms of evaluation strategy, specifically when and how they consume or produce
-elements.
+A function with a sequence parameter consumes it; it's a sequence consumer. A function with a sequence result produces it; it's a sequence producer. In both cases, consumption and production occurs on only one element before continuing with the rest.
 
 ### Sequence Consumers: Partially Applied Functions as Parameters
 
 A consumer is a function that **processes** a sequence, consuming its
 elements. Consumers should be written as higher-order functions that take a
-partially applied function as an argument. This allows for lazy evaluation,
+function parameter. This allows for deferred evaluation,
 ensuring that elements are fetched one at a time instead of forcing the entire
 sequence to be evaluated upfront.
 
-#### Example: `Seq.iter` (Consumer)
+#### Consumer Example: `Seq.iter`
 
 ```ocaml
-# let print_seq s = Seq.iter print_int s;;
+# let print_seq = Seq.iter print_int;;
 val print_seq : int Seq.t -> unit = <fun>
 
 ```
 
-In `print_seq`, `Seq.iter` takes a function (`print_int`) and applies it to each
-element as they are generated. If `Seq.iter` were to be fully applied, it would
-start consuming elements immediately. The consumer function (`print_int`) is a
-partially applied function that only takes one argument at a time.
+In `print_seq`, `Seq.iter` takes the function `print_int` and applies it to each
+element as they are generated. If `List.iter` was used, the whole integer list would be needed before displaying them starts.
 
-### Sequence Producers: Partially Applied Functions as Results
+### Sequence Producers: Functions as Results
 
-A producer is a function that **generates** a sequence. Producers return a
-partially applied function so that elements are only computed when needed.  This
-ensures lazy evaluation and avoids unnecessary computation.
+A producer is a function that **generates** a sequence. Producers return a function so that elements are only computed when needed.  This
+ensures defered evaluation and avoids unnecessary computation.
 
-#### Example: `Seq.unfold` (Producer)
+#### Producer Example: `Seq.unfold`
 
 ```ocaml
  # let naturals =
@@ -364,12 +360,7 @@ ensures lazy evaluation and avoids unnecessary computation.
 val naturals : int Seq.t = <fun>
 ```
 
-`Seq.unfold` returns a function of type `unit -> 'a Seq.node`, making it a lazy
-producer. Each time this function is called, a new element is produced.
-
-To recap, consumers fetch elements only when needed and producers delay
-computation until elements are requested. This pattern ensures that sequences
-remain lazy.
+This application of `Seq.unfold` has type  `unit -> int Seq.node`, making it a function, a deferred producer. Each time this function is called, a new element is produced.
 
 ## Be Aware of Seq.Cons vs Seq.cons
 
@@ -378,19 +369,15 @@ The `Seq` module in the OCaml Standard Library contains two version of
 behavior.
 
 We have already been introduced to the `Seq.Cons` variant constructor. As a
-refresher, below is the type declaration for the `Seq` module:
+refresher, here is how to display its definition:
 
  ``` ocaml
-# type +'a node =
-    | Nil
-    | Cons of 'a * 'a t
-  and 'a t = unit -> 'a node;;
-type 'a node = Nil | Cons of 'a * 'a t
-and 'a t = unit -> 'a node
+# #show Seq.node;;
+type 'a node = 'a Seq.node = Nil | Cons of 'a * 'a Seq.t
 
  ```
 
-The other version of "cons-ing" supported by the Standard Library is the
+The other version of "cons-ing" is the
 function `Seq.cons` (with a lowercase "c") with the following value declaration:
 
 ```ocaml
@@ -398,7 +385,7 @@ val cons : 'a -> 'a Seq.t -> 'a Seq.t
 ```
 
 From the signature, we gather that it is a function that takes two parameters, a
-value and a sequence. Its function declaration is the following:
+value and a sequence. Its definition is the following:
 
 ``` ocaml
  # let cons x next () = Cons (x, next);;
@@ -459,7 +446,7 @@ Stack overflow during evaluation (looping recursion?).
 
 This produces a never-ending recursion that leads to a stack overflow.
 
-### Figs with `Seq.Cons`
+### Fibs with `Seq.Cons`
 
 Next, lets define `fibs_v2` using the constructor `Seq.Cons`:
 
--- a/data/tutorials/language/3ds_05_seq.md
+++ b/data/tutorials/language/3ds_05_seq.md
@@ -77,11 +77,11 @@ With this understanding, we can manually construct a sequence like so:
     fun () -> Seq.Cons (1,
       fun () -> Seq.Cons (2,
         fun () -> Seq.Cons (3,
-          fun () -> Seq.Nil)))
+          fun () -> Seq.Nil)));;
 val seq_123 : unit -> int Seq.node = <fun>
 ```
 
-**Note:** The second component of each `Seq.Con`'s tuple is a function. This has
+**Note:** The second component of each `Seq.Cons`' tuple is a function. This has
 the effect of providing a means of acquiring a value rather than providing a
 value directly.
 
@@ -350,7 +350,7 @@ element as they are generated. If `List.iter` was used, the whole integer list w
 ### Sequence Producers: Functions as Results
 
 A producer is a function that **generates** a sequence. Producers return a function so that elements are only computed when needed.  This
-ensures defered evaluation and avoids unnecessary computation.
+ensures deferred evaluation and avoids unnecessary computation.
 
 #### Producer Example: `Seq.unfold`
 
@@ -378,7 +378,7 @@ type 'a node = 'a Seq.node = Nil | Cons of 'a * 'a Seq.t
  ```
 
 The other version of "cons-ing" is the
-function `Seq.cons` (with a lowercase "c") with the following value declaration:
+function `Seq.cons` (with a lowercase `c`) with the following value declaration:
 
 ```ocaml
 val cons : 'a -> 'a Seq.t -> 'a Seq.t
@@ -408,8 +408,7 @@ val new_cons_v2 : int list = [1; 2; 3; 4; 5; 6; 7; 8; 9; 10]
 ```
 
 Now that we have seen these two versions of "cons-ing" can construct the same
-sequence, it begs the question: what does the function `Seq.cons` provide us
-that the constructor `Seq.Cons` does not?
+sequence, it begs the question: what makes `Seq.cons` and `Seq.Cons` different?
 
 Lets look at how confusing `Seq.Cons` and `Seq.con` can lead to unintended
 behavior.
--- a/data/tutorials/language/3ds_05_seq.md
+++ b/data/tutorials/language/3ds_05_seq.md
@@ -396,15 +396,18 @@ val cons : 'a -> 'a t -> unit -> 'a node = <fun>
 sequence:
 
  ``` ocaml
-# let base_cons = List.to_seq [2; 3; 4; 5; 6; 7; 8; 9]
+# let ints_from_2 = Seq.ints 2
+  let ints_a () = Seq.Cons (1, ints_from_2)  (* With Seq.Cons *)
+  let ints_b = Seq.cons 1 ints_from_2;;      (* With Seq.cons *)
+val ints_from_2 : int Seq.t = <fun>
+val ints_a : unit -> int Seq.node = <fun>
+val ints_b : int Seq.t = <fun>
 
-  let new_cons_v1 = (fun () -> Seq.Cons(1, base_cons))
-                                 |> List.of_seq            (* With Seq.Cons *)
+# ints_a |> Seq.take 3 |> List.of_seq;;
+- : int list = [1; 2; 3]
 
-  let new_cons_v2 = Seq.cons 1 base_cons |> List.of_seq;;  (* With Seq.cons *)
-val base_cons : int Seq.t = <fun>
-val new_cons_v1 : int list = [1; 2; 3; 4; 5; 6; 7; 8; 9; 10]
-val new_cons_v2 : int list = [1; 2; 3; 4; 5; 6; 7; 8; 9; 10]
+# ints_b |> Seq.take 3 |> List.of_seq;;
+- : int list = [1; 2; 3]
 ```
 
 Now that we have seen these two versions of "cons-ing" can construct the same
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: data/tutorials/language/3ds_05_seq.md [hunk 1]
```diff
@@ -73,9 +73,10 @@ values are thunks. With the analogy used earlier, `a` is frozen in its thunk.
 With this understanding, we can manually construct a sequence like so:
 
 ``` ocaml
-let my_seq  =
-  fun () ->
-  Seq.Cons (1, fun () -> Seq.Cons (2, fun () -> Seq.Cons (3, fun () -> Seq.Nil)))
+# let my_seq  =
+    fun () ->
+      Seq.Cons (1, fun () -> Seq.Cons (2, fun () -> Seq.Cons (3, fun () -> Seq.Nil)))
+val my_seq : unit -> int Seq.node = <fun>
```
cuihtlauac:
```suggestion
# let seq_123  =
    fun () -> Seq.Cons (1,
      fun () -> Seq.Cons (2,
        fun () -> Seq.Cons (3,
          fun () -> Seq.Nil)))
val seq_123 : unit -> int Seq.node = <fun>
```

Is this formatting helpful?

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +322,201 @@ This is the second line.
...
+### Sequence Consumers: Partially Applied Functions as Parameters
 
-This example illustrates a non-trivial use of the accumulator, and also the fact that the producer function loops until it finds a new prime to yield, because Seq.unfold does not allow the producer to “skip” a value: it must produce a new element or end the sequence. If we were allowed to skip values, we could more simply do:
+A consumer is a function that **processes** a sequence, consuming its
+elements. Consumers should be written as higher-order functions that take a
+partially applied function as an argument. This allows for lazy evaluation,
+ensuring that elements are fetched one at a time instead of forcing the entire
+sequence to be evaluated upfront.
+
+#### Example: `Seq.iter` (Consumer)
```
cuihtlauac:
```suggestion
A consumer is a function that **processes** a sequence, consuming its
elements. Consumers should be written as higher-order functions that take a
function parameter. This allows for deferred evaluation,
ensuring that elements are fetched one at a time instead of forcing the entire
sequence to be evaluated upfront.

#### Consumer Example: `Seq.iter`
```

Be careful not to confuse “parameter” and “argument.”
* parameter: that's when you define a function, you declare the function's parameters
* argument: that's when you call a function, you pass arguments to the function

Be careful not to confuse “lazy” and “deferred” computation
* deferred: that's the thunk hack that prevents eager evaluation 
* lazy: that's sharing combined with sharing and reuse of previously computed values

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +322,201 @@ This is the second line.
...
+
+```ocaml
+# let print_seq s = Seq.iter print_int s;;
```
cuihtlauac:
```suggestion
# let print_seq = Seq.iter print_int;;
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +322,201 @@ This is the second line.
...
-        Yield (n, (n::prime_list, n+1))
-      else
-        Skip (prime_list, n+1)
-    end
-    ([], 2)
+
+In `print_seq`, `Seq.iter` takes a function (`print_int`) and applies it to each
+element as they are generated. If `Seq.iter` were to be fully applied, it would
+start consuming elements immediately. The consumer function (`print_int`) is a
+partially applied function that only takes one argument at a time.
```
cuihtlauac:
```suggestion
In `print_seq`, `Seq.iter` takes the function `print_int` and applies it to each
element as they are generated. If `List.iter` was used, the whole integer list would be needed before displaying them starts.
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +322,201 @@ This is the second line.
...
 
-```
-let rec next_prime prime_list n =
-  if List.for_all (fun p -> n mod p <> 0) prime_list then
-    n
-  else next_prime prime_list (n+1)
+Sequences can be viewed as functions that are either consumers or
+producers. This distinction refers to how functions interact with sequences in
+terms of evaluation strategy, specifically when and how they consume or produce
+elements.
```
cuihtlauac:
```suggestion
A function with a sequence parameter consumes it; it's a sequence consumer. A function with a sequence result produces it; it's a sequence producer. In both cases, consumption and production occurs on only one element before continuing with the rest.
```

Not sure this is perfect. Let's keep things simple an concrete

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +322,201 @@ This is the second line.
...
+In `print_seq`, `Seq.iter` takes a function (`print_int`) and applies it to each
+element as they are generated. If `Seq.iter` were to be fully applied, it would
+start consuming elements immediately. The consumer function (`print_int`) is a
+partially applied function that only takes one argument at a time.
+
+### Sequence Producers: Partially Applied Functions as Results
+
+A producer is a function that **generates** a sequence. Producers return a
+partially applied function so that elements are only computed when needed.  This
+ensures lazy evaluation and avoids unnecessary computation.
```
cuihtlauac:
```suggestion
A producer is a function that **generates** a sequence. Producers return a function so that elements are only computed when needed.  This
ensures defered evaluation and avoids unnecessary computation.
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +322,201 @@ This is the second line.
...
+
+#### Example: `Seq.unfold` (Producer)
```
cuihtlauac:
```suggestion
#### Producer Example: `Seq.unfold`
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +322,201 @@ This is the second line.
...
-        Skip (prime_list, n+1)
-    end
-    ([], 2)
+
+In `print_seq`, `Seq.iter` takes a function (`print_int`) and applies it to each
+element as they are generated. If `Seq.iter` were to be fully applied, it would
+start consuming elements immediately. The consumer function (`print_int`) is a
+partially applied function that only takes one argument at a time.
+
+### Sequence Producers: Partially Applied Functions as Results
```
cuihtlauac:
```suggestion
### Sequence Producers: Functions as Results
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +322,201 @@ This is the second line.
...
 ```
--->
 
-## Sequences Are Functions
+`Seq.unfold` returns a function of type `unit -> 'a Seq.node`, making it a lazy
+producer. Each time this function is called, a new element is produced.
+
+To recap, consumers fetch elements only when needed and producers delay
+computation until elements are requested. This pattern ensures that sequences
+remain lazy.
```
cuihtlauac:
```suggestion
This application of `Seq.unfold` has type  `unit -> int Seq.node`, making it a function, a deferred producer. Each time this function is called, a new element is produced.
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +322,201 @@ This is the second line.
...
+We have already been introduced to the `Seq.Cons` variant constructor. As a
+refresher, below is the type declaration for the `Seq` module:
+
+ ``` ocaml
+# type +'a node =
+    | Nil
+    | Cons of 'a * 'a t
+  and 'a t = unit -> 'a node;;
+type 'a node = Nil | Cons of 'a * 'a t
+and 'a t = unit -> 'a node
```
cuihtlauac:
```suggestion
# #show Seq.node;;
type 'a node = 'a Seq.node = Nil | Cons of 'a * 'a Seq.t
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +322,201 @@ This is the second line.
...
+remain lazy.
+
+## Be Aware of Seq.Cons vs Seq.cons
+
+The `Seq` module in the OCaml Standard Library contains two version of
+"cons-ing" that warrant special attention due to their similar names and distinct
+behavior.
+
+We have already been introduced to the `Seq.Cons` variant constructor. As a
+refresher, below is the type declaration for the `Seq` module:
```
cuihtlauac:
```suggestion
We have already been introduced to the `Seq.Cons` variant constructor. As a
refresher, here is how to display its definition:
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +322,201 @@ This is the second line.
...
+    | Cons of 'a * 'a t
+  and 'a t = unit -> 'a node;;
+type 'a node = Nil | Cons of 'a * 'a t
+and 'a t = unit -> 'a node
+
+ ```
 
-The `Seq` module contains this definition:
+The other version of "cons-ing" supported by the Standard Library is the
+function `Seq.cons` (with a lowercase "c") with the following value declaration:
```
cuihtlauac:
```suggestion
The other version of "cons-ing" is the
function `Seq.cons` (with a lowercase "c") with the following value declaration:
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +322,201 @@ This is the second line.
...
 
 ```ocaml
 val cons : 'a -> 'a Seq.t -> 'a Seq.t
 ```
 
-Although `Seq.cons x seq` and `Seq.Cons (x, seq)` are the same, `Seq.cons` is a function and `Seq.Cons` is a variant's constructor, which is not the same in OCaml. This can lead to subtle bugs. This section illustrates this.
-<!--
-No need to introduce another mathematical sequence, we can reuse earlier examples (better for pedagogy):
+From the signature, we gather that it is a function that takes two parameters, a
+value and a sequence. Its function declaration is the following:
```
cuihtlauac:
```suggestion
value and a sequence. Its definition is the following:
```

Beware not to confuse “declaration” and “definition”
* a ”declaration” says “this is a name”
* a “definition” says “this is what this names stands for”

Definitions are declarations, but not the other way around

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +322,201 @@ This is the second line.
...
-Although this looks like a possible way to define the [Fibonacci
-sequence](https://en.wikipedia.org/wiki/Fibonacci_number):
+Now that we have seen these two versions of "cons-ing" can construct the same
+sequence, it begs the question: what does the function `Seq.cons` provide us
+that the constructor `Seq.Cons` does not?
+
+In short, `Seq.Cons` provides a convenient means of recursively defining a
+sequence generator and a clumsy means to prepend a value to a sequence.
+Conversely, `Seq.cons` provides a convenient means to prepend a value to a
+sequence and an impossible means to recursively defining a sequence generator.
```
cuihtlauac: This text may come too early. I understand it as an introduction to the `fibs` example. Unfortunately, it turns out into an abstract thing. Therefore it's hard to understand.

Don't be afraid to start with the example, the code, without saying much about it. Understanding `fibs_v1` code is easier and more helpful than understanding this text, and it is “the real thing”.

This is a general rule we try to use in all the tutorials. Show the code first, show the examples first. Once understood, it's possible to express a more general rule or principle that explains the code and other similar codes.

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +322,201 @@ This is the second line.
...
 
-This definition is behaving as expected (spot the differences, there are four): <!-- How do you count four?
+This produces a never-ending recursion that leads to a stack overflow.
 
-1. `Seq.Cons` vs `Seq.cons`
-2. Input is a tuple vs being a pair of parameters
-3. Possesses a Unit value parameter
-4. ?
--->
+### Figs with `Seq.Cons`
```
cuihtlauac:
```suggestion
### Fibs with `Seq.Cons`
```

cuihtlauac (changes requested):
Thanks, @jakubsvec001, you're making this better. A couple of changes.

data/tutorials/language/3ds_05_seq.md
jakubsvec001: I like it!

data/tutorials/language/3ds_05_seq.md
jakubsvec001: Thanks, this was silly of me :)

data/tutorials/language/3ds_05_seq.md
jakubsvec001: Thats a good point!

jakubsvec001: @cuihtlauac I moved this section after the `fibs` example and gave it a heading of "### A Mental Model for `Seq.Cons` vs `Seq.cons`"

data/tutorials/language/3ds_05_seq.md
```diff
@@ -73,9 +73,12 @@ values are thunks. With the analogy used earlier, `a` is frozen in its thunk.
...
 
 ``` ocaml
-let my_seq  =
-  fun () ->
-  Seq.Cons (1, fun () -> Seq.Cons (2, fun () -> Seq.Cons (3, fun () -> Seq.Nil)))
+# let seq_123  =
+    fun () -> Seq.Cons (1,
+      fun () -> Seq.Cons (2,
+        fun () -> Seq.Cons (3,
+          fun () -> Seq.Nil)))
```
cuihtlauac:
```suggestion
          fun () -> Seq.Nil)));;
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -73,9 +73,12 @@ values are thunks. With the analogy used earlier, `a` is frozen in its thunk.
...
+val seq_123 : unit -> int Seq.node = <fun>
 ```
 
 **Note:** The second component of each `Seq.Con`'s tuple is a function. This has
```
cuihtlauac:
```suggestion
**Note:** The second component of each `Seq.Cons`' tuple is a function. This has
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +324,189 @@ This is the second line.
...
-    end
-    ([], 2)
+
+In `print_seq`, `Seq.iter` takes the function `print_int` and applies it to each
+element as they are generated. If `List.iter` was used, the whole integer list would be needed before displaying them starts.
+
+### Sequence Producers: Functions as Results
+
+A producer is a function that **generates** a sequence. Producers return a function so that elements are only computed when needed.  This
+ensures defered evaluation and avoids unnecessary computation.
```
cuihtlauac:
```suggestion
ensures deferred evaluation and avoids unnecessary computation.
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +324,189 @@ This is the second line.
...
+
+ ``` ocaml
+# #show Seq.node;;
+type 'a node = 'a Seq.node = Nil | Cons of 'a * 'a Seq.t
 
-The `Seq` module contains this definition:
+ ```
+
+The other version of "cons-ing" is the
+function `Seq.cons` (with a lowercase "c") with the following value declaration:
```
cuihtlauac:
```suggestion
function `Seq.cons` (with a lowercase `c`) with the following value declaration:
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +324,189 @@ This is the second line.
...
+val new_cons_v1 : int list = [1; 2; 3; 4; 5; 6; 7; 8; 9; 10]
+val new_cons_v2 : int list = [1; 2; 3; 4; 5; 6; 7; 8; 9; 10]
 ```
--->
 
-Although this looks like a possible way to define the [Fibonacci
-sequence](https://en.wikipedia.org/wiki/Fibonacci_number):
+Now that we have seen these two versions of "cons-ing" can construct the same
+sequence, it begs the question: what does the function `Seq.cons` provide us
+that the constructor `Seq.Cons` does not?
```
cuihtlauac:
```suggestion
sequence, it begs the question: what makes `Seq.cons` and `Seq.Cons` different?
```

cuihtlauac (changes requested):
Small changes

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +324,188 @@ This is the second line.
...
+The `Seq` module in the OCaml Standard Library contains two version of
+"cons-ing" that warrant special attention due to their similar names and distinct
+behavior.
+
+We have already been introduced to the `Seq.Cons` variant constructor. As a
+refresher, here is how to display its definition:
+
+ ``` ocaml
+# #show Seq.node;;
+type 'a node = 'a Seq.node = Nil | Cons of 'a * 'a Seq.t
```
cuihtlauac:
```suggestion
# #show Seq.node;;
type 'a node = 'a Seq.node = Nil | Cons of 'a * 'a Seq.t
```

data/tutorials/language/3ds_05_seq.md
```diff
@@ -325,124 +324,188 @@ This is the second line.
...
+ ``` ocaml
+# let base_cons = List.to_seq [2; 3; 4; 5; 6; 7; 8; 9]
+
+  let new_cons_v1 = (fun () -> Seq.Cons(1, base_cons))
+                                 |> List.of_seq            (* With Seq.Cons *)
+
+  let new_cons_v2 = Seq.cons 1 base_cons |> List.of_seq;;  (* With Seq.cons *)
+val base_cons : int Seq.t = <fun>
+val new_cons_v1 : int list = [1; 2; 3; 4; 5; 6; 7; 8; 9; 10]
+val new_cons_v2 : int list = [1; 2; 3; 4; 5; 6; 7; 8; 9; 10]
```
cuihtlauac:
```suggestion
# let ints_from_2 = Seq.ints 2
  let ints_a () = Seq.Cons (1, ints_from_2)  (* With Seq.Cons *)
  let ints_b = Seq.cons 1 ints_from_2;;      (* With Seq.cons *)
val ints_from_2 : int Seq.t = <fun>
val ints_a : unit -> int Seq.node = <fun>
val ints_b : int Seq.t = <fun>

# ints_a |> Seq.take 3 |> List.of_seq;;
- : int list = [1; 2; 3]

# ints_b |> Seq.take 3 |> List.of_seq;;
- : int list = [1; 2; 3]
```

cuihtlauac (dismissed):

data/tutorials/language/3ds_05_seq.md
jakubsvec001: @cuihtlauac I am unable to commit this suggestion: the tooltip says "This suggestion is invalid because no changes were made to the code".

I tried marking this conversation as resolved, but it did not change the status of the request for change.

sabine (approved):
Thank you so much @jakubsvec001, and thanks for your patience. We got around to reviewing this and it's adding a lot of helpful context / examples!

PR #3032, opened 2025-03-25
Fix: cannot type / in playground editor

dinakajoy:
Since the search bar is displayed on every page, @contificate suggestions was implemented
> Perhaps the handler should be adapted to check whether the editor component is focused?


sabine (approved):
Thanks @dinakajoy, this is very helpful!

patricoferris:
I've hit this so many times, thanks @dinakajoy for taking the time to fix this.

PR #3038, opened 2025-03-28
Cookbook recipes to parse command-line arguments

boechat107:
The recipes in this PR contains three alternative ways to parse command-line arguments:

1. Manipulating `Sys.argv` directly.
2. Using the `Arg` module from the standard library.
3. Using `Cmdliner` installed through `opam`.

### DIFF
--- a/data/cookbook/parse-command-line-arguments/00-stdlib.ml
+++ b/data/cookbook/parse-command-line-arguments/00-stdlib.ml
@@ -9,26 +9,23 @@ let simplest_parser () =
   |> Array.to_list
   |> List.iteri (Printf.printf "argument %d: %s\n")
 
-
 (* `Arg` is a module for parsing command-line arguments, and it is part of
    OCaml's standard library
    ([documentation](https://ocaml.org/manual/5.3/api/Arg.html)).
    In this function we define the structure of the command-line arguments we
    expect, the argument types types and their documentation. This is basically
-   the same function defined in the module's documentation.
-
- *)
+   the same function defined in the module's documentation. *)
 let arg_module_parser () =
   let usage_msg =
     "mycmd [--verbose] <file1> [<file2>] ... -o <output>"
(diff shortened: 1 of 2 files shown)


boechat107:
I can't understand why the CI failed. I'll investigate `ood-gen`.

If someone has seeing this before, I would appreciate any help.

data/cookbook/parse-command-line-arguments/00-stdlib.ml
```diff
@@ -0,0 +1,65 @@
...
+   the same function defined in the module's documentation.
+
+ *)
+let arg_module_parser () =
+  let usage_msg =
+    "mycmd [--verbose] <file1> [<file2>] ... -o <output>"
+  and verbose = ref false
+  and input_files = ref []
+  and output_file = ref "" in
+  (* This function is called once for each anonymous argument. *)
```
boechat107: `ood-gen` is parsing this whole block as a `Str.Text`, except for the final delimiter `*)`. There seems to be something wrong with the string splitting ([`Str.full_split`, line 72](https://github.com/boechat107/ocaml.org/blob/cookbook-parse-command-line/tool/ood-gen/lib/cookbook.ml#L72)).

boechat107: Ok, got it. The regex expect the delimiter `(*` always at the beginning of the line. This code has indented comments.

boechat107: I couldn't find any comments about this comments restriction. Would it make sense to add a comment? Or would it be better to change the regex pattern?

cuihtlauac:
This needs to be rebased against a fresh `main`

cuihtlauac:
See also #2573

sabine:
Closing because https://github.com/ocaml/ocaml.org/pull/2573 was merged.

PR #3069, opened 2025-04-19
Add the changelog for opam 2.4.0~alpha1

kit-ty-kate:


cuihtlauac:
Hi @kit-ty-kate. 

Do we need to have prereleases in the ocaml.org changelog? Aren't those versions only for the interested?

sabine:
IMO pre-releases need to be announced and they belong here until the Changelog is split into a stable and an unstable channel.

sabine:
(And this split is bound to happen soon, with dedicated social channels (stable vs experimental/unstable) linked to distribute these announcements to the respective audiences.)

cuihtlauac:
I don't see Rust, Go, Scala, Kotlin and Swift listing prereleases. Python does. Some projects list the latest prerelease, but it's transient. Should we revert #2971?

sabine:
I don't know why they were removed in the first place, instead of splitting up the Changelog into the two audiences we want to reach. No need to revert right now, but I will definitely recover this content the moment there is a dedicated channel for this.

sabine (approved):

PR #3072, opened 2025-04-23
Remove `ocaml-eglot` from the Changelog

xvw:


sabine:
Merging because we can re-add it after ocaml-eglot becomes an official incubating OCaml Platform Tool, or we find a way to make emerging editors tooling like this fall under some umbrella

PR #3094, opened 2025-05-09
fix: update `r2c` in industry users

brandonspark:
This PR updates one of the `Success stories`, specifically `r2c`, which has [since been renamed to Semgrep](https://en.wikipedia.org/wiki/Semgrep).

I also updated some stuff about what we currently do, and added our new logo.

I am a current employee of Semgrep.


cuihtlauac (approved):
Thanks @brandonspark

PR #3101, opened 2025-05-12
Use Posthog rather than Plausible

mtelvers:
cc @sabine


avsm:
This appears to have no material change in the data collected or any privacy implications, so I'm fine with it in principle. (Except, from reading Mark's excellent [blog post](https://www.tunbury.org/posthog/) about it, do we really need _16 gigabytes of RAM_ to host this service? What's it doing with all that RAM?!)

rikusilvola:
Could we run both until the end of Q2 to not lose track of the metrics gathered until now, or if we choose to return to plausible after a trial?

sabine:
Indeed, retiring Plausible after Q2 makes sense.

mtelvers:
The statistics from Plausible show a broadly flat usage pattern; therefore, it's unlikely that a missing data block would cause a significant loss of insight. Running both would add an additional burden on the user's browser.

rikusilvola:
Good point @mtelvers .
@sabine, please make sure to capture the latest data from Plausible for all the metrics we track before the transition to DataHog.

Alternatively we can postpone the change until end of Q2.

sabine:
I expect the burden on the user's browser to be minimal from Plausible's side, as it has so few features in the community edition, their script is small, and they state that they care about performance (https://plausible.io/lightweight-web-analytics).

We don't even know at this point whether Posthog will replace Plausible. Running them in parallel until end of Q2 makes sense, IMO, and also creates less issues with exporting data from Plausible and later combining it with Posthog data.

PR #3102, opened 2025-05-13
Fix: Syntax Error in Functors Tutorial

### LINKED ISSUE #3085: Syntax error in tutorial
[https://ocaml.org/docs/functors](https://ocaml.org/docs/functors)

```ocaml
module type SeedType : sig   
  val v : int array
end

module type S : sig
```

StevenH34:
Fixes issue #3085 

This change fixes the syntax error in ```random.ml``` on page [https://ocaml.org/docs/functors](https://ocaml.org/docs/functors).


cuihtlauac (approved):
Thanks @StevenH34

PR #3118, opened 2025-05-20
Add Ability to ignore scraped OCaml Planet posts, fix existing posts

sabine:
* set default for `only_ocaml` flag to true - importing everything should be the exception, not the default
* add ability to ignore scraped posts that shouldn't be on the OCaml Planet
* Jon's RSS feed had `<a>` tags in the titles of the posts
* ignore some of Anil's posts that are too far out of context for the OCaml Planet
* remove Ryan's blog since posts don't provide enough context for the OCaml Planet


tool/ood-gen/lib/blog.ml
```diff
@@ -56,39 +56,44 @@ module Post = struct
...
+            | Ok s -> s
+            | Error (`Msg e) -> (
+                match m.source with
+                | Some { name; url } ->
+                    {
+                      id = "";
+                      name;
+                      url;
+                      description = "";
+                      only_ocaml = false;
```
yawaramin, line 82: Should this be `true`?

PR #3125, opened 2025-05-22
Don't save or republish full OCaml Planet blog post content

### LINKED ISSUE #3120: Planet: are post bodies needed?
Currently, blog posts are scraped for OCaml Planet and stored in the repo, including the full bodies of the posts: https://github.com/ocaml/ocaml.org/blob/main/data/planet/dra27/dipping-toes-into-oxcaml.md?plain=1

However on the Planet page only post blurbs are shown and the 'Read Article' links just go to the original posts: https://ocaml.org/ocaml-planet

Ie the fully body text of the posts never seems to be shown on the Planet page. Am I missing something, or does it seem like we don't actually need to store the post bodies in this repo?

yawaramin:
Don't save the full post content in Markdown files like https://github.com/ocaml/ocaml.org/blob/main/data/planet/dra27/dipping-toes-into-oxcaml.md?plain=1 or republish the full content in the Planet RSS feed, https://ocaml.org/planet.xml

There's no need to save the full post content because we never display it anywhere in our site. We just directly link to the original blog posts.

Fix #3120

### DIFF
--- a/src/ocamlorg_data/data_intf.ml
+++ b/src/ocamlorg_data/data_intf.ml
@@ -61,7 +61,6 @@ module Blog = struct
     authors : string list;
     date : string;
     preview_image : string option;
-    body_html : string;
   }
 end
 
--- a/tool/ood-gen/lib/blog.ml
+++ b/tool/ood-gen/lib/blog.ml
@@ -62,7 +62,7 @@ module Post = struct
 
   let all_sources = Source.all ()
 
-  let of_metadata ~source ~body_html m : Data_intf.Blog.post option =
+  let of_metadata ~source m : Data_intf.Blog.post option =
     if Option.value ~default:false m.ignore then None
     else
       Some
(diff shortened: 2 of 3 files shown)


sabine:
We republish in the RSS feed, though.

sabine:
But I think those shortened content pieces in the RSS feed could as well be replaced by an AI summary or something like that.

ETA: I'll address this when I touch the RSS feed / social media automations again.

cuihtlauac:
How does this compare to #2573 ?

yawaramin:
I was trying to say that there's no need to republish the full content of the posts in the Planet RSS feed; in fact many people prefer that feed aggregators don't do that and rather they send the traffic to the original posts.

sabine (approved):
@yawaramin Ok, you're right, this is reasonable and the feed we publish gets a lot smaller this way.

PR #3127, opened 2025-05-22
Add GitHub Actions workflow for dune pkg

Sudha247:
This PR adds a GitHub Actions workflow for ocaml.org. This would help us catch any failures with `dune pkg`. Also fixed a deprecated function in `Mirage_crypto_rng` along the way.

### DIFF
--- a/.github/workflows/build-dune.yml
+++ b/.github/workflows/build-dune.yml
@@ -1,0 +1,19 @@
+name: Build on Dune PM
+
+on: [push, pull_request]
+
+jobs:
+  build:
+    strategy:
+      matrix:
+        runs-on: [ ubuntu-latest, macos-latest ]
+    runs-on: ${{ matrix.runs-on }}
+    continue-on-error: true
+
+    steps:
+      - name: Checkout code
+        uses: actions/checkout@v4
+      - name: Use dune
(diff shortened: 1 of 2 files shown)


Sudha247:
To add more context: This is a preliminary check whether the project builds with dune developer preview. It seems there's some issue with depext, I'll investigate this tomorrow.

cuihtlauac:
Wouldn't it be more prudent to begin by a local setup allowing the use of `dune pkg` instead of a GitHub action?

Sudha247:
> Wouldn't it be more prudent to begin by a local setup allowing the use of dune pkg instead of a GitHub action?

Indeed, and this setup was working on my machine - it was building and passing all tests. Maybe because depexts were already installed on my machine. I also discovered that it installs OCaml 4.14.x compiler version when I build it with `dune pkg`. Is that expected?

sabine:
Thank you Sudha, seeing that it builds on 4.14 and how little needed to be done for that was helpful. I'm closing this in favor of https://github.com/ocaml/ocaml.org/pull/3132 where I tried to get OCaml.org to work with Dune Package Management on OCaml 5.2.0 :orange_heart:

PR #3132, opened 2025-05-26
chore: Use Dune package management

sabine:
This patch changes the build configuration to use Dune Package Management.

Done / To do:
* [x] we are waiting on a stable installation path for standalone Dune so that we can tell contributors to install via that way (instead of pulling a pinned version of the nightly "developer preview" build)
* [x] update `dune-workspace` to pin `opam-repository` just as we currently do (moved the pins from `dune-project` there / removed `river` pin, as it's no longer necessary / pin opam-repository inside `dune-workspace`)
* [x] update `Dockerfile`
* [x] remove `.opam` files
* [x] update `Makefile`s to reflect using only Dune, and no longer using `opam`
* [x] update `CONTRIBUTING.md`  and `HACKING.md`
* [x] update `playground/` to build with Dune Developer Preview
* [x] update the CI to use Dune developer preview (thanks Sudha for pointing this out and giving a pointer to a GH action script for Dune Developer Preview)
* [x] update the scraper to use Dune developer preview
* [x] Use a pinned version of dune dev preview so the ocaml.org build does not break unexpectedly. Rationale: we sometimes needs to update on short notice and can't afford to have a broken build
* [x] How do we do power clean? That is getting back to point were `dune pkg` needs to be initialised. Rationale: this is needed to measure first build time without using docker
* [x] ocaml.org build is using `~/.cache/ocamlorg` is `dune pkg` also using `~/.cache/dune`. What's the impact of deleting it?
* [x] Create a GitHub action that uses daily dune preview and current version of ocaml.org build as a regression/blackbox test. But let's not do that with the production version
* [x] Try to merge unrelated changes as much as possible from the patch into main to reduce PR size [on @cuihtlauac]
* [x] Update contributor documentation to make it clear that removing `dune.lock` directory is enough to go back to the old build -- relevant only for open PRs / current work in progress, we should encourage people to rebase on main and use Dune package management after the merge of this patch
* [x] Rebase tested. It should be done to minimize diff surface. Also it needs to be done again when PRs #3178 #3179 #3180 and #3181 are merged

## To test this

```
$ make start

$ make playground

$ make scrape
```

## Current status

I believe this is ready to merge.


gridbugs:
I've been using dune package management to build this project whenever I've needed to contribute to it over the past year and it's worked fine until recently. Here are some PRs/issues relating to fixing the problems I've encountered recently:
- https://github.com/ocaml/ocaml.org/pull/3147
- https://github.com/ocaml/ocaml.org/pull/3146
- https://github.com/ocaml/ocaml.org/issues/3148

gridbugs:
There's also currently an issue with the `ocamlnet` package source being missing. I've [added](https://github.com/ocaml/opam-source-archives/pull/46) it to the opam-source-archives repo and I have a [PR](https://github.com/ocaml/opam-repository/pull/27986) to point the `ocamlnet` package at the archived source. Once the latter PR is merged, updating the opam-repository revision in this project's dune-workspace to the current tip of opam-repository will fix the problem with `ocamlnet`.

Sudha247:
This is great! Thanks for the work, @sabine and @gridbugs. Upon offline discussion with Sabine, we wanted to test this on MacOS and Windows to ensure the workflow remains unaffected for contributors on those platforms. I tested with WSL2 (which is the recommended way to develop ocaml.org on Windows), and the ocaml.org project (with `make start`), builds and runs locally on WSL2, in this branch with `dune pkg`. Moreover, the testsuite passes too.

The ocaml.org playground doesn't build, as expected - as it seems to have some dependency issues. I'll try to investigate what's happening there!

sabine:
Sounds great! Looks like we'll be migrating fully to `dune pkg` as soon as the playground builds.

sabine:
Thanks @Sudha247 for https://github.com/sabine/ocaml.org/pull/53. This makes the playground dune build succeed.

One thing I observe on my machine:

Both on the main branch, and on this branch, running the playground build just deletes the `.cmi` files in the `asset/` folder of the playground. It should be copying the `.cmi` files from somewhere. The script `gen_stdlib.sh` performs this part by looking up via `opam` the location of these `.cmi` files, so we need to either find where they reside with a dune build, or find another way to regenerate these files. :thinking:

HACKING.md
cuihtlauac, line 1: Replace all ````` by ````bash` in this file

dune-project
cuihtlauac, line 1: Could the removal of the dependency on Anton's river fork be merged ahead in a separate PR?

cuihtlauac (changes requested):

Dockerfile
```diff
@@ -10,39 +13,34 @@ RUN sudo apk -U upgrade --no-cache && sudo apk add --no-cache \
...
-RUN cd ~/opam-repository && git reset --hard 584630e7a7e27e3cf56158696a3fe94623a0cf4f && opam update
-
-WORKDIR /home/opam
-
-# Install opam dependencies
-COPY --chown=opam ocamlorg.opam .
-RUN opam install . --deps-only
+# Install Dune Developer Preview
+RUN curl -fsSL https://get.dune.build/install | sh
+RUN /bin/bash -c 'source "/root/.local/share/dune/env/env.bash"'
```
Leonidas-from-XIV: Does this have any effect? The environment disappears after this call.

Dockerfile
```diff
@@ -1,7 +1,10 @@
-FROM ocaml/opam:alpine-3.21-ocaml-5.2 AS build
+FROM alpine:3.21 AS build
 
 # Install system dependencies
-RUN sudo apk -U upgrade --no-cache && sudo apk add --no-cache \
+RUN apk -U upgrade --no-cache && apk add --no-cache \
+    # to download and install Dune Developer Preview with alpine:3.21
+    build-base patch tar ca-certificates git rsync curl sudo bash \
```
Leonidas-from-XIV: Is there a need for `sudo`, `rsync`, `bash` or `nano`?

playground/dune-project
```diff
@@ -33,15 +31,55 @@
...
-  (js_of_ocaml
-   (>= 5.0))
-  js-top-worker
-  code-mirror
-  merlin-js
-  rpclib
-  rpclib-lwt))
+  (astring (= 0.8.5))
+  (base-bigarray (= base))
+  (base-bytes (= base))
```
Leonidas-from-XIV: I don't think it is customary to depend on all these `base-*` packages. After all, they're just transitive dependencies of the compiler selected.

Dockerfile
cuihtlauac: We must not download a dev preview on each build, that's brittle and wasteful.

Makefile
```diff
@@ -1,74 +1,62 @@
...
 
 .PHONY: fmt
 fmt: ## Format the codebase with ocamlformat
-	opam exec -- dune build --root . --auto-promote @fmt
+	dune build --root . --auto-promote @fmt
 
 .PHONY: watch
 watch: ## Watch for the filesystem and rebuild on every change
-	opam exec -- dune build @run -w --force --no-buffer
+	dune build @run -w --force --no-buffer
```
v-gb: I read the HACKING doc a bit quickly, and ran `make watch` instead of `make start` first:

```
$ make watch
dune build @run -w --force --no-buffer

File "src/ocamlorg_web/lib/dune", line 13, characters 2-15:
13 |   mirage-kv-mem))
       ^^^^^^^^^^^^^
Error: Library "mirage-kv-mem" not found.
-> required by library "ocamlorg_web" in _build/default/src/ocamlorg_web/lib
-> required by executable main in src/ocamlorg_web/bin/dune:2
-> required by _build/default/src/ocamlorg_web/bin/main.exe
-> required by alias src/ocamlorg_web/bin/run in src/ocamlorg_web/bin/dune:8
######################################################################## 100.0%                                              
Done: 95% (95/100, 5 left, 1 failed) (jobs: 1)Browserslist: caniuse-lite is outdated. Please run:
  npx update-browserslist-db@latest
  Why you should do it regularly: https://github.com/browserslist/update-db#readme

Rebuilding...

Done in 1911ms.
```

Shouldn't every rule that does a build depend on `dune.lock`, instead of just a couple of them? (or taking a step back, I thought the usual thing to do is commit the `dune.lock`, so I'm a bit confused about why it's being generated on the fly).

Dockerfile
yawaramin: You also probably don't need `curl`. Docker can itself download remote files. See https://docs.docker.com/reference/dockerfile/#adding-files-from-a-url

Makefile
sabine: thanks, you're right. need to update the Makefile

Dockerfile
Sudha247: Indeed, but dune uses curl internally for `(url ..)` fields, so we still need it.

dev-tools.locks/ocamlformat/ocamlformat.pkg
```diff
@@ -0,0 +1,14 @@
+(version
+ 0.26.2+binary-ocaml-5.2.1-built-2024-12-04.0-x86_64-unknown-linux-musl)
```
gridbugs: Be careful checking this file in as it's not portable. This version of ocamlformat will only work on linux. I'd recommend adding the dev-tools.locks directory to .gitignore so it doesn't get checked in and each user can install the version of the dev tools appropriate to their platform.

shonfeder:
The install script is ready here, thanks to @gridbugs' work: https://github.com/ocaml-dune/dune-bin-install

@Sudha247 is planning to add documentation here about how to use it.

It also sounds like we should take Steve's advice about ignoring the lock dir, and IIUC @cuihtlauac's review has been addressed and just needs to be re-requested to clear the change request?

dune-workspace
```diff
@@ -1,8 +1,19 @@
...
-(repository (name pinned_opam_repository) (url git+https://github.com/ocaml/opam-repository#584630e7a7e27e3cf56158696a3fe94623a0cf4f))
+(repository
+ (name pinned_opam_repository)
+ (url git+https://github.com/ocaml/opam-repository#584630e7a7e27e3cf56158696a3fe94623a0cf4f))
+
+(pin
+(name olinkcheck)
+ (url "git+https://github.com/tarides/olinkcheck")
+ (package
+  (name olinkcheck)))
```
cuihtlauac: PR #3144 Removed olinkcheck. I dont think this is needed any longer

dune-workspace
```diff
@@ -1,8 +1,19 @@
...
 
 (repository (name pinned_overlay_repository) (url git+https://github.com/ocaml-dune/opam-overlays#2a9543286ff0e0656058fee5c0da7abc16b8717d))
 
 (lock_dir
-  (repositories pinned_opam_repository pinned_overlay_repository))
+  (repositories pinned_opam_repository pinned_overlay_repository)
+  (pins olinkcheck)
+   (version_preference newest)
+  )
```
cuihtlauac: Same remark w.r.t. PR#3144

.github/workflows/ci.yml
```diff
@@ -22,21 +22,13 @@ jobs:
...
-      - name: Use OCaml ${{ matrix.ocaml-compiler }}
-        uses: ocaml/setup-ocaml@v3
-        with:
-          ocaml-compiler: ${{ matrix.ocaml-compiler }}
-          dune-cache: ${{ matrix.os != 'macos-latest' }}
-          opam-repositories: |
-            pin: git+https://github.com/ocaml/opam-repository#584630e7a7e27e3cf56158696a3fe94623a0cf4f
-          opam-disable-sandboxing: true
+      - name: Use Dune Developer Preview
+        uses: ocaml-dune/setup-dune@v0
```
cuihtlauac, line 31: I'm not sure I fully understand this. 

Are we attempting to build ocaml.org with a build tool that changes daily?

cuihtlauac (changes requested):
Small remarks:
1. File `.github/workflows/scrape_changelog.yml` should be updated.
2. There are a few remaining references to `olinkcheck` that should be removed, it is no longer used

Big question

Is this PR proposing to build ocaml.org with a version of Dune that changes daily? If that is the case, I must object. This is unwise. I support building ocaml.org as a regression test in Dune developer preview CI. However, using an unstable build tool in a production delivery chain is reckless.

Sudha247:
> Is this PR proposing to build ocaml.org with a version of Dune that changes daily? If that is the case, I must object.

The PR in its current state proposes this approach. However, as @shonfeder noted above, we can use @gridbugs' installer to install the released Dune binary (currently version 3.19.1). I'll update the PR and documentation accordingly. The idea is to use `dune pkg` for building and deploying the website. Since `dune pkg` is available in the released version of Dune, we don’t need the developer preview as the experimental features it offers aren’t essential for this PR.

HACKING.md
```diff
@@ -24,32 +24,20 @@ git clone https://github.com/<username>/OCaml.org.git
 cd OCaml.org
 ```
 
-Ensure you have `opam` installed. Opam will manage the OCaml compiler along with all of the OCaml packages needed to build and run the project. By this point, we should all be using some Unix-like system (Linux, macOS, WSL2), so you should [run the opam install script](https://opam.OCaml.org/doc/Install.html#Binary-distribution). There are also manual instructions for people that don't want to run a script from the internet. We assume you are using `opam.2.1.0` or later, which provides a cleaner, friendlier experience when installing system dependencies.
+Ensure you have [Dune Developer Preview](https://preview.dune.build) installed. Dune will manage the OCaml compiler along with all of the OCaml packages needed to build and run the project. By this point, we should all be using some Unix-like system (Linux, macOS, WSL2). We assume you are using the most recent version of Dune Developer Preview.
```
shonfeder: IIUC, once we update this to advise (and instruct) users on installing the stable version via Steve's script, this PR should be good to merge!

dev-tools.locks/ocamlformat/ocamlformat.pkg
shonfeder: Oh, this should be addressed too, I think?

HACKING.md
shonfeder: I'm wrong, the pending comments from @cuihtlauac and @gridbugs also need to be addressed. But then we should be good!

Sudha247: The pending comments are addressed in https://github.com/sabine/ocaml.org/pull/60, I think. As usual, it needs to be rebased to @sabine's branch..

shonfeder: Oh, I see! Sorry @Sudha247. I did not realize that you could not push to this branch. Thank you for clarifying.

@sabine could you please incorporate https://github.com/sabine/ocaml.org/pull/60 when you have a free cycle, so we get this wrapped up! Thanks :)

cuihtlauac:
As it is today, several versions of Dune would be used concurrently:
* Developer Preview from [2025 July 17](https://get.dune.build/2025-07-17/x86_64-unknown-linux-musl/dune-2025-07-17-x86_64-unknown-linux-musl.tar.gz) in the Dockerfile
* Day of run Developer Preview in CI action (file .github/workflows/ci.yml)
* Some unspecified version of the Dune's Developer Preview in local setups

As a stable and production-grade project, ocaml.org should not depend on experimental tools. If you'd like to use ocaml.org to test the developer preview, that's fine, but then fork or vendor it.

shonfeder:
@cuihtlauac are you taking into account the changes pending from https://github.com/sabine/ocaml.org/pull/60 ?

sabine:
Merged @Sudha247's changes and rebased on `main`.

cuihtlauac (changes requested):
The following changes are required:

* [ ] Update `.github/workflows/scrape_changelog.yml`
* [ ] Update `README.md`, section “Getting Started” 
* [ ] Do not use developer preview in `ci.yml` the same build configuration must be used there, in `Dockerfile` and locally
* [ ] Fix `make docker` build under Ubuntu 25.04
* [ ] Remove file `ocamlorg.opam`

shonfeder:
>  Remove file ocamlorg.opam

IIUC, this is not desirable, as we still want to support opam-based workflows.

shonfeder:
For the record, at one point @cuihtlauac was getting this failure on the docker build

```
ocaml.org:dune_pkg>make docker                                                                                                                    ocaml.org
docker build -f Dockerfile . -t ocamlorg:latest
[+] Building 115.5s (13/23)                                                                                                                  docker:default
 => [internal] load build definition from Dockerfile                                                                                                   0.0s
 => => transferring dockerfile: 1.79kB                                                                                                                 0.0s
 => [internal] load metadata for docker.io/library/alpine:3.21                                                                                         0.4s
 => [internal] load .dockerignore                                                                                                                      0.0s
 => => transferring context: 163B                                                                                                                      0.0s
 => [internal] load build context                                                                                                                      0.1s
 => => transferring context: 278.91kB                                                                                                                  0.1s
 => [build  1/11] FROM docker.io/library/alpine:3.21@sha256:b6a6be0ff92ab6db8acd94f5d1b7a6c2f0f5d10ce3c24af348d333ac6da80685                           0.0s
 => CACHED [build  2/11] RUN apk -U upgrade --no-cache && apk add --no-cache     build-base patch tar ca-certificates git        libx11-dev coreutils  0.0s
 => CACHED [build  3/11] RUN curl -sSL https://github.com/ocaml-dune/dune-bin-install/releases/download/v1/install.sh | sh -s 3.19.1 --install-root /  0.0s
 => CACHED [build  4/11] RUN dune --version                                                                                                            0.0s
 => CACHED [build  5/11] WORKDIR /root/ocaml.org                                                                                                       0.0s
 => CACHED [build  6/11] COPY --chown=root . .                                                                                                         0.0s
 => CACHED [build  7/11] RUN ls                                                                                                                        0.0s
 => CACHED [build  8/11] RUN dune pkg lock --verbose                                                                                                   0.0s
 => ERROR [build  9/11] RUN dune build @pkg-install                                                                                                  115.0s
------                                                                                                                                                      
 > [build  9/11] RUN dune build @pkg-install:                                                                                                               
104.7 File "dune.lock/stdune.pkg", line 17, characters 3-73:                                                                                                
104.7 17 |    https://github.com/ocaml/dune/releases/download/3.19.0/dune-3.19.0.tbz)                                                                       
104.7         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
104.7 Error: curl returned an invalid error code 6
104.7        
104.7        
106.5 File "dune.lock/result.pkg", line 11, characters 3-76:
106.5 11 |    https://github.com/janestreet/result/releases/download/1.5/result-1.5.tbz)
106.5         ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
106.5 Error: curl returned an invalid error code 6
106.5        
106.5        
------
Dockerfile:25
--------------------
  23 |     
  24 |     RUN dune pkg lock --verbose
  25 | >>> RUN dune build @pkg-install
  26 |     RUN dune build @install --profile=release #--verbose -j1 --no-buffer
  27 |     
--------------------
ERROR: failed to solve: process "/bin/sh -c dune build @pkg-install" did not complete successfully: exit code: 1
make: *** [Makefile:62: docker] Error 1
```

This seemed to be an issue with the latest Ubuntu that was addressed by

```
DOCKER_BUILDKIT=1 docker build --network=host .
```

cuihtlauac:
Unfortunately using `--network=host` isn't a fix. It's a workaround. Something is wrong in the way dune triggers networking activity. That's what needs to be fixed.

shonfeder:
> Something is wrong in the way dune triggers networking activity.

Oh! That's an interesting claim. This is a good finding if so. But afaik, we have only seen problems on a particular OS version when running specifically on docker. What about this leads us to the conclusions that is a general problem with dune's network requests?

sabine:
Closing and reopening this, in order to not bottleneck progress on my availability to respond to pull requests on my fork. Here's the new PR: https://github.com/ocaml/ocaml.org/pull/3281

PR #3133, opened 2025-05-26
Fix a few typos in the release notes for the compiler

shindere:


sabine (approved):
Thanks @shindere

shindere:
Oh, many thanks for the prompt review / approval / emrge.

I noticed that the Markdown Lint workflow failed. I assume this is
orthogonal to the current PR...

sabine:
> I noticed that the Markdown Lint workflow failed. I assume this is orthogonal to the current PR...

Yes, that's correct. The workflow was added recently, but without going over all the existing files to make them pass the linter. Maybe there's a markdown formatter we can use to make all the files pass.

shindere:
sabine (2025/05/27 00:46 -0700):
> sabine left a comment (ocaml/ocaml.org#3133)
>
> > I noticed that the Markdown Lint workflow failed. I assume this is orthogonal to the current PR...
>
> Yes, that's correct. The workflow was added recently, but without
> going over all the existing files to make them pass the linter. Maybe
> there's a markdown formatter we can use to make all the files pass.

I'm willing to help but it was not obvious to me from the workflow file
how to reproduce locally wha the GHA does.

Issue #3145, opened 2025-06-04
Can't solve with dune package management since min ocaml version bumped to 5.2.0


gridbugs:
Opam is still able to find a solution but dune cannot. This has been a problem since https://github.com/ocaml/ocaml.org/pull/3131. Reverting that PR and dune is able to find a solution, however it chooses `ocaml.4.14.2`. I'm investigating why dune doesn't choose a later version of ocaml.

gridbugs:
The problem seems to be related to the `river` package which is pinned to a specific revision. Dune might be treating this pin differently from opam which has implications on constraints on the versions of other packages but I don't understand the full story yet. Removing `river` from the dependencies makes the problem go away.

gridbugs:
In the pinned version of `river` (https://github.com/aantron/river#476dc945a908a69548bddd267f143a3e5d9c8a1a) there is a dependency on the package `ocamlnet` which depends on `"ocaml" {>= "4.02" & < "5"}` which explains why dune refused to choose a version of ocaml above 5. It doesn't explain why opam _was_ able to find a solution. Did it ignore the pin? When I solve the project with opam I don't see a dependency on `ocamlnet` even though it claims to be using a pinned version of `river`, but maybe opam just takes the package source from the pin and ignores the package metadata (including dependencies)?. Not sure.

Anyway as long as ocaml.org depends on a pinned version of river with `ocamlnet` among its dependencies I think we should not constrain the minimum version of ocaml to > 5.

yawaramin:
The pin was removed in #3178, the project is now depending on upstream `river`.

PR #3146, opened 2025-06-04
Reduce minimum version of ocaml

gridbugs:
The pinned version of the package "river" has a dependency on the package "ocamlnet" which is not compatible with ocaml 5. This fixes an issue where Dune is unable to find a package solution for this project.

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -34,7 +34,7 @@
   "This repository contains the server that serves the official OCaml website at https://ocaml.org.")
  (depends
   (ocaml
-   (>= 5.2.0))
+   (>= 4.14.0))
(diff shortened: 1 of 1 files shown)


gridbugs:
Fixes https://github.com/ocaml/ocaml.org/issues/3145

sabine:
We should be removing the pin on river instead. If necessary, we need to cut a release of river.

Working with one specific version of OCaml instead of supporting multiple is better for build stability and for the contributor experience.

gridbugs:
Can we use this in the short term while we work on removing the river pin. I'm currently unable to build ocaml.org without this patch.

PR #3147, opened 2025-06-04
Add overlay repo to dune-workspace

gridbugs:
When setting the repositories associated with a lockdir the default list over repos is overriden with the new list. Dune package management relies on some forks of some low-level packages that have been patched by the dune developers to work with dune package management. This change adds the overlay back in, fixed at a given commit.


sabine (approved):
Thanks

PR #3176, opened 2025-06-19
refactor(homepage): rename `fac` function snippet to `factorial`

craigfe:
Calling this function `fac` on the homepage is in my opinion a bit cryptic and "inside baseball".

This PR suggests changing it to `factorial` to be more explicit: OCaml is not just for mathematicians after all 😅


cuihtlauac:
Nice suggestion @craigfe

PR #3187, opened 2025-06-26
(data) clarify difference between utop prompt and commands

Firobe:
I've witnessed a beginner confuse the `utop` prompt (`#`) and the prefix for utop commands (like `#quit`). This adds a small clarification that this prefix really has to be typed and is not part of the prompt.


cuihtlauac:
Thanks @Firobe

PR #3201, opened 2025-07-05
Cookbook: Use Basic Authentication to Secure a Route

dinakajoy:
This PR creates a web server that returns a public and secured route
- If user tries to access `/dashboard`, the browser prompts user to enter a username and password
- There is a `mock_authorized_users` to mimic authorized users

### DIFF
--- a/data/cookbook/use-basic-auth-on-web-server/00-dream-server.ml
+++ b/data/cookbook/use-basic-auth-on-web-server/00-dream-server.ml
@@ -1,59 +1,60 @@
----
-packages:
-- name: "dream"
-  tested_version: "1.0.0~alpha8"
-  used_libraries:
-  - dream
-- name: "base64"
-  tested_version: "3.5.1"
-  used_libraries:
-  - base64
-discussion: |
-  This example uses Dream, a simple and type-safe web framework for OCaml.
----
+(* 
+   ⚠️  WARNING: DO NOT USE THIS IN PRODUCTION! ⚠️
+   
+   This is a DEMONSTRATION ONLY showing basic HTTP authentication concepts.
+   
+   This code has SERIOUS SECURITY FLAWS:
+   - Passwords stored in plain text
+   - No password hashing
+   - Vulnerable to timing attacks
+   - No rate limiting
+   - No HTTPS enforcement
+   - Improper credential validation
+   
+   For production use:
+   - Consult OWASP Authentication Cheat Sheet
+   - Use established authentication frameworks
+   - Implement proper password hashing (bcrypt, argon2)
+   - Use HTTPS only
+   - Consider OAuth 2.0, JWT, or session-based auth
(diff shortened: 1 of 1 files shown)


cuihtlauac (dismissed):
Thanks @dinakajoy.

I fear this should not be published without significant changes. We should let those who don't take security seriously devise their own weak code.

yawaramin:
@cuihtlauac wouldn't a production-grade auth system massively complicate this recipe? Aren't Cookbook recipes supposed to be on the simpler side?

cuihtlauac:
> @cuihtlauac wouldn't a production-grade auth system massively complicate this recipe? Aren't Cookbook recipes supposed to be on the simpler side?

I agree, @yawaramin. Here is the problem I see. If we have a simple recipe for this matter, we must accompany it with a warning saying, "Don't do that; it's much more complicated." But then we have a contraction. Publishing the recipe is a claim about simplicity, but the warning says the opposite. That leaves learners clueless, experts irritated, and LLMs poisoned. 

Therefore, I believe there should be no recipe for this.

dinakajoy:
@cuihtlauac 
I am building an authentication system with RBAC using Dream and Brr.
`session storage` or `irmin` can be used for simplicity for this and then linked to the full system (if approved).
What do you think?

yawaramin:
Sounds like an RBAC-based auth system would not be using basic auth, right? In that case we might as well use this PR to remove this 'basic auth' recipe from the cookbook altogether.

sabine (approved):
I think it's still better to have this than not have it because it shows how to manipulate headers, and it's a good place to point people to the OWASP cheatsheet.

data/cookbook/use-basic-auth-on-web-server/00-dream-server.ml
```diff
@@ -0,0 +1,60 @@
...
+   - Passwords stored in plain text
+   - No password hashing
+   - Vulnerable to timing attacks
+   - No rate limiting
+   - No HTTPS enforcement
+   - Improper credential validation
+   
+   For production use:
+   - Consult OWASP Authentication Cheat Sheet
+   - Use established authentication frameworks
```
yawaramin: I would add:

Consider using passwordless, phishing-resistant auth like [WebAuthn](https://ocaml.org/p/webauthn) (passkeys).

sabine: Adding passkey-auth as a recipe to the cookbook sounds like a great idea, in addition.

Having comprehensive docs on stuff like auth is a pretty important signal to people actually building stuff for production.

```suggestion
   - Use established authentication frameworks and schemes (e.g. passwordless, phishing-resistant passkey auth like [WebAuthn](https://ocaml.org/p/webauthn))
```

sabine (approved):
Merging this. Thanks @dinakajoy !

PR #3223, opened 2025-07-18
Add the changelog for opam 2.4.0

kit-ty-kate:
The changelogs for the alphas were in the wrong directory so i simply removed them


sabine (approved):
Thank you @kit-ty-kate!

PR #3258, opened 2025-08-16
Remove non-existent auth package

yawaramin:
Also add a bit more desciption of the `webauthn` package.


yawaramin:
Just fyi https://github.com/robur-coop/u2f - repo is deprecated and package is removed from opam.

PR #3260, opened 2025-08-17
Update planet-sources.yml

aguluman:
This pull request adds a new OCaml-focused blog source to the `data/planet-sources.yml` file.

New blog source added:

* Added Chukwuma Akunyili's blog (`aguluman`) with its RSS feed to the OCaml-only sources list.
* Added RSS feed from blog posts and journal entries at https://fearful-odds.rocks


sabine:
@aguluman thanks!

PR #3272, opened 2025-08-25
Cleanup Blog Syndication

cuihtlauac:
Rename the filter, now `publish_all, ' meaning all posts from the feed will be published.

Make publish_all default to true.

Fix data/planet-source.yml
* Specify publish_all when set to false
* Do not specify publish_all when the name or the URL contains "caml"
* Disable some always error feeds

### DIFF
--- a/data/planet-sources.yml
+++ b/data/planet-sources.yml
@@ -1,6 +1,7 @@
 - id: practicalocaml
   name: Practical OCaml
   url: https://practicalocaml.com/rss/
+  disabled: true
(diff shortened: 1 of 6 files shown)


cuihtlauac:
@hannesm here is a cleanup on blog filtering

hannesm:
That's very nice of you @cuihtlauac. I don't have much understanding of the code and desired semantics (and not enough time at my hands), so I won't be able to carefully review it -- sorry about that.

PR #3311, opened 2025-09-12
Update web.md with vif, ocaml-h1 and httpcats

dinosaure:
Add vif, ocaml-h1 and httpcats as packages for web development


cuihtlauac (approved):
Thanks @dinosaure

PR #3322, opened 2025-09-19
feat: add docs for neovim setup

### LINKED ISSUE #3319: Add Editor Setup Instructions for Neovim
We have editor Setup instructions for VSCode, Emacs, and Vim. However we keep encountering Neovim users who tell us how it was not obvious how to set up OCaml support (too many choices?).

It would be great if we had some instructions in the Get Started > Editor Setup pages.

ETA: relevant file: https://github.com/ocaml/ocaml.org/blob/main/data/tutorials/getting-started/2_00_editor_setup.md

jpoly1219:
Closes #3319 .

Added docs for setting up the LSP server on Neovim.

Add docs for using `nvim-lspconfig` and `vim.lsp`. Both methods are frequently used, and it's a good idea to add instructions for `nvim-lspconfig` because `vim.lsp` only supports Neovim versions 0.11 and above.

### DIFF
--- a/data/tutorials/getting-started/2_00_editor_setup.md
+++ b/data/tutorials/getting-started/2_00_editor_setup.md
@@ -175,29 +175,30 @@ vim.lsp.config['ocamllsp'] = {
 vim.lsp.enable 'ocamllsp'
 ```
 
-You can also use `vim.lsp` with a modular config.
+See `:h lsp-config` for more detail on configuration options.
 
-Assume that your config has the following structure. The internal structure of `lua` does not matter much.
-```text
-.
-├── init.lua
-└── lua
-    ├── custom
-    │   └── plugins
-    │       └── some-plugin.lua
-    └── kickstart
-        ├── health.lua
-        └── plugins
-            └── some-plugin.lua
-```
+#### Modular Config With Runtimepath
 
-Then run the following at the root of your config.
+You can also use `vim.lsp` with a modular config via `runtimepath`. Putting your config table inside `lsp/<some_name>.lua` or `after/lsp/<some_name>.lua` will allow Neovim to search for them automatically.
+
+See `:h runtimepath` for more detail.
+
+Run the following at the root of your config.
 ```text
 mkdir lsp
 touch lsp/ocamllsp.lua
 ```
 
-We now add our LSP configs to `lsp/ocamllsp.lua`...
+Your Neovim config should have the following structure now.
+```text
+.
+├── init.lua
+├── lsp
+│   └── ocamllsp.lua
+└── ...
+```
+
+Add your LSP config to `lsp/ocamllsp.lua`.
 ```lua
 return {
   cmd = { 'ocamllsp' },
@@ -207,7 +208,7 @@ return {
 }
 ```
 
-...and import them in the toplevel `init.lua`.
+Then enable them in the toplevel `init.lua`.
 ```lua
 vim.lsp.enable 'ocamllsp'
 ```
@@ -226,4 +227,6 @@ Add this to your `nvim-lspconfig` setup.
   end,
 },
 ```
+
+There is no need to pass more settings to `setup` because `nvim-lspconfig` provides reasonable defaults. See [here](https://github.com/neovim/nvim-lspconfig/blob/master/lsp/ocamllsp.lua) for more info.
 
--- a/data/tutorials/getting-started/2_00_editor_setup.md
+++ b/data/tutorials/getting-started/2_00_editor_setup.md
@@ -167,8 +167,19 @@ Add this to your toplevel `init.lua`.
 ```lua
 vim.lsp.config['ocamllsp'] = {
   cmd = { 'ocamllsp' },
-  filetypes = { 'ocaml', 'ocaml.interface', 'ocaml.menhir', 'reason' },
-  root_markers = { { 'dune-project', 'dune-workspace' }, '.git' },
+  filetypes = { 
+    'ocaml',
+    'ocaml.interface',
+    'ocaml.menhir',
+    'ocaml.ocamllex',
+    'dune',
+    'reason'
+  },
+  root_markers = {
+    { 'dune-project', 'dune-workspace' },
+    { "*.opam", "esy.json", "package.json" },
+    '.git'
+  },
   settings = {},
 }
 
@@ -202,8 +213,19 @@ Add your LSP config to `lsp/ocamllsp.lua`.
 ```lua
 return {
   cmd = { 'ocamllsp' },
-  filetypes = { 'ocaml', 'ocaml.interface', 'ocaml.menhir', 'reason' },
-  root_markers = { { 'dune-project', 'dune-workspace' }, '.git' },
+  filetypes = { 
+    'ocaml',
+    'ocaml.interface',
+    'ocaml.menhir',
+    'ocaml.ocamllex',
+    'dune',
+    'reason'
+  },
+  root_markers = {
+    { 'dune-project', 'dune-workspace' },
+    { "*.opam", "esy.json", "package.json" },
+    '.git'
+  },
   settings = {},
 }
 ```
--- a/data/tutorials/getting-started/2_00_editor_setup.md
+++ b/data/tutorials/getting-started/2_00_editor_setup.md
@@ -177,9 +177,9 @@ vim.lsp.enable 'ocamllsp'
 
 See `:h lsp-config` for more detail on configuration options.
 
-#### Modular Config With Runtimepath
+#### Using vim.lsp With runtimepath
 
-You can also use `vim.lsp` with a modular config via `runtimepath`. Putting your config table inside `lsp/<some_name>.lua` or `after/lsp/<some_name>.lua` will allow Neovim to search for them automatically.
+You can also move your LSP config to a separate file via `runtimepath` if you'd like to keep your `init.lua` minimal. Putting your config table inside `lsp/<some_name>.lua` or `after/lsp/<some_name>.lua` will allow Neovim to search for them automatically.
 
 See `:h runtimepath` for more detail.
 
--- a/data/tutorials/getting-started/2_00_editor_setup.md
+++ b/data/tutorials/getting-started/2_00_editor_setup.md
@@ -158,33 +158,13 @@ opam install ocaml-lsp-server ocamlformat
 ```
 
 There are two main ways to install and manage LSP servers.
-- A more traditional way is to use `nvim-lspconfig`, which is a package used for this purpose. For more info, `kickstart.nvim` has a great example setup.
-- A newer way is to use the new Neovim LSP API for versions newer than v0.11.0 via `vim.lsp`.
+- A newer, more recommended way is to use the new Neovim LSP API for versions newer than v0.11.0 via `vim.lsp`.
+- A more traditional way is to use `nvim-lspconfig`. For more info, `kickstart.nvim` has a great example setup.
 
-Using the Lazy package manager with `nvim-lspconfig`:
+### Using vim.lsp:
+
+Add this to your toplevel `init.lua`.
 ```lua
-require('lazy').setup({
-  -- other packages...
-  {
-    'neovim/nvim-lspconfig',
-    config = function()
-      vim.api.nvim_create_autocmd('LspAttach', {
-        -- your keybindings here...
-      })
-
-      -- this part may contain your existing LSP setup that play along nicely with Mason...
-
-      -- add this line specifically for OCaml
-      require('lspconfig').ocamllsp.setup {}
-    end,
-  },
-  -- other packages...
-})
-```
-
-Using `vim.lsp`:
-```lua
--- add this to your toplevel init.lua
 vim.lsp.config['ocamllsp'] = {
   cmd = { 'ocamllsp' },
   filetypes = { 'ocaml', 'ocaml.interface', 'ocaml.menhir', 'reason' },
@@ -211,29 +191,39 @@ Assume that your config has the following structure. The internal structure of `
             └── some-plugin.lua
 ```
 
-Then run the following.
+Then run the following at the root of your config.
 ```text
-cd lua
 mkdir lsp
-touch lsp/init.lua
-touch lsp/servers.lua
+touch lsp/ocamllsp.lua
 ```
 
-We now add our LSP configs to `lsp/servers.lua`, export them from `lsp/init.lua`, and import them in the toplevel `init.lua`.
+We now add our LSP configs to `lsp/ocamllsp.lua`...
 ```lua
--- path/to/config/lua/lsp/servers.lua
-vim.lsp.config['ocamllsp'] = {
+return {
   cmd = { 'ocamllsp' },
   filetypes = { 'ocaml', 'ocaml.interface', 'ocaml.menhir', 'reason' },
   root_markers = { { 'dune-project', 'dune-workspace' }, '.git' },
   settings = {},
 }
-
-vim.lsp.enable 'ocamllsp'
-
--- path/to/config/lua/lsp/init.lua
-require("lsp.servers")
-
--- path/to/config/init.lua
-require("lsp")
 ```
+
+...and import them in the toplevel `init.lua`.
+```lua
+vim.lsp.enable 'ocamllsp'
+```
+
+### Using nvim-lspconfig
+
+Add this to your `nvim-lspconfig` setup.
+```lua
+{
+  'neovim/nvim-lspconfig',
+  config = function()
+    -- rest of config...
+
+    -- add this line specifically for OCaml
+    require('lspconfig').ocamllsp.setup {}
+  end,
+},
+```
+
--- a/data/tutorials/getting-started/2_00_editor_setup.md
+++ b/data/tutorials/getting-started/2_00_editor_setup.md
@@ -145,3 +145,95 @@ opam user-setup install
 - Type `:MerlinTypeOf` and press <kbd>Enter</kbd>.
 - The type information will be displayed in the command bar.
 Other Merlin commands for Vim are available and you can checkout their usage on the [Merlin official documentation for Vim](https://ocaml.github.io/merlin/editor/vim/).
+
+## Neovim
+
+Neovim comes with an LSP client.
+
+One note here is that is that `ocaml-lsp-server` is sensitive to versioning, and often does not play well with the sometimes outdated sources in Mason, a popular package manager for language services. We recommend you install the LSP server directly in the switch, and pointing your Neovim config to use that.
+
+To install the LSP server and the formatter, run the following.
+```shell
+opam install ocaml-lsp-server ocamlformat
+```
+
+There are two main ways to install and manage LSP servers.
+- A more traditional way is to use `nvim-lspconfig`, which is a package used for this purpose. For more info, `kickstart.nvim` has a great example setup.
+- A newer way is to use the new Neovim LSP API for versions newer than v0.11.0 via `vim.lsp`.
+
+Using the Lazy package manager with `nvim-lspconfig`:
+```lua
+require('lazy').setup({
+  -- other packages...
+  {
+    'neovim/nvim-lspconfig',
+    config = function()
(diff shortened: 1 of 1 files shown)


data/tutorials/getting-started/2_00_editor_setup.md
```diff
@@ -145,3 +145,95 @@ opam user-setup install
...
+- A more traditional way is to use `nvim-lspconfig`, which is a package used for this purpose. For more info, `kickstart.nvim` has a great example setup.
+- A newer way is to use the new Neovim LSP API for versions newer than v0.11.0 via `vim.lsp`.
+
+Using the Lazy package manager with `nvim-lspconfig`:
+```lua
+require('lazy').setup({
+  -- other packages...
+  {
+    'neovim/nvim-lspconfig',
+    config = function()
```
syaiful6: here, I think we need to set `mason = false`. Otherwise it will not work, at least this is how experienced in the past.

jpoly1219: Oh interesting, my current configuration uses Mason alongside ocamllsp. I don't think I've ran into issues compared yet.

syaiful6: yeah, this is probably because ocamllsp is version sensitive as you mentioned, probably at that time the version installed in mason not compatible. So, I decided to install ocaml-lsp-server for each project.

anyway, I think this is look good to me.

data/tutorials/getting-started/2_00_editor_setup.md
```diff
@@ -145,3 +145,95 @@ opam user-setup install
...
+
+We now add our LSP configs to `lsp/servers.lua`, export them from `lsp/init.lua`, and import them in the toplevel `init.lua`.
+```lua
+-- path/to/config/lua/lsp/servers.lua
+vim.lsp.config['ocamllsp'] = {
+  cmd = { 'ocamllsp' },
+  filetypes = { 'ocaml', 'ocaml.interface', 'ocaml.menhir', 'reason' },
+  root_markers = { { 'dune-project', 'dune-workspace' }, '.git' },
+  settings = {},
+}
```
boltlessengineer: I think it would be more idiomatic in Neovim to put lsp server config under `lsp/ocamllsp.lua` rather than `lua/lsp/*.lua`.

Now even less code and they are in right place rather than imported manually.

`:h 'runtimepath'`

```lua
-- path/to/config/lsp/ocamllsp.lua
return {
  cmd = { 'ocamllsp' },
  filetypes = { 'ocaml', 'ocaml.interface', 'ocaml.menhir', 'reason' },
  root_markers = { { 'dune-project', 'dune-workspace' }, '.git' },
  settings = {},
}

-- init.lua
vim.lsp.enable('ocamllsp')
```

boltlessengineer (changes requested):
Do we really need to explain about `lazy.nvim` and `neovim/nvim-lspconfig` here? They aren't really a dependency for proper setup so they should be at least less preferred here.

I know many people are still using them, but I think they might introduce too much complexity considering the purpose of this document.

`nvim-lspconfig` is already using `vim.lsp` internally afaik so I think just directly showing `vim.lsp` based config here can actually be enough.

We can let users switch to `nvim-lspconfig` for even shorter one-liner setup _if_ they have an existing neovim config with `lspconfig`.

```
-- this should be enough for them. Others are just boilerplates.
require('lspconfig').ocamllsp.setup()
```

data/tutorials/getting-started/2_00_editor_setup.md
boltlessengineer: Isn't that option related to `LazyVim` and not `lazy.nvim`?

jpoly1219:
@boltlessengineer Thank you for the review! I made the necessary changes. Let me know if you'd prefer to see an even shorter version.

Your point makes a lot of sense in that we shouldn't cover the full setup of package managers and setup, because that's out of scope of this guide. And I like your comment on the `vim.lsp` changes.

I think the notion of adding boilerplate was to give more context to the users, as I used to struggle a lot with terse setup documentation. Call it skill issue but I would've appreciated having a minimal boilerplate in a setup guide, as it's a lot easier to make sense of where to actually put the config😅 I'd rather not assume that all Neovim users are wizards who know the full ins and outs of their config.

syaiful6 (approved):

data/tutorials/getting-started/2_00_editor_setup.md
```diff
@@ -145,3 +145,85 @@ opam user-setup install
...
+```lua
+return {
+  cmd = { 'ocamllsp' },
+  filetypes = { 'ocaml', 'ocaml.interface', 'ocaml.menhir', 'reason' },
+  root_markers = { { 'dune-project', 'dune-workspace' }, '.git' },
+  settings = {},
+}
+```
+
+...and import them in the toplevel `init.lua`.
```
boltlessengineer: You are missunderstanding `lsp/` folder.
It's a `runtimepath` file. Read automatically by Neovim on startup. It's not a lua module.

When calling `vim.lsp.enable`, we aren't *importing* it. We've already done configuring by placing the lsp config table in `<rtp>/lua/<name>.lua` files or by calling/setting `vim.lsp.config` directly.

data/tutorials/getting-started/2_00_editor_setup.md
```diff
@@ -145,3 +145,85 @@ opam user-setup install
...
+
+Add this to your `nvim-lspconfig` setup.
+```lua
+{
+  'neovim/nvim-lspconfig',
+  config = function()
+    -- rest of config...
+
+    -- add this line specifically for OCaml
+    require('lspconfig').ocamllsp.setup {}
```
boltlessengineer, line 248: Maybe we can explain why we don't need the configuration step with lspconfig here?

> lspconfig ships with default ocamllsp config preloaded. that's why we don't need manual configuration step here. You can see their builtin configuration table [here](https://github.com/neovim/nvim-lspconfig/blob/master/lsp/ocamllsp.lua)

boltlessengineer (changes requested):
While I appreciate the change on `nvim-lspconfig` section, the `vim.lsp` section... got worse. I should have explained it more clearly.

There are so many ways to *configure* a LSP with `vim.lsp`.

- base configuration defined for the `*` name
- configuration defined in `<rtp>/lsp/<name>.lua`
- configuration defined by `vim.lsp.config()` in anywhere else

We cannot explain all of these.

So I think separating explanations of configuration and activation(enabling) can be clear enough.

1. explain that you should 1) configure and 2) enable the language server.
2. show the single `init.lua` example as that is best for copy-pasting
3. show that you can put configuration table under `<rtp>/lsp/<name>.lua`.\
  **NOTE: this is not a lua *module*!! it's a rtp file**
4. link to `:h lsp-config` so that users can learn about this from official document.
5. put *common* `lazy.nvim` way to do this

Or remove the part 2 and only show runtimepath (`lsp/` folder) way. Because that's also what `nvim-lspconfig` is using internally.

So:

> ... explain that you should configure and enable the server.

### Using vim.lsp:

See `:h lsp-config` for detailed information.

```
-- in your `init.lua` file

-- configure ocamllsp
vim.lsp.config('ocamllsp', {
  -- ...
})

-- enable ocamllsp
vim.lsp.enable('ocamllsp')
```

#### placing ocamllsp configuration in separate file

Instead of calling `vim.lsp.config`, you can put ocamllsp configuration table in separate file. Neovim will automatically search for `lsp/<name>.lua` or `after/lsp/<name>.lua` files and merge them all in single config. (see `:h lsp-config` and `:h 'runtimepath'`)
```
-- in `lsp/ocamllsp.lua` (not `lua/lsp/ocamllsp.lua`)
return {
  -- ...
}
```

Now you can enable ocamllsp from your `init.lua`

```
vim.lsp.enable('ocamllsp')
```

> ... show `lazy.nvim` and `nvim-lspconfig` way

data/tutorials/getting-started/2_00_editor_setup.md
syaiful6: @boltlessengineer this correct, my understanding is he mean to create `lsp/` folder in root config.  It will get installed to `~/.config/nvim/lsp/ocamllsp.lua`.

jpoly1219: @boltlessengineer should've been more pedantic about the wording here, mb. What I meant by import was to call require.

jpoly1219: I don't think I said to add the lsp configs to `lua`? But I can see how giving more context on the config folder structure can be potentially misleading.

boltlessengineer: I'm saying that we aren't requiring `lsp/ocamllsp.lua` by calling `vim.lsp.enable()`.
`vim.lsp.config` and `vim.lsp.enable` are completely different things.

Both term 'import' and 'require' doesn't fit here.

What `vim.lsp.enable` is doing is just: Setup that language server with configured value which is pre-merged `vim.lsp.config` table. It isn't about loading & merging configs, but about setting up a required autocmds. (to auto-start language server on filetypes)

When someone wants to *configure* a server, defining a table in `<rtp>/lsp/<name>.lua` is just enough. It will be auto-merged whenever you indexed the `vim.lsp.config` like `vim.lsp.config['ocamllsp']`

check this out:

- write `lsp/ocamllsp.lua` and do nothing else, don't enable it yet
- open neovim and run `:=vim.lsp.config['ocamllsp']
- see it already contains the merged config.

jpoly1219:
@boltlessengineer made more edits, lmk if it looks better.

data/tutorials/getting-started/2_00_editor_setup.md
```diff
@@ -175,29 +175,30 @@ vim.lsp.config['ocamllsp'] = {
...
-    └── kickstart
-        ├── health.lua
-        └── plugins
-            └── some-plugin.lua
-```
+#### Modular Config With Runtimepath
+
+You can also use `vim.lsp` with a modular config via `runtimepath`. Putting your config table inside `lsp/<some_name>.lua` or `after/lsp/<some_name>.lua` will allow Neovim to search for them automatically.
+
+See `:h runtimepath` for more detail.
```
boltlessengineer: The term "modular" might confuse users with typical *lua modules*. They aren't. Lua modules in Neovim config only lives in `lua/` directory and `lsp/*.lua` are just plain standalone lua files loaded by neovim directly.

I think "multi-file" can be better for clarity.

data/tutorials/getting-started/2_00_editor_setup.md
```diff
@@ -207,7 +208,7 @@ return {
 }
 ```
 
-...and import them in the toplevel `init.lua`.
+Then enable them in the toplevel `init.lua`.
```
boltlessengineer, line 233: another nit: Can we explain what `vim.lsp.enable()` do here? It will help users understand the difference between configuration and enabling.

> Then enable them in the toplevel `init.lua`. This will make Neovim to automatically start the ocamllsp for associated filetypes.
..codeblock..
> To start ocamllsp manually, you can use `vim.lsp.start`. (see `:h vim.lsp.start`)
> ```
> :lua vim.lsp.start(vim.lsp.config['ocamllsp'])
> ```

Not sure if we want to mention `vim.lsp.start` here. I think it can help users to understand what `vim.lsp.enable` actually does, but it can be yet another code snippet they should read and understand.

boltlessengineer (changes requested):
Except for some nits on wording, all seems good now!

I'm curious if mentioning `vim.lsp.start` can help here. Do you think it can help? or is it too much?

data/tutorials/getting-started/2_00_editor_setup.md
jpoly1219: I'm not sure if that distinction is strictly necessary 🤔  I think most people understand what a modular config means in general and are more familiar with that term. I've seen lots of modular config out there but not a single multi-file config.

data/tutorials/getting-started/2_00_editor_setup.md
jpoly1219: I feel like this is a bit out of scope for the purpose of this doc. I think the curious readers will go out and research this further on their own.

data/tutorials/getting-started/2_00_editor_setup.md
boltlessengineer: People don't use runtimepath much for lsp/general config because of historical reasons.
But now they are available and is completely different thing from structuring everything under `lua/` directory.

Even though we clearly stated that they should put that file under `lsp/` and not `lua/lsp/`, I think calling it as a "module" or saying this structure "modular" might confuse users between runtimepath and lua-modules.

jpoly1219: @boltlessengineer I'll make the changes and request for another review 👍

boltlessengineer (approved):
LGTM!

data/tutorials/getting-started/2_00_editor_setup.md
```diff
@@ -145,3 +145,88 @@ opam user-setup install
...
+- A newer, more recommended way is to use the new Neovim LSP API for versions newer than v0.11.0 via `vim.lsp`.
+- A more traditional way is to use `nvim-lspconfig`. For more info, `kickstart.nvim` has a great example setup.
+
+### Using vim.lsp:
+
+Add this to your toplevel `init.lua`.
+```lua
+vim.lsp.config['ocamllsp'] = {
+  cmd = { 'ocamllsp' },
+  filetypes = { 'ocaml', 'ocaml.interface', 'ocaml.menhir', 'reason' },
```
edwintorok: `ocamllsp.lua` from `nvim-lspconfig` also adds `ocaml.ocamllex`, and `dune` to this list, but it does it in an awkward way, the filetype is actually called `ocamlinterface` and it gets mapped to `ocaml.interface` using `get_language_id`.

data/tutorials/getting-started/2_00_editor_setup.md
```diff
@@ -145,3 +145,88 @@ opam user-setup install
...
+├── lsp
+│   └── ocamllsp.lua
+└── ...
+```
+
+Add your LSP config to `lsp/ocamllsp.lua`.
+```lua
+return {
+  cmd = { 'ocamllsp' },
+  filetypes = { 'ocaml', 'ocaml.interface', 'ocaml.menhir', 'reason' },
```
edwintorok: same as above, this filetype list could probably be longer

data/tutorials/getting-started/2_00_editor_setup.md
edwintorok: Maybe instead of calling `lsp.enable` with just `ocamllsp` we could show that you only need to call this once, with a list of languages (OCaml being just one of them), but it is fine as it is too.
Maybe add a link to the official NeoVim LSP docs? https://neovim.io/doc/user/lsp.html

data/tutorials/getting-started/2_00_editor_setup.md
```diff
@@ -145,3 +145,88 @@ opam user-setup install
...
+    -- rest of config...
+
+    -- add this line specifically for OCaml
+    require('lspconfig').ocamllsp.setup {}
+  end,
+},
+```
+
+There is no need to pass more settings to `setup` because `nvim-lspconfig` provides reasonable defaults. See [here](https://github.com/neovim/nvim-lspconfig/blob/master/lsp/ocamllsp.lua) for more info.
+
```
edwintorok, line 254: Should we explain about tree-sitter? (with `nvim-treesitter`)?

```
local treesitter_langs = {
 'menhir', 'ocaml', 'ocaml_interface', 'ocamllex'
}

require('nvim-treesitter').install(treesitter_langs)
```

It isn't strictly required, although might help with performance on large files, before the LSP finishes parsing it.
Maybe leave it out for now, I'll do some experiments with using just the LSP, and see how well that works.

data/tutorials/getting-started/2_00_editor_setup.md
boltlessengineer: We are already mentioning it. `:h lsp-config` should be enough.

edwintorok (commented):
Thanks, this is a good start!

We can perhaps have a separate section that shows the more advanced features (inlay hints, etc.), I can try to prepare a separate PR about that.

data/tutorials/getting-started/2_00_editor_setup.md
jpoly1219: Sounds good. Let me know if you have more updates 👍

jpoly1219:
@edwintorok I went ahead and added more filetypes and root path. Could you take another quick look at it please?

I'll leave the settings field empty until you play around with more advanced lsp/merlin features.

data/tutorials/getting-started/2_00_editor_setup.md
```diff
@@ -145,3 +145,110 @@ opam user-setup install
...
+Add your LSP config to `lsp/ocamllsp.lua`.
+```lua
+return {
+  cmd = { 'ocamllsp' },
+  filetypes = { 
+    'ocaml',
+    'ocaml.interface',
+    'ocaml.menhir',
+    'ocaml.ocamllex',
```
edwintorok, line 220: This one doesn't actually seem to work on NeoVim 0.11.4 and ocamllsp 1.23.0.
ocamllsp says 'unsupported file extension'.
But removing it from this list doesn't work either, because then it falls back to the `ocaml` filetype on the `.mll` file, and calls `ocamllsp` anyway.

The rest seems to work.

edwintorok (approved):

data/tutorials/getting-started/2_00_editor_setup.md
edwintorok: Ah `ocaml.ocamllex` is to future-proof it for https://github.com/ocaml/vim-ocaml/pull/61 (which isn't merged yet)

data/tutorials/getting-started/2_00_editor_setup.md
edwintorok: I figured out how to configure tree-sitter for ocamllex, but it is quite complicated: https://github.com/ocaml/vim-ocaml/pull/61#issuecomment-3315360034.

For now I'd leave out tree-sitter from the official docs, until that PR is merged (and an equivalent PR is merged into NeoVim).

The workaround is (fixes both picking the right tree-sitter syntax and the LSP errors):
```lua
vim.filetype.add({
    extension = {
        mll = 'ocamllex',
        mly = 'menhir',
        mli = 'ocamlinterface'
    }
})
```

followed by the usual tree-sitter config:
```lua
local treesitter_langs = {
 'menhir', 'ocaml', 'ocaml_interface', 'ocamllex'
}

require('nvim-treesitter').install(treesitter_langs)
```

(This doesn't require a new release of NeoVim to work, but if your tree-sitter grammar fails to install you'll be without syntax highlighting, so I wouldn't propose it in the official docs, even though I'll start using it myself).

data/tutorials/getting-started/2_00_editor_setup.md
edwintorok: https://github.com/ocaml/ocaml.org/pull/3322#discussion_r2365869097 workaround, but `ocaml.ocamllex` is fine as it is in the suggested doc, even though it doesn't work yet.

data/tutorials/getting-started/2_00_editor_setup.md
```diff
@@ -145,3 +145,110 @@ opam user-setup install
...
+    'dune',
+    'reason'
+  },
+  root_markers = {
```
edwintorok, line 178: Nice!

I proposed this simplification to the [upstream `nvim-lspconfig`](https://github.com/neovim/nvim-lspconfig/pull/4098), which was using a `root_dir` function and `lspconfig.util` instead (which appears deprecated).

sabine (approved):
Hey, thanks for adding this @jpoly1219 and thanks for reviewing @edwintorok, @boltlessengineer, @syaiful6!

Since I haven't used neovim, I can't say anything about the instructions, but it's great to have something to help guide new OCaml users on neovim now!

jpoly1219:
@sabine glad I could help :)

PR #3333, opened 2025-09-30
Routing for assetlinks.json

mtelvers:
This is a fix for #3168. The PR used `Dream.from_filesystem` which works locally as that file is present, but when deployed, `assets/**` is loaded using `ocaml-crunch` into an in-memory filesystem and the content of `assets/**` are not directly contained in the Docker container.

 The obvious fix for this is to use the static loader:
```
Dream.get "/.well-known/assetlinks.json" (Dream.static ~loader:asset_loader "assetlinks.json");
```
However, the `asset_loader` ignores the root path when calling out to `Ocamlorg_static.Asset.read`. An update like this would address that, but I'm not familiar enough with the codebase to propose it.

```
  let asset_loader =
    Static.loader
      ~read:(fun root path ->
        let full_path = if root = "" then path else root ^ "/" ^ path in
        Ocamlorg_static.Asset.read full_path |> Lwt.return)
      ~digest:(fun root path ->
        let full_path = if root = "" then path else root ^ "/" ^ path in
        Option.map Dream.to_base64url (Ocamlorg_static.Asset.digest full_path))
      ~not_cached:[ "robots.txt"; "/robots.txt" ]
```

The proposed change here is to make the minimum change.

cc @samoht @cuihtlauac


cuihtlauac (approved):
Thanks @mtelvers, that looks good to me (and much better than my initial patch)

src/ocamlorg_web/lib/router.ml
```diff
@@ -147,7 +147,10 @@ let router t =
   Dream.router
     [
       Dream.get "/.well-known/assetlinks.json"
-        (Dream.from_filesystem "asset/.well-known" "assetlinks.json");
+        (fun _request ->
+          match Ocamlorg_static.Asset.read "assetlinks.json" with
+          | Some content -> Dream.respond ~headers:[("Content-Type", "application/json")] content
+          | None -> Dream.respond ~status:`Not_Found "Not found");
```
cuihtlauac:
```suggestion
      Dream.get "/.well-known/assetlinks.json" (fun _request ->
          match Ocamlorg_static.Asset.read "assetlinks.json" with
          | Some content ->
              Dream.respond
                ~headers:[ ("Content-Type", "application/json") ]
                content
          | None -> Dream.respond ~status:`Not_Found "Not found");
```

cuihtlauac (changes requested):
Fix formatting

cuihtlauac (approved):

PR #3344, opened 2025-10-06
Changelog: add ocaml.5.4.0~beta2 and ocaml.5.4.0~rc1

sabine:
@Octachron I'm adding the two announcements you made on Discuss. Markdown reconstructed by Claude


Octachron:
@sabine , don't hesitate to tell me if  you would like that I send the changelog entry for the prerelease version of the compilers. I only stopped doing this when such entries were removed from ocaml.org.

PR #3347, opened 2025-10-06
Changelog: add merlin 5.6-503 5.6-504 and ocaml-lsp 1.23.1 1.24.0

sabine:
@xvw Thanks for making an announcement on Discuss.

Since the current structure of the OCaml Changelog (with automated scraping of releases from GitHub) can't yet accomodate joint announcements of multiple tools, I split the content up like this.


xvw:
Nice! Thanks for the reactivity !

PR #3343, opened 2025-10-06
Changelog: add release announcement for opam-publish 2.6.0

sabine:
@kit-ty-kate would that work as an announcement for `opam-publish.2.6.0` for the OCaml Changelog?

I saw there was no release announcement on Discuss I could copy from. Maybe we want to post this on Discuss as well?

Feel free to update / change as needed, or just give me a thumbs up for publication. :heart:

PR #3363, opened 2025-10-09
Announce opam-publish 2.6.0 and 2.7.0

kit-ty-kate:
As a replacement for https://github.com/ocaml/ocaml.org/pull/3343


sabine (approved):
Thanks @kit-ty-kate, this is much clearer!

PR #3371, opened 2025-10-12
Fix adjacency list to match graph example

feefs:
The adjacency list representation of the graph (https://ocaml.org/media/problems/graph1.gif) is missing an entry. Follow-up to #2769


sabine (approved):
Thanks @feefs

PR #3382, opened 2025-10-26
doc: remove unnecessary information in PPX tutorial

syedtaz:
I got a little tripped up by this description of environment variables in the PPX tutorial. PPXs are tagged as an advanced topic and I don't think we need to repeat that the preprocessor runs at compile time three times. Also, I don't think USER is defined on Windows so this statement that "this code should work on most systems without configuration" isn't true.


sabine (approved):
Thank you @syedtaz, I think this indeed reads clearer with this edit making it less verbose.

PR #3413, opened 2025-12-10
Update LexiFi job offer

nojb:
We recently started a new recruitment campaign for an OCaml dev, so we would like to "refresh" the entry in this page. Thanks!


cuihtlauac:
Thanks @nojb

PR #3452, opened 2026-01-16
Fix incorrect plural feets → feet

jorgenboganes:
Fixes a spelling error in the documentation where "feets" was used instead of the correct plural "feet".

### Changes
- `data/tutorials/getting-started/1_01_a_tour_of_ocaml.md`: corrected "feets" → "feet"


cuihtlauac:
Thanks @jorgenboganes

PR #3462, opened 2026-01-26
governance: improve the security team contacts

hannesm:
/cc @nojb @cuihtlauac

### DIFF
--- a/data/governance.yml
+++ b/data/governance.yml
@@ -126,12 +126,18 @@ teams:
     for high-impact vulnerabilities, and maintains the advisory database
     and associated tooling.
   contacts:
-  - name: Mailing List
+  - name: Website
+    link: https://ocaml.org/security
+    kind: website
+  - name: Public Announcement Mailing List
     link: https://sympa.inria.fr/sympa/info/ocsf-ocaml-security-announcements
     kind: email
   - name: GitHub
     link: https://github.com/ocaml/security-advisories
     kind: github
+  - name: Private email
+    link: security@ocaml.org
+    kind: email
   members:
   - name: Hannes Mehnert
     github: hannesm
--- a/data/governance.yml
+++ b/data/governance.yml
@@ -129,15 +129,6 @@ teams:
   - name: Website
     link: https://ocaml.org/security
     kind: website
-  - name: Public Announcement Mailing List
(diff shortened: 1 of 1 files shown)


cuihtlauac:
To avoid duplication, should we remove the list of names from ocaml.org/security and replace it with a link to ocaml.org/governance/security?

data/governance.yml
```diff
@@ -126,12 +126,18 @@ teams:
...
+    kind: website
+  - name: Public Announcement Mailing List
     link: https://sympa.inria.fr/sympa/info/ocsf-ocaml-security-announcements
     kind: email
   - name: GitHub
     link: https://github.com/ocaml/security-advisories
     kind: github
+  - name: Private email
+    link: security@ocaml.org
+    kind: email
```
nojb: Should we put the link to `https://ocaml.org/security` only and let people go there to learn about the other ways to contact/interact with the security team? That way we duplicate the other information in fewer places.

nojb:
> To avoid duplication, should we remove the list of names from ocaml.org/security and replace it with a link to ocaml.org/governance/security?

I was actually suggesting the opposite. The `/security` page is richer while the `/governance` cards are more limited and more of an "overview", which is why I think it makes sense to have the `/governance` page point to `/security`, rather than the other way around.

data/governance.yml
```diff
@@ -126,12 +126,18 @@ teams:
     for high-impact vulnerabilities, and maintains the advisory database
     and associated tooling.
   contacts:
-  - name: Mailing List
+  - name: Website
+    link: https://ocaml.org/security
+    kind: website
+  - name: Public Announcement Mailing List
```
cuihtlauac: @nojb 
> I was actually suggesting the opposite. The /security page is richer while the /governance cards are more limited and more of an "overview", which is why I think it makes sense to have the /governance page point to /security, rather than the other way around.

You are right, such a link is needed. The above will create it. I was suggesting having two way links :-) Sorry I wasn't clear.

data/governance.yml
hannesm: I pushed 867ffa86, and think this is good then?

hannesm:
I'm fine to have the team data duplicated. I think it is valuable to have it nicely layed out in the governance. I don't think it is much work to keep them in sync.

nojb (approved):
LGTM

PR #3238, opened 2025-08-03
initial steps towards I18n of tutorial docs

mt-caret:
As a first step towards supporting multi-lingual guides, I've added "language" as a new frontmatter tag, and threaded it through dream so that we can dispatch to different markdown files based on language.

I've also added a language selector which shows up when there are multiple language choices, and added a Japanese translation of a single markdown file (`data/tutorials/getting-started/1_00_install_OCaml.ja.md`) as something to start us off (and also to demonstrate the functionality).
Screenshot 2026-02-08 at 11 31 50 AM
Screenshot 2026-02-08 at 11 31 59 AM

Future steps:
- translate the rest of the guides to Japanese
- set up some automated job which triggers on updates to English markdown files to propose updates to the Japanese versions via LLMs.

PR #3489, opened 2026-02-08
More japanese translations

mt-caret:


cuihtlauac:
What's the relationship between this and #3238?

PR #3491, opened 2026-02-10
Fix typos and grammar across tutorials

bbatsov:
I went through all the tutorials and fixed the typos and grammar issues I spotted. About 100 fixes across 41 files covering all four tutorial sections (getting-started, guides, language, platform).

Here's the gist of what's fixed:

- Misspellings: "specifcations", "suffient", "downnloaded", "fucntion's", "libaries", "Hashtabl", etc.
- Wrong words: "must" instead of "most", "status" instead of "stratus", "that" instead of "than", "form" instead of "from"
- Subject-verb disagreement: "engines fixes", "break hint lead", "syntax makes" (with plural subject), etc.
- Missing words: missing articles, prepositions, subjects ("In OCaml, modules...", "a local opam switch", etc.)
- Doubled words: "the the", "operator operator", "function function"
- Wrong function references: `Option.fold` should be `Option.bind` in the options tutorial
- Wrong type: `String.compare` returns `int`, not `bool` (functors tutorial)
- Punctuation: missing periods, extra commas, spacing issues

### DIFF
--- a/data/tutorials/language/3ds_04_hashtbl.md
+++ b/data/tutorials/language/3ds_04_hashtbl.md
@@ -22,7 +22,7 @@ information at a nearly instantaneous linear time complexity (O(1)).
 A hash table data structure achieves efficient reads and writes by employing a
 hashing function that converts the key of a key/value pair into an
 algorithmically unique "fingerprint" known as a hash. OCaml has a built-in
-hashing function within the `Hashtabl` module that is available for each key
+hashing function within the `Hashtbl` module that is available for each key.
 The `Hashtbl` module implements an efficient, mutable lookup table.
 
 ## Creating a Polymorphic Hash Table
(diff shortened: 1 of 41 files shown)


data/tutorials/language/3ds_04_hashtbl.md
```diff
@@ -22,7 +22,7 @@ information at a nearly instantaneous linear time complexity (O(1)).
 A hash table data structure achieves efficient reads and writes by employing a
 hashing function that converts the key of a key/value pair into an
 algorithmically unique "fingerprint" known as a hash. OCaml has a built-in
-hashing function within the `Hashtabl` module that is available for each key
+hashing function within the `Hashtbl` module that is available for each key.
```
cuihtlauac, line 25: Nice catch!

cuihtlauac (approved):
Very nice @bbatsov, thanks a lot

PR #3497, opened 2026-02-10
Update URL for Thomas Letan’s Blog in YAML

lthms:
I now have a OCaml-specific RSS feed that will be more relevant for OCaml Planet.


cuihtlauac (approved):
Thanks @lthms

PR #3496, opened 2026-02-10
Remove dead code: unused function and stale build dep

bbatsov:
I noticed a couple of small bits of dead code while poking around the codebase:

- **`Url.blog_post`** in `src/global/url.ml` — this function isn't called anywhere. `local_blog` is used, but `blog_post` appears to be a leftover.
- **`planet-local-blogs` source_tree dep** in the root `dune` file — references `data/planet-local-blogs/`, which doesn't exist. Likely from an earlier iteration of the planet feed setup.

Neither removal affects any functionality. Just tidying things up.

PR #3495, opened 2026-02-10
Fix typos and lint across smaller data directories

bbatsov:
## Summary
- Added `siblings_only: true` to MD024 config to allow duplicate subheadings under different parents
- Fixed markdown lint issues (trailing spaces, blank lines, heading levels) across ~50 files
- Fixed typos in conferences ("Workskop" → "Workshop", "Ljublijana, Solvenia" → "Ljubljana, Slovenia"), books ("prerequiresite" → "prerequisite"), industrial users ("analyzis" → "analysis", "Ocaml" → "OCaml"), tool pages ("lifecyle" → "lifecycle", "wether" → "whether"), and more
- Added language specifiers to fenced code blocks in platform roadmap (MD040)
- Converted bold emphasis used as headings to proper `#####` headings in platform roadmap (MD036)
- Removed emphasis from persona taglines in platform users doc (MD036) — the linter flags standalone paragraphs that are entirely italic/bold, since it can't distinguish decorative emphasis from emphasis-as-heading

## Test plan
- [x] `npx markdownlint-cli2` reports 0 errors across all targeted directories

PR #3494, opened 2026-02-10
Fix typos and lint across exercise files

bbatsov:
## Summary
- Fixed markdown lint issues across all 87 exercise files (headings, blank lines, broken links)
- Converted `# Solution` / `# Statement` to `## Solution` / `## Statement` (MD025)
- Fixed several typos in prose and comments: "need so" → "need to", "but and" → "and", stray backtick code, "i.e" → "i.e.", "remember than" → "remember that", "a functions" → "a function", "acyclic path" → "acyclic paths"
- Renamed `084_minimal_splanning_tree.md` → `084_minimal_spanning_tree.md`
- Removed broken cross-exercise link fragments

## Test plan
- [x] `npx markdownlint-cli2` reports 0 errors on `data/exercises/*.md`

PR #3493, opened 2026-02-10
Fix typos and markdown lint in root-level and developer docs

bbatsov:
## Summary

- Fix typos across root-level docs: "successuful" → "successful", "Maintnainers" → "Maintainers", "solultion" → "solution", "straighforward" → "straightforward", "Ror" → "For", duplicate "https://https://", "mixture or" → "mixture of", "Checkout out" → "Check out", etc.
- Fix markdown lint issues: add missing code block languages, remove trailing whitespace, fix broken link fragments, replace non-descriptive `[here]` link text, fix multiple H1 headings, add trailing newlines, remove extra blank lines
- Update `.markdownlint-cli2.jsonc` to add `siblings_only` to MD024 config so repeated subsection headings under different parents are allowed

PR #3500, opened 2026-02-11
Mechanical markdown lint fixes extracted from #3493, #3494, #3495

cuihtlauac:
## Summary

Extracts the purely mechanical, machine-verifiable lint fixes from @bbatsov's PRs #3493, #3494, and #3495 into separate commits — one commit per kind of change.

This approach was suggested by @voodoos in [#3496 (comment)](https://github.com/ocaml/ocaml.org/pull/3496#issuecomment-3884665735):

> They do mix multiple changes, and it would have been nice to have one commit per such change instead of a single one

Each commit message includes a shell pipe that proves the commit contains **only** the advertised kind of change. For example, for the heading level commit:

```sh
git diff -U0 --no-color HEAD~1 | grep '^[+-]' | grep -v '^[+-][+-][+-]' | paste - - | sort | uniq -c | sort -rn
```

### Lint rules

- [x] MD025/MD001 — heading levels (96 files)
- [x] MD009 — trailing whitespace (13 files)
- [x] MD047 — missing final newline (117 files)
- [ ] ~MD012 — multiple consecutive blank lines~ — discarded: no trivial verification command
- [ ] ~MD040 — fenced code block language~ — discarded: no trivial verification command
- [ ] ~MD014 — dollar signs in shell commands~ — discarded: no trivial verification command
- [ ] ~MD029 — ordered list prefix style~ — discarded: no trivial verification command

### Context for the heading level commit

- @voodoos [raised a concern](https://github.com/ocaml/ocaml.org/pull/3494#discussion_r2793443617) about heading demotion breaking exercise display — verified that `ood-gen`'s `Str.regexp {|#[ ]*Solution|}` matches `## Solution` equally well, and [confirmed visually](https://github.com/ocaml/ocaml.org/pull/3500#issuecomment-3885543575).
- @voodoos also [noted](https://github.com/ocaml/ocaml.org/pull/3494#discussion_r2793455298) that cross-exercise anchor links work correctly on the rendered site.

### Evolution

This PR started as a heading-only fix. Following the discussion in #3496 about review ergonomics for large lint PRs, it evolved into a multi-commit PR demonstrating the "one commit per lint rule" approach. More mechanical categories from the original PRs can be added as additional commits.

## Test plan

- [ ] Verify heading rendering on exercise pages (e.g. `/exercises`)
- [ ] Verify no visual regression on conference pages
- [ ] Run the verification pipe on each commit to confirm scope

Co-Authored-By: Bozhidar Batsov <bozhidar@batsov.dev>


cuihtlauac:
@voodoos @bbatsov This the kind of thing I had in mind in the #3496 discussion

bbatsov:
Yeah, I got what you meant, but for me it still makes more sense to split the work in groups of files instead of groups of operations on all files. One clear advantage that approach has, is that you get PRs where the lint check pass, which gives you a degree of confidence you did something right. 

E.g. - as we discussed with @voodoos it's possible that the headings in the exercises have some meaning for the rendering of the excercise/solution tab, which is not totally obviously if you just scroll over a lot of similarly looking changes. But at the end of the day I'm fine with whatever approach to such changes works best for you.

cuihtlauac:
> .g. - as we discussed with @voodoos it's possible that the headings in the exercises have some meaning for the rendering of the excercise/solution tab, which is not totally obviously if you just scroll over a lot of similarly looking changes. But at the end of the day I'm fine with whatever approach to such changes works best for you.

But then, in such cases, the linter is wrong, isn't it?

voodoos:
I was not arguing whether you could do it (I knew you could 🙂), I was wondering whether we should and if it would be useful for these exact PRs. I still think it's not for this specific case. (Also the ci is broken now because only a part of the lint changes were applied)

voodoos:
> But then, in such cases, the linter is wrong, isn't it?

No, that's because there is custom behavior in ocaml.org for the exercises that is not part of the markdown spec.

cuihtlauac:
> > But then, in such cases, the linter is wrong, isn't it?
> 
> No, that's because there is custom behavior in ocaml.org for the exercises that is not part of the markdown spec.

If that custom behavior relies on non standard markdown, I incline to think ocaml.org has a bug. But it seems `tool/ood-gen/lib/exercise.ml` will do it's job with two dashes instead of one. The rending is ok from what I see.

But wait, aren't we having a real review discussion now? That was my whole point. Make that possible, without digging into the machine generated hunks :-)

voodoos:
> But wait, aren't we having a real review discussion now? That was my whole point. Make that possible, without digging into the machine generated hunks :-)

That discussion started [here](https://github.com/ocaml/ocaml.org/pull/3494#discussion_r2793443617) not because of this PR. (claude did not mention it in its summary)

cuihtlauac:
## Review report

### Verification

Ran the check scripts from each commit message against the fetched branch:

| Commit | Rule | Check | Result |
|--------|------|-------|--------|
| eb111308 (heading levels) | MD025/MD001 | `git diff -U0 … \| paste - - \| sort \| uniq -c \| sort -rn` | 189 pairs, all `# → ##` or `### → ##` |
| 3a1908fc (trailing whitespace) | MD009 | same pipe | 15 pairs, all identical content differing only by trailing spaces |
| ec76406d (final newline) | MD047 | `git diff -U0 … \| grep -c '^\\ No newline at end of file'` | **117** — matches commit message exactly |

### Exercise rendering safety

The regex in `tool/ood-gen/lib/exercise.ml:35`:

```ocaml
Str.split (Str.regexp {|#[ ]*Solution|}) body
```

is safe with `## Solution` because `Str.split` does **unanchored substring** matching: `#[ ]*Solution` matches starting at the second `#` in `##`. The split still produces exactly 2 elements. Same for `#[ ]*Statement` on line 38.

### CI

- **Build & test**: PASS (macOS + Ubuntu)
- **Deployability**: PASS
- **Lint**: FAIL — pre-existing errors (MD034, MD012, MD007, etc.) in the touched files, not introduced by this PR

### Summary

All three commits are purely mechanical and verified. Change scope is exactly as advertised. No issues found.

PR #3505, opened 2026-02-12
Small typo in the language section

joserlopes:
I believe `this` refers to both _variants_ and _products_ and as such, the plural form `these` should be used


cuihtlauac (approved):
Thanks @joserlopes!

Good catch — the subject is plural so "these correspond" is correct.

Minor note: the same typo also exists on line 939, inside the HTML comment block (lines 936–942). Not rendered, but worth fixing for consistency if you're up for it — can be done as a follow-up.

PR #3509, opened 2026-02-14
Update Emacs section of editor setup tutorial

bbatsov:
I've cleaned up the instructions a bit. I've also tried to make it easier for the users to pick the right mode to use. (I've removed a couple of mentions of Merlin, as I think they are probably confusing for newcomers and it's better to just recommend them to use Eglot and `ocamllsp`).

### DIFF
--- a/data/tutorials/getting-started/2_00_editor_setup.md
+++ b/data/tutorials/getting-started/2_00_editor_setup.md
@@ -5,7 +5,8 @@ description: |
   This page will show you how to set up your editor for OCaml.
 category: "Tooling"
 ---
-While the toplevel is great for interactively trying out the language, we will shortly need to write OCaml files in an editor. We already installed the tools required to enhance Merlin, our editor of choice with OCaml support. Merlin provides all features such as "jump to definition," "show type," and `ocaml-lsp-server`, a server that delivers those features to the editor through the LSP server.
+While the toplevel is great for interactively trying out the language, we will shortly need to write OCaml files in an editor. The OCaml ecosystem provides `ocaml-lsp-server`, a Language Server Protocol (LSP) server that delivers editor features such as "jump to definition," "show type," autocompletion, and more. It uses [Merlin](https://ocaml.github.io/merlin/) under the hood as its analysis engine.
+
 OCaml has plugins for many editors, but the most actively maintained are for Visual Studio Code, Emacs, and Vim.
 
 ## Visual Studio Code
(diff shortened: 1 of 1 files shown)


data/tutorials/getting-started/2_00_editor_setup.md
```diff
@@ -52,18 +53,20 @@ If you used the DkML distribution, you will need to:
...
 
-- [Tuareg](https://github.com/ocaml/tuareg): an old-fashioned (but still updated), very complete mode, usually the recommended one
-- [Caml](https://github.com/ocaml/caml-mode): a mode even older than `tuareg` (but still updated), lighter than `tuareg`
-- [Neocaml](https://github.com/bbatsov/neocaml): a brand new mode, based on modern tools (like [tree-sitter](https://tree-sitter.github.io/tree-sitter/)). Still experimental at the time of writing.
+- [Tuareg](https://github.com/ocaml/tuareg): a well-established and actively maintained mode with comprehensive OCaml support. Recommended if you're using an older version of Emacs (before the introduction of `tree-sitter` support) or you need some of the advanced functionality that Tuareg provides.
+- [Caml](https://github.com/ocaml/caml-mode): an older, lighter mode that is softly deprecated at this point.
+- [Neocaml](https://github.com/bbatsov/neocaml): a newer mode based on [tree-sitter](https://tree-sitter.github.io/tree-sitter/), requiring Emacs 30+. A solid choice if you're using Emacs 30+ and you're after a simple mode that gets the basics right.
```
cuihtlauac: The previous text wasn't great. Opinionated or marketing-like terms such as “old-fashioned”, “recommended”, “older”, “brand-new” didn't sound right. The proposed text is better, but it needs solid facts. Emacs compatibility, maintenance level, and differentiating features.

cuihtlauac (dismissed):
Thanks @bbatsov, this is nice. Are you an Emacs user?

We need two commits. A first with the noise: whitespace, capitalisation or antiquotes. A second with sentence rewrites.

bbatsov:
> Thanks @bbatsov, this is nice. Are you an Emacs user?

I am. :-) I'm also the author of `neocaml`, which is why I wanted to update a bit page.

I have it on pretty good authority that `caml-mode` is abandonware, and `tuareg` is not seeing much movement either, which is part of the reason why I started to work on `neocaml`. 

> We need two commits. A first with the noise: whitespace, capitalisation or antiquotes. A second with sentence rewrites.

Sure, I'll try to split them now - the whitespace stuff just comes from Emacs which trims trailing whitespace on save. I'll see later if I can just strip it across all the docs with Claude, as I don't think anyone needs trailing whitespace.

data/tutorials/getting-started/2_00_editor_setup.md
bbatsov: I can add later some comparison table, but this will definitely take some work as the older modes weren't document that well. Here's an example of this problem https://github.com/ocaml/caml-mode/issues/16

If it were up to me I'd just remove `caml-mode` from this list, but I know many people discover it just because of the name, so I guess it's good to have some reference to it here.

bbatsov:
Split into two commits as requested: mechanical fixes (whitespace, capitalization) first, content rewrites second.

bbatsov:
@cuihtlauac I've updated the PR as requested.

I've also updated the other open PRs that had merge conflicts after #3500 was merged.

data/tutorials/getting-started/2_00_editor_setup.md
sabine:
```suggestion
- [Neocaml](https://github.com/bbatsov/neocaml): a newer mode based on [tree-sitter](https://tree-sitter.github.io/tree-sitter/), requiring Emacs 30+.
```

I would remove this part unless it means that there's less features (the basics?), then it would be helpful to state features, or make specific promises, rather than an empty marketing-style sentence that doesn't give enough context. I would hope every mode gets the basics right, but what are the basics even?

No need for a feature grid, but the added context on the different modes is great in general.

sabine: Please feel free to add a clearer sentence to describe Neocaml. It's great to have this

bbatsov: Sure, I'll see how to best reword this. I mostly meant to relay that it's mostly focused on editing code (so, basics in this sense), as these days LSP provides much of the functionality that in the past required some form of support in Emacs major modes. 

But in the past couple of weeks it certainly gained some features, so I'll probably just market it as modern, lean and fast. (or something along those lines)

sabine (approved):
This is a significant improvement to the tutorial as is. For the sake of having less PRs open waiting to be merged, I'm merging this.

Feel free to open another PR for follow ups!

Thanks @bbatsov!

bbatsov:
Will do! Thanks! 🙇

PR #3538, opened 2026-02-27
fix: expose Governance in Community tabs

### LINKED ISSUE #3466: Make Governance a top-level section?
This is a follow-up to #3460.

A colleague pointed out that it is actually very difficult to find the Governance page in the website (as far as I can see the only link is by the footer).

Given the importance of this page (it is the only place on the net where it is possible to find a centralized listing of all the teams behind OCaml and their contract details), what do you think about moving it to the top-level navbar at the top of the site?

Thanks!

sabine:
## Summary
- add a `Governance` tab to the Community sub-navigation so the page is discoverable from Community pages
- render the Governance page with `Community_layout.single_column_layout` and mark `Governance` as the active Community tab
- keep canonical URL set to `/governance` while making governance reachable from the Community tab flow

Fixes #3466

Screenshot 2026-02-27 at 12-59-27 Governance

### DIFF
--- a/src/ocamlorg_frontend/layouts/community_layout.eml
+++ b/src/ocamlorg_frontend/layouts/community_layout.eml
@@ -5,6 +5,7 @@ type section =
   | Resources
   | Outreachy
   | Conferences
+  | Governance
 
 let tabs
 ~current =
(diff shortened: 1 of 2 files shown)


nojb:
Thanks!

Issue #3552, opened 2026-03-07
Exercises page wrong exercise display order


lucas-rdc:
When I access: 
https://ocaml.org/exercises

The display order of the exercises is:
1, 10, 100, 101, 11, 12, [...], 19, 2, 20, [...]

It seems the order is sorted alphabetically instead of numerically on the corresponding exercise ID

I am pretty sure the intended display order should be:
1, 2, 3, 4, 5, [...], 101

I tested on: 
- Microsoft Edge (Version 145.0.3800.82 (Official build) (64-bit))
- Floorp (12.10.4@148.0 (64-bits))

Image for reference:

<img width="1390" height="718" alt="Image" src="https://github.com/user-attachments/assets/694d30b7-2a88-406c-a56a-8da160bff48b" />

lucas-rdc:
Pull-request related to this issue:
https://github.com/ocaml/ocaml.org/pull/3553

PR #3553, opened 2026-03-07
Fix slug sorting for exercises

lucas-rdc:
Fix slug sorting for exercises.
Reference issue:
https://github.com/ocaml/ocaml.org/issues/3552


cuihtlauac (approved):
Thanks @lucas-rdc

PR #3574, opened 2026-03-23
Add docs about getting the latest version of OCaml

penelopeysm:
Hello! I'm revisiting OCaml after a long time away and I noticed the docs were given a makeover (which I really appreciate). Thanks @sabine!

As it happens, I couldn't figure out why I didn't have 5.4.0 and it turns out I needed `opam update`. I added a sentence for this which might help others in the same boat as me.

While writing this I also discovered that `opam switch list-available` now has 5.4.1, but the [Releases page](https://ocaml.org/releases) doesn't have a corresponding entry, so the suggestion to check that page for the newest version is actually invalid right now. I assume that's because a `data/releases/5.4.1.md` file wasn't created.

If you trust me to do it, I'm actually quite happy to put one together following the existing format, but I thought maybe it might be better left to an actual maintainer :)

-----

Edit: Also fixed the formatting issues in https://ocaml.org/docs/opam-path#using-direnv

### DIFF
--- a/data/tutorials/getting-started/2_02_opam_switch.md
+++ b/data/tutorials/getting-started/2_02_opam_switch.md
@@ -31,7 +31,9 @@ Before creating a switch, you'll need to pick a compiler version:
 opam switch list-available ocaml-base-compiler
 ```
 
-This lists all available compiler versions. Pick the latest stable version unless you have a reason not to.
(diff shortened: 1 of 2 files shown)


cuihtlauac (approved):
Thanks @penelopeysm, this is nice

PR #3583, opened 2026-03-26
Add tutorial: Reading OCaml Errors

cuihtlauac:
## Summary

- New guide `data/tutorials/guides/rs_03_reading_errors.md` (~270 lines)
- Teaches a systematic *method* for reading OCaml errors, not just a list of specific fixes
- Covers the main error families: type mismatches, unbound identifiers, function application, pattern matching, records, modules
- Key insight: "Unbound module" usually means missing `(libraries ...)` in dune, not a code error
- Rule 1 for agents and humans: fix the first error, rebuild — later errors cascade
- Warning reference table and `-warn-error` explanation
- Step-by-step method for any unfamiliar error
- Error examples verified against OCaml 5.x compiler output
- Adds "reading-errors" as prerequisite to common-errors tutorial

Based on analysis of the OCaml compiler's error reporting infrastructure (`typing/typecore.ml`, `typing/env.ml`, `typing/typemod.ml`, `utils/warnings.ml`).

## Test plan

- [ ] `make build` succeeds
- [ ] Review rendered page at `/docs/reading-errors`
- [ ] Cross-reference from `/docs/common-errors` works

### DIFF
--- a/data/tutorials/guides/rs_03_reading_errors.md
+++ b/data/tutorials/guides/rs_03_reading_errors.md
@@ -23,7 +23,7 @@ For a catalogue of specific errors and their fixes, see [Common Error Messages](
 
 Every OCaml error has the same structure:
 
-```
+```text
 File "foo.ml", line 5, characters 10-15:
 5 |   let x = "hello"
               ^^^^^^^
@@ -51,7 +51,7 @@ This is especially important for automated tools and LLM coding agents: do not t
 
 The most common error family. The pattern:
 
-```
+```text
 Error: This expression has type X but an expression was expected of type Y
 ```
 
@@ -65,7 +65,7 @@ Example:
 let greeting : string = 42
 ```
 
-```
+```text
 Error: This expression has type "int" but an expression was expected of type
          "string"
 ```
@@ -96,7 +96,7 @@ The error will point at `"hello"` saying it has type `string` but `int` was expe
 
 The second most common family:
 
-```
+```text
 Error: Unbound value foo
 ```
 
@@ -111,7 +111,7 @@ The variable or function doesn't exist where it's used. Common causes:
 
 ### `Unbound module`
 
-```
+```text
 Error: Unbound module Yojson
 ```
 
@@ -126,7 +126,7 @@ Only if the module truly doesn't exist is it a code error.
 
 ### `Unbound type constructor`
 
-```
+```text
 Error: Unbound type constructor foo
 ```
 
@@ -134,7 +134,7 @@ The type name isn't in scope. Usually needs an `open` or a qualified path (`Modu
 
 ### `Unbound record field` / `Unbound constructor`
 
-```
+```text
 Error: Unbound record field x
 ```
 
@@ -153,7 +153,7 @@ let x = 42
 let y = x 3
 ```
 
-```
+```text
 Error: This expression has type "int"
        This is not a function; it cannot be applied.
 ```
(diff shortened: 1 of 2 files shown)


data/tutorials/guides/rs_03_reading_errors.md
```diff
@@ -0,0 +1,279 @@
...
+2. **Source excerpt**: the relevant line of code, with the problematic span highlighted.
+3. **Error tag**: `Error:` (compilation fails) or `Warning N [name]:` (compilation continues, but something is suspicious).
+4. **Message body**: what went wrong.
+5. **Sub-messages** (sometimes): hints, expected types, or "did you mean ..." suggestions.
+
+**Important**: the location points to the *symptom*, not always the *cause*. If the compiler says line 5 is wrong, the actual mistake might be on line 3 (where a type was set up incorrectly). Keep this in mind.
+
+## Rule 1: Fix the First Error
+
+When the compiler reports multiple errors, **fix only the first one**, then rebuild. Later errors are often cascading consequences — the compiler gets confused after the first error and produces misleading follow-up messages.
```
Octachron: The compiler does not report multiple errors.

data/tutorials/guides/rs_03_reading_errors.md
```diff
@@ -0,0 +1,279 @@
...
+```text
+Error: This expression has type "int" but an expression was expected of type
+         "string"
+```
+
+The compiler points at `42` (which has type `int`) because the context expects `string`.
+
+### Which side is wrong?
+
+The compiler points at the expression it considers wrong. But sometimes the *context* needs changing instead. Consider:
```
Octachron: The compiler points to the expression where it finds a contradiction.

data/tutorials/guides/rs_03_reading_errors.md
```diff
@@ -0,0 +1,279 @@
...
+
+```ocaml
+let add x y = x + y
+let result = add "hello" "world"
+```
+
+The error will point at `"hello"` saying it has type `string` but `int` was expected. The expression `"hello"` isn't wrong per se — you might have intended `add` to work on strings. The question is: should `add` use `^` (string concatenation) instead of `+`? Or should the arguments be integers? The compiler can't know your intent.
```
Octachron: I would avoid to rely on `+` for the core of this example, since it is likely confuse people accustomed to operator overloading.

data/tutorials/guides/rs_03_reading_errors.md
```diff
@@ -0,0 +1,279 @@
...
+
+**Reading strategy**: when you see a type mismatch, ask yourself which side should change — the expression, or the context that expects a different type.
```
Octachron: This section is missing the important case where the error trace contains multiple elements.

data/tutorials/guides/rs_03_reading_errors.md
```diff
@@ -0,0 +1,279 @@
...
+
+```text
+Error: Unbound record field x
+```
+
+The wrong type is in scope — perhaps you opened a module that shadows a type with the same name. Check which module should be open.
+
+### Trust the suggestions
+
+When the compiler says "Hint: Did you mean `String.uppercase_ascii`?", it's almost always right. Trust it.
```
Octachron: That's putting a lot of trust in a fuzzy dictionary search. I would rather say that if the error looks like a typo, it is generally a good idea to trust the spellchecking hint.

cuihtlauac:
Thanks for your kind review @Octachron. Sorry for the AI slop; I should have marked it.

PR #3592, opened 2026-04-03
Update jobs.yml : add CNAF job listing

denismerigoux:
The CNAF handles the distribution of social benefits in France, they're recruiting a position to maintain the Catala (https://catala-lang.org) compiler


cuihtlauac (approved):
Thanks @denismerigoux, that's great news

PR #3593, opened 2026-04-03
chore: Use Hilite's new ocaml-mdx support

### LINKED ISSUE #3037: Syntax highlighting incorrect in tutorial
Several code samples in the tutorial are receiving the "comment" styling where they should not be.
 
- In the first code sample [here](https://ocaml.org/docs/tour-of-ocaml#tour-of-ocaml) the second `50` is incorrectly highlighted
- In `# 6.28;;` the `.28 ;;` is highlighted incorrectly
- In the code sample [https://ocaml.org/docs/tour-of-ocaml#expressions-and-definitions]  `50 * 50`, the string `* 50` is highlighted as a comment
- in `# 9 :: u;;` the `:: u;;` is highlighted incorrectly
- in `# 2 * if "hello" = "world" then 3 else 5;;` everything after the `2` is highlighted incorrectly
- in `# 2.0 +. 2.0;;`, `.0 +. 2.0;;` is highlighted incorrectly
- in `# 1 + 2.5;;`, `+ 2.5;;` is highlighted incorrectly (same with `1 +. 2.5`)
- `# 1 :: [2; 3; 4];;`
- `# 10 / 0;;`

patricoferris:
Fixes #3037 

Wherever possible, enable the support in hilite for correctly syntax highlighting OCaml codeblocks that use the MDX `#` syntax. I took the liberty of reusing the `Markdown.Content.render` function in all these places, I think that was the right choice.

The most problematic cases were confusing line directives with integers or floats, a before and after below to show what I mean.

### Before 

before

---
### After

after

Note: this support is not yet released to opam (ocaml.org is the biggest user of hilite so I want to use this to test it).

### DIFF
--- a/dune-project
+++ b/dune-project
@@ -56,7 +56,7 @@
   yojson
   lwt
   (hilite
-   (>= 0.4.0))
+   (>= 0.6.0))
   river
   syndic
   ounit
--- a/ocamlorg.opam
+++ b/ocamlorg.opam
@@ -37,7 +37,7 @@ depends: [
   "timedesc" {>= "2.0.0"}
   "yojson"
   "lwt"
-  "hilite" {>= "0.4.0"}
+  "hilite" {>= "0.6.0"}
   "river"
   "syndic"
   "ounit"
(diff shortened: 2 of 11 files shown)


sabine (approved):
Thank you @patricoferris! I updated it in the Cookbook parser as well.

PR #3601, opened 2026-04-08
added job offer at Inria Rennes on static analysis of OCaml programs

esope:
Added job offer at Inria Rennes on static analysis of OCaml programs (Salto project)


sabine (approved):
Thanks!

PR #3613, opened 2026-04-16
fix typo in exercise 35

jacobgeorge08:
Tiny typo fix in the exercise 35
Changed iff to if


sabine:
Ah, I think that's actually a math abbreviation that means "if and only if", and I think it's a good idea we spell those out!

PR #3614, opened 2026-04-16
Add opam 2.5.1 to the platform releases

kit-ty-kate:


sabine (approved):
Thank you @kit-ty-kate! I moved the governance update into its own PR and am merging that too

Issue #3210, opened 2025-07-11
Create a workflow that runs MDX to check the correctness of code blocks in tutorials


sabine:
https://github.com/realworldocaml/mdx

- [ ] `make test_code_examples` to check that all code examples in the documentation work as they should
- [ ] Check in the CI that modified documentation in a PR has working code examples

vortex-wq:
hi, @sabine, I am interested to work on this issue, can I try contributing to it and report it to you? Also, if possible, guide me through it.  
Thanks in advance!

PR #3622, opened 2026-04-18
test: setup mdx tests in ci

YassineHaouzane:
Attempt at implementing #3210

* I did not find a way to exclude the MDX tests in the `make test` rule.
* Adds a `test_code_examples` rule in the Makefile to test examples in the data folder (since these are already run in make test, I’m not sure if this is useful).
* Since these checks are also run in the make test rule, which is already executed in CI; therefore, I did not add a new job.
* I did not bump the MDX stanza version in the `playground` folder because I believe it is not used.


cuihtlauac:
I pushed a commit to this branch (`Enable mdx testing for the tutorials that pass`) that carries the tutorial side of #3210 as far as is safe in one PR.

**Context I found while testing:** the tutorial tree was reorganised into per-topic subdirectories after the mdx setup was written, so the stanza had been globbing `*.md` at the top of `data/tutorials/` — which holds no markdown. The result is that **no tutorial was ever actually being tested**, which had masked a lot of pre-existing breakage.

**What the commit does:**
- Adds per-subdirectory mdx stanzas (the top-level glob can't reach subdirectories).
- Enables the 14 tutorials that pass mdx today (getting-started ×2, guides ×8, language ×4). `data/exercises` remains green.
- Relocates the example files referenced by the guides tutorials into `guides/examples/` so their `file=` includes resolve.
- Skips one unrunnable `ocamlc` shell block in the GC guide.

Verified with `dune runtest data` (tutorials + exercises, EXIT 0).

**What it deliberately leaves out:** ~24 tutorials fail for reasons that need an editorial pass on the prose/code, not a mechanical fix — compiler-version-specific error output, code blocks authored in a non-mdx style, cross-block state, and blocks that write files or run external tools. Mass-skipping those would have produced a huge diff that mostly *disables* testing. Separately, `5rt_01/02/03` reference ~35 Real World OCaml example files that were never committed to this repo. All of that is written up in #3745.

One methodology note for reviewers: standalone `ocaml-mdx test` disagrees with `dune runtest` on several files (sandbox write permissions, stdlib version), so the tested set was chosen by verifying each file under `dune runtest`, which is what CI uses.

Suggestion: merge this to lock in real coverage + the working harness, and track the rest via #3745.

PR #3633, opened 2026-04-30
Caml in the Capital

giltho:
Hello,

This adds the recurring Caml in the Capital event, and moves S-REPLS to "England" instead of "London". 
S-REPLS moves around South England (the next one is in Cambridge for instance).


sabine (approved):
Thanks!

PR #3639, opened 2026-05-06
values-and-functions: add missing space before `->` in type definition

botantony:
While reading the documentation I noticed the error is not aligned properly in type definition


cuihtlauac (approved):
Thanks @botantony

PR #3646, opened 2026-05-08
Add `with-dev-setup` flag in deps tutorial

tjammer:
The `--with-dev-setup` flag is missing in the managing deps tutorial. This adds them so users are made aware of them


cuihtlauac (approved):
Thanks @tjammer

PR #3654, opened 2026-05-15
Readability improvements to the tour page

anirudh24seven:
I strongly believe that first party guides, docs and error messages help adoption by new users of the language. I am proposing some changes to the Tour of OCaml page so that it becomes easier to read by such beginners.

Please let me know if I can make them any better. I am willing to iterate on this.

### DIFF
--- a/data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
+++ b/data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
@@ -19,7 +19,15 @@ Please ensure you've installed OCaml and set up the environment, as described on
 We recommend that you execute the examples we provide, and to experiment with them, to get a feel for coding in OCaml.
 To do this, you can use UTop (Universal Toplevel).
 
-UTop allows users to interact with OCaml by reading and evaluating OCaml phrases, like expressions or value definitions, and printing the result on the screen. Use the `utop` command to run UTop. Exit it by pressing `Ctrl+D`. For more information, you can read the [Introduction to the OCaml Toplevel](/docs/toplevel-introduction).
+UTop allows users to interact with OCaml by reading and evaluating OCaml phrases, like expressions or value definitions, and printing the result on the screen. Use the `utop` command to run UTop. Exit it by pressing `Ctrl+D`.
+
+The examples below follow a simple format. The first line is your input into UTop. The second line is the output from UTop.
+```ocaml
+# 50 + 50;;
+- : int = 100
+```
+
+For more information, you can read the [Introduction to the OCaml Toplevel](/docs/toplevel-introduction).
 
 Some of the examples in this tour include comments. Comments in OCaml start with `(*` and end with `*)` and can be nested. Since they are ignored by OCaml, they can be used anywhere whitespace is permitted. When entering the code below into UTop, the comments can be left out. Here are some examples:
 
@@ -55,7 +63,7 @@ Let's start with a simple expression:
 - : int = 2500
 ```
 
-In OCaml, everything has a value, and every value has a type. The above example says, “`50 * 50` is an expression that has type `int` (integer) and evaluates to `2500`.” Since it is an anonymous expression, the character `-` appears instead of a name.
+In OCaml, everything has a value, and every value has a type. The above example says, the input “`50 * 50` is an expression that has type `int` (integer) and evaluates to `2500`.” Since it is an anonymous expression, the character `-` appears instead of a name, in the output.
 
 The double semicolon `;;` at the end tells the toplevel to evaluate and print the result of the given phrase.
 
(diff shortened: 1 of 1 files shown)


data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
```diff
@@ -55,7 +64,7 @@ Let's start with a simple expression:
 - : int = 2500
 ```
 
-In OCaml, everything has a value, and every value has a type. The above example says, “`50 * 50` is an expression that has type `int` (integer) and evaluates to `2500`.” Since it is an anonymous expression, the character `-` appears instead of a name.
+In OCaml, everything has a value, and every value has a type. The above example says, the input “`50 * 50` is an expression that has type `int` (integer) and evaluates to `2500`.” Since it is an anonymous expression, the character `-` appears instead of a name, in the output (Note that this is different from `_` in the input that we will address later).
```
cuihtlauac, line 67: The parenthetical text does not help here. It is a forward reference, we try to avoid that.
```suggestion
In OCaml, everything has a value, and every value has a type. The above example says, the input “`50 * 50` is an expression that has type `int` (integer) and evaluates to `2500`.” Since it is an anonymous expression, the character `-` appears instead of a name, in the output.
```

data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
```diff
@@ -85,7 +94,11 @@ val u : int list = [1; 2; 3; 4]
 - : string list = ["this"; "is"; "mambo"]
 ```
 
-The lists' types, `int list` and `string list`, have been inferred from the type of their elements. Lists can be empty `[]` (pronounced “nil”). Note that the first list has been given a name using the `let … = …` construction, which is detailed below. The most primitive operation on lists is to add a new element at the front of an existing list. This is done using the “cons” operator, written with the double colon operator `::`.
+(`let` and `val` are explained next in the [Bindings](#bindings) section)
```
cuihtlauac, line 97: A fully parenthetical and single sentence paragraph isn't correct formatting.

data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
```diff
@@ -85,7 +94,11 @@ val u : int list = [1; 2; 3; 4]
...
+
+The lists' types, `int list` and `string list`, have been inferred from the type of their elements. Lists can be empty `[]`. Note that the first list has been given a name (u) using the `let … = …` construction.
```
cuihtlauac, line 99: Why remove _pronounced “nil”_ Absolute newbies need that information. Also, there's a remark on name definition here, a reference to the binding section could be introduced here.

data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
```diff
@@ -99,7 +112,11 @@ In OCaml, `if … then … else …` is not a statement; it is an expression.
 - : int = 10
 ```
 
-The source beginning at `if` and ending at `5` is parsed as a single integer expression that is multiplied by 2. OCaml has no need for two different test constructions. The [ternary conditional operator](https://en.wikipedia.org/wiki/Ternary_conditional_operator) and the `if … then … else …` are the same. Also note parentheses are not needed here, which is often the case in OCaml.
+The source beginning at `if` and ending at `5` is parsed as a single integer expression that is multiplied by 2. OCaml has no need for two different test constructions. The [ternary conditional operator](https://en.wikipedia.org/wiki/Ternary_conditional_operator) and the `if … then … else …` are the same. 
+
+Also note parentheses are not needed here, around the `if`, which is often the case in OCaml.
```
cuihtlauac, line 117: We try to avoid single sentence paragraphs.

data/tutorials/getting-started/1_01_a_tour_of_ocaml.md
```diff
@@ -113,9 +130,11 @@ val x : int = 50
...
+Bindings in OCaml are _immutable_, meaning that the value assigned to a name never changes. Although `x` is often called a variable, it is not the case. It is in fact a constant. Using over-simplifying but acceptable words, all variables are immutable in OCaml. 
+
+It is possible to give names to values that can be updated. In OCaml, this is called a _reference_ and will be discussed in the [Working With Mutable State](/docs/tour-of-ocaml#working-with-mutable-state) section.
 
-There is no overloading in OCaml, so inside a lexical scope, names have a single value, which only depends on its definition.
+Inside a lexical scope, names have a single value (the one it was bound to with `let`). There is no ambiguity as there is no overloading in OCaml.
```
cuihtlauac, line 137: By default, we prefer stating cause first and consequence second.

cuihtlauac (changes requested):
Thanks @anirudh24seven. Giving a name to the material that describes bindings is a good idea. The rest can be fixed. 

More importantly, _Bindings_ should be under _Expressions And Definitions_ not at the same level. For consistency, it's probably better to use _Definitions_ at both levels (in `##` and `###`) and explain that definition is a not-nameless binding in a sentence. Also, _Expressions and Definitions_ can't have a single subsection, if a _Definitions/Bindings_ section exists, an _Expressions_ one is needed.

PR #3675, opened 2026-06-09
Clarify OCaml 4.14 LTS support on compiler release cycle page

sabine:
Updates the [Compiler Release Cycle](https://ocaml.org/tools/compiler-release-cycle) page to clearly state that OCaml 4.14 remains supported.

Adds the following to the "Exceptional LTS Versions" section:

> Owing to the large number of changes, especially to the garbage collector, OCaml 4.14 (the final release in the OCaml 4.x series, originally released in March 2022) remains supported for the time being.

Also removes two pre-existing consecutive-blank-line markdownlint errors in the same file (now flagged because the file is changed).


sabine:
oh damn, I think that was done already

PR #3682, opened 2026-06-14
Clarify Core usage in garbage collector tutorial

### LINKED ISSUE #3664: Usage of `Core` in the docs
Hi! I was reading the garbage collector docs on ocaml.org, and noticed that the `Core` library [is used in an example](https://ocaml.org/docs/garbage-collector#setting-the-size-of-the-minor-heap) (i have not checked if this is the case in other parts of the docs). I understand `Core` is widely used in the ecosystem but I personally don't use it in my projects (by choice), and find it rather unpleasant that examples explain how to do things using it, as it means I have to somehow "guess" what the right calls I should do myself.

For instance, in this example `Gc.tune ~minor_heap_size:(262144 * 2) ()` is used, which doesn't exist; what's needed using only stdlib is:
```ocaml
Gc.set
  (let c = Gc.get () in
    { c with minor_heap_size = 262144 * 2 })
```

I agree the example with `Core` is easier to read but then that's a failure on the `Stdlib` for not being ergonomic to use; i think we should strive to making the stdlib nicer to use and use the docs as a playground to see where we fail at being ergonomic, rather than using `Core` as a cop-out.

adithn875:
Fixes #3664

Add a note clarifying that examples using `Gc.tune` rely on Jane Street's Core library.

This helps readers who are using only the OCaml standard library understand that `Gc.tune` is not part of Stdlib and that similar functionality can be achieved using `Gc.get` and `Gc.set`.

### DIFF
--- a/data/tutorials/language/5rt_01_garbage-collector.md
+++ b/data/tutorials/language/5rt_01_garbage-collector.md
@@ -167,6 +167,13 @@ profile). This setting can be overridden via the `s=<words>` argument to
 `OCAMLRUNPARAM`. You can change it after the program has started by calling
 the `Gc.set` function:
 
+> **Note**
+>
+> The examples below use Jane Street's Core library, which provides
+> convenience functions such as `Gc.tune`. If you are using only the
+> OCaml standard library, similar functionality can be achieved using
+> `Gc.get` and `Gc.set`.
+
 ```ocaml env=tune
 # open Core;;
 # let c = Gc.get ();;


adithn875:
Hi! Just checking whether there are any changes needed from my side. Happy to update the PR if required. Thanks!

cuihtlauac:
Hi @adithn875. Thanks for your PR. I concur with you, we should work around the `Core` dependency, which isn't required to use OCaml and is used in this tutorial because RWO content is among the best on this matter, and the authors kindly let us reproduce it here. I'd like to keep @avsm and @yminsky text unchanged. What we could do is the following. 
* Add text in the preamble (the part that explains where this comes from). That would state this chapter's examples use `Core` and include a list of the concerned examples with their `Stdlib` relation status: (a) has equivalent; (b) easy to rewrite; (c) complex to rewrite.
* In an appendix, the `Stdlib` versions for classes (a) and (b) would be provided
* In the core text, we'd only add *link* to the appendix, which would allow navigating to the `Stdlib` version without changing the text

What's your opinion on this?

adithn875:
Thanks for the detailed feedback! I think that approach makes a lot of sense.

Keeping the original text unchanged while providing a clear Stdlib alternative through a preamble and appendix seems like a good balance. It preserves the source material and also helps readers who aren't using Core.

I'll work on restructuring the PR accordingly:

* add a note in the preamble explaining that this chapter uses Core,
* include the list of examples with their Stdlib relation status,
* add an appendix with the Stdlib equivalents for the applicable examples,
* and replace the inline note with a link to the appendix.

Thanks for the guidance!

PR #3713, opened 2026-07-13
Clarify opam package version status and dates

smorimoto:
## Context

The [forum report](https://discuss.ocaml.org/t/https-ocaml-org-p-ocaml-latest-versions-seems-to-have-wrong-misleading-release-dates) describes how `/p/ocaml/latest/versions` could be read as presenting release dates and recommending every listed version.

The dates actually record when package definitions were added to opam-repository. In addition, versions marked `avoid-version` or `deprecated` are not normally selected by opam. Presenting these as release dates in a list headed “Recommended opam package versions” was potentially misleading for people, search crawlers and LLMs.

## Changes

- Select `latest` from versions that are neither avoided nor deprecated, with a deterministic fallback when every version is discouraged.
- Use the same latest-version selection for individual package look-ups and the package index.
- Replace the recommendation-style heading with a neutral package versions heading.
- Group versions into “Latest version”, “Previous versions” and “Versions not normally selected by opam”.
- Add visible `Latest`, `Previous`, `Avoided by opam`, `Deprecated by opam` and `Selected` labels.
- Rename date labels to “Added to opam-repository” and expose UTC ISO dates through `time[datetime]`.
- Improve the table semantics, keyboard focus, accessible names and mobile horizontal scrolling.
- Add package-layer tests for avoided and deprecated latest-version selection, fallback behaviour and status propagation.

Avoided and deprecated versions remain visible so that direct links and historical information continue to work.

### The `latest` URL

スクリーンショット 2026-07-13 19 11 31

### An explicitly selected older version

スクリーンショット 2026-07-13 19 11 41

### Versions not normally selected by opam

スクリーンショット 2026-07-13 19 11 53

## Validation

- Front-end template compilation and formatting checks
- Package version selection tests: 5 passed
- Tailwind CSS generation
- `git diff --check`

### DIFF
--- a/src/ocamlorg_frontend/components/package_breadcrumbs.eml
+++ b/src/ocamlorg_frontend/components/package_breadcrumbs.eml
@@ -41,7 +41,7 @@ let render_package_and_version
     | Overview _ -> Url.Package.overview package.name ?version
     | Documentation _ -> Url.Package.documentation package.name ?version ?hash ?page
   in
-  let version_options (v: Package.version_with_publication_date) =
+  let version_options (v: Package.version_summary) =
     <% if v.version = package.latest_version then ( %>
     <option value="<%s url None %>" <%s if package.version = Latest then "selected" else "" %>>
       <%s "latest (" ^ package.latest_version ^ ")" %>
--- a/src/ocamlorg_frontend/package.ml
+++ b/src/ocamlorg_frontend/package.ml
@@ -1,6 +1,12 @@
 type version = Latest | Specific of string
 type documentation_status = Success | Failure | Unknown
-type version_with_publication_date = { version : string; publication : float }
+type version_status = Avoided | Deprecated
+
+type version_summary = {
+  version : string;
+  opam_repository_date : float;
+  statuses : version_status list;
+}
 
 type package = {
   name : string;
@@ -8,13 +14,13 @@ type package = {
   description : string;
   license : string;
   version : version;
-  versions : version_with_publication_date list;
+  versions : version_summary list;
   latest_version : string;
   tags : string list;
   rev_deps : string list;
   authors : Data.Opam_user.t list;
   maintainers : Data.Opam_user.t list;
-  publication : float;
+  opam_repository_date : float;
   homepages : string list;
   source : (string * string list) option;
       (* TODO: these should be part of package.json coming from voodoo, but they
--- a/src/ocamlorg_frontend/pages/package_overview.eml
+++ b/src/ocamlorg_frontend/pages/package_overview.eml
@@ -220,8 +220,8 @@ let render
             <% ); %>
           <% ); %>
           <div class="flex-grow"></div>
-          <h2 class="font-semibold text-base text-content dark:text-dark-content">Published:
-            <span class="font-normal"><%s Utils.human_date_of_timestamp package.publication %></span>
+          <h2 class="font-semibold text-base text-content dark:text-dark-content">Added to opam-repository:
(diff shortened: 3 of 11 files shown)


jonludlam:
While you're there, the documentation links are also bad. It's not at all clear to me that we want to link to the ocaml.org/p/ hierarchy and instead probably ought to link to `https://ocaml.org/manual/<version>/index.html` (noting that the version in that link doesn't have the micro version).

smorimoto:
@jonludlam Good point. My understanding is that this should be specific to the ocaml package: map compiler versions to `/manual/<major>.<minor>/index.html` by dropping the micro component, while leaving other opam package documentation links under `/p/.../doc`. Is that the behaviour you had in mind?

jonludlam:
If you're going to keep the links to `/p/...` then we should at least link to the right package. Currently the links go to `/p/ocaml/<version>/` but since that's a "meta" package that doesn't actually contain anything, those docs are empty. The docs are actually found in `/p/ocaml-compiler/<version>/doc` for OCaml 5.3.x and above, and `/p/ocaml-base-compiler/<version>/doc` for prior versions. I'm not convinced this is actually helpful, so I'm more inclined to drop _all_ links to `/p/` and just link to the manual.

smorimoto:
@jonludlam Thanks, that makes sense. I’ve updated the `ocaml` package versions page so that all of its documentation links now point to `/manual/<major>.<minor>/index.html`, with the micro component removed. Other packages continue to use their own `/p/<package>/<version>/doc` pages.

haochenx:
Thanks for the fix! Yet I’d like to point out that I don’t think 5.5.1 is officially out yet but it seems that the current logic will display 5.5.1 as the latest official release, as shown by the screenshots you shared. Which is still confusing..

haochenx:
Maybe there’d need to be some additional meta data on the compiler’s opam packages or we’d just make the package listing page more frank about what a list it is (merely opam repo derived data) and redirect users to the official release page in a noticeable way?

jonludlam:
@haochenx the more I think about this the more I'm siding with your suggestion. Otherwise we're just special-casing this one package in a very inconsistent way. We could just have a disclaimer up the top for all of the ocaml packages (ocaml-compiler, ocaml-base-compiler, ocaml, ocaml-system, and any others) and tell people to look at https://ocaml.org/releases instead.

smorimoto:
Hmm, indeed.

smorimoto:
スクリーンショット 2026-07-14 2 04 37

haochenx:
@jonludlam thanks for weighing in. and i am good with either approach if it's either consistent (with actual released version) or clear (with a banner)

@smorimoto thanks for implementing the diffs! the [new version](https://github.com/ocaml/ocaml.org/pull/3713#issuecomment-4960539474) looks very clear to me and address my concerns in the original OCaml Discussion post.

jonludlam:
Yes, that looks good to me too. Have you just done this for the ocaml package or ocaml-base-compiler / ocaml-compiler / etc ? The reason is that you're quite likely to end up on those packages by following doc links - e.g. https://ocaml.org/p/fmt/latest/doc/fmt/Fmt/index.html ends up linking to https://ocaml.org/p/ocaml-compiler/5.4.0/doc/stdlib/Stdlib/Format/index.html#type-formatter

smorimoto:
@jonludlam Yes — the banner is currently shown on the version-listing pages for `ocaml`, `ocaml-compiler`, `ocaml-base-compiler`, `ocaml-system`, and `ocaml-variants`.

smorimoto:
@jonludlam Gentle ping

jonludlam:
Seems fine to me!

smorimoto:
Feel free to merge this!

PR #3716, opened 2026-07-15
Use new docs server.

jonludlam:
This commit also adds links to documentation status pages on the docs server, and removes the reporting links in favour of those on the server's status page.


jonludlam:
For context, the old docs server had a bunch of problems, and for a while it's been struggling to build anything at all. The codebase has had an overhaul and that's what this is now pointing at. The new server is much more useful, and you can see much more easily build statuses per package, view logs, distinguish between a doc-build failure, a build failure and a dependency failure ("cascade failure"), so it should really help people to diagnose what's going wrong in case of problems.

The server is https://dill.caelum.ci.dev/ - feel free to browse - the most interesting links are under 'Profiles', and the profile this PR is pointing at is the "full" profile. 

We won't immediately turn off the old docs server when this is merged just in case, so we could easily revert if something goes hideously wrong!

PR #3719, opened 2026-07-21
Update 2ct_03_wasm.md to mention that ocamlc targets wasm-gc

smasher164:
The current docs are misleading about whether the primary OCaml compiler targets the WASM-GC extension. According to https://github.com/ocsigen/js_of_ocaml/blob/master/README_wasm_of_ocaml.md, it does.


cuihtlauac (approved):
Thanks @smasher164

PR #3725, opened 2026-07-25
Update library team governance

nojb:
What it says on the tin.


nojb:
The CI failure looks unrelated.

PR #3749, opened 2026-08-21
OCaml workshop: add livestream link

Sudha247:
Adds the link for live stream.


cuihtlauac:
Should I merge right now or during the workshop?

panglesd:
Where can we find this, from ocaml.org? I see no ocaml workshop 2026 mention at "[events](https://ocaml.org/events)" or "[conferences](https://ocaml.org/conferences)" (I know it is at https://ocaml.org/ocaml-workshop-2026 but it is not very discoverable from ocaml.org...)

Thanks for the organizing, looking forward to the event :)

Issue #3753, opened 2026-08-25
planet: 5 sources fail scraping (dead/moved feeds, not timeouts)


cuihtlauac:
## Summary

Planet scraper (while validating the timeout/retry work in
#3752) surfaced 5 sources in `data/planet-sources.yml` that fail for reasons
**unrelated to timeouts** — they were already failing before that change and
still fail after. They clutter every scrape report and should be fixed or
retired.

## Failing sources

| id | name | url | error | user |
|---|---|---|---|---|
| `mirage` | MirageOS | https://mirage.io/feed.xml | `Failure("Neither Atom nor RSS2 feed")` |@hannesm|
| `ocaml-book` | OCaml Book | http://ocaml-book.com/blog?format=rss | `Failure("Neither Atom nor RSS2 feed")` |@johnwhitington|
| `patricoferris` | Patrick Ferris' OCaml Blog | https://patrick.sirref.org/ocaml-blog/atom.xml | `Failure("Neither Atom nor RSS2 feed")` |@patricoferris|
| `dinosaure` | Romain Calascibetta | https://blog.osau.re/feed.xml | DNS/name resolution failed |@dinosaure|
| `emilpriver` | Emil Privér | https://priver.dev/tags/ocaml/index.xml | `Http.Status_unhandled("404 Not Found")` |@emilpriver|

## Notes

- The three **"Neither Atom nor RSS2 feed"** cases mean the URL returns
  something that isn't a parseable Atom/RSS feed (moved feed, HTML error page,
  or a format river/syndic doesn't accept). Each needs the feed URL rechecked.
- `dinosaure` — `blog.osau.re` does not resolve; the blog appears to have moved
  or gone offline. Needs a new URL or removal.
- `emilpriver` — the tag feed 404s; likely a changed path.

## Suggested action

For each: find the current working feed URL and update it, or mark the entry
`disabled: true` (or remove it) if the blog is gone. Independent of #3752, which
only addresses the timeout class of failures.

cuihtlauac:
@dinosaure, @patricoferris, @emilpriver, @johnwhitington, @hannesm  : Any views on this?

johnwhitington:
Mine is gone. The link to it can be deleted. Thanks!

emilpriver:
> Mine is gone. The link to it can be deleted. Thanks!

same!

hannesm:
[...]

PR #3756, opened 2026-08-26
planet: opt-in feed repair, enable it for the MirageOS feed (#3753)

cuihtlauac:
Part of #3753. Depends on **tarides/river#17**.

## Problem

`https://mirage.io/feed.xml` is valid Atom, but its `type="xhtml"` content embeds HTML **void elements** that aren't self-closed (``). Strict XML parsing (xmlm → Syndic) rejects the *whole* feed, so the planet scraper fails it with `Failure "Neither Atom nor RSS2 feed"` and MirageOS posts never appear. hannesm confirmed on #3753 that the feed is otherwise fine.

## Approach — opt-in, per feed

river now has an opt-in `~repair` flag (tarides/river#17) that, **only when a feed fails to parse as-is**, retries after self-closing HTML void elements. Well-formed feeds are never touched.

This PR wires that through the scraper as a **per-source** option, so repair is applied only to feeds we've verified are recoverable this way — not globally:

- `data/planet-sources.yml`: new `repair:` key, set `repair: true` on `mirage`.
- `blog_scraper.ml`: parse the key and pass `~repair` to `River.fetch`. Because repair is purely a scrape-time concern, it's deliberately kept **out** of the published `Data_intf.Blog.source` data model (which would ripple into the bin-packed `data.ml`); it's carried in a local `resolved` record instead.
- `blog_parser.ml`: accept and ignore the new key so the site build still reads the file.
- `ocamlorg.opam.template`: pin river to the branch with `~repair` until tarides/river#17 is released.

## Verification

- river branch built + unit-tested (unclosed / redundantly-closed / self-closed `` shapes; well-formed feed left untouched).
- Proven on the **real** mirage.io feed: strict parse fails; with repair, 78 entries parse.
- `dune build tool/data-scrape/bin/scrape.exe` and `dune build tool/data-packer` both green with the pinned river.

## Relationship to #3755

#3755 removes the two genuinely dead feeds (`ocaml-book`, `emilpriver`) and adds a "wait for upstream" note on `mirage`. This PR **supersedes that note** by actually fixing `mirage`. If #3755 merges first, I'll resolve the trivial overlap on the mirage entry.

## Follow-up

Once tarides/river#17 is released to opam-repository, drop the `river` pin-depends.

### DIFF
--- a/data/planet-sources.yml
+++ b/data/planet-sources.yml
@@ -127,9 +127,13 @@
   name: Mike McClurg
   url: https://mcclurmc.wordpress.com/feed/
   publish_all: false
+# repair: the feed is valid Atom but embeds unclosed HTML void elements
+# (<img>) in its type="xhtml" content, which strict XML parsing rejects.
+# river's ~repair self-closes them before parsing. See tarides/river#17, #3753.
 - id: mirage
   name: MirageOS
   url: https://mirage.io/feed.xml
+  repair: true
 - id: ocaml-book
   name: OCaml Book
   url: http://ocaml-book.com/blog?format=rss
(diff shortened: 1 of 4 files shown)


hannesm:
please don't - we fixed the feed

PR #3767, opened 2026-09-09
package: bound doc-backend fetches and prefer status.json search-index digest (#3765)

### LINKED ISSUE #3765: Doc-page render downloads full odoc search index for a digest → OOM (awso 371 MB index)
## Summary

On 2026-09-08, ocaml.org served intermittent Varnish `503 Backend fetch failed` errors (reported in ocaml/infrastructure#192). @mtelvers diagnosed the root cause: the web-server replicas were being **OOM-killed by the kernel**, repeatedly, because rendering a documentation page downloads the package's entire odoc search index into memory to compute a digest — and one newly published package (`awso`) has a **371 MB** index.

This issue tracks the ocaml.org-side code fixes. Full incident write-up: ocaml/infrastructure#192 (comment by @mtelvers).

## Root cause (confirmed)

- `v3b.ocaml.org` runs two server replicas with no memory limit, on a 15 GB box with no swap. Each replica grew to 10–12 GB and was SIGKILLed (exit 137), roughly every 90 s–3 min. 23 kills, first at 16:10 UTC — lining up with the report at 16:05.
- Varnish itself was fine; it just had no backend to talk to, hence `Backend fetch failed`. Refreshing worked when it hit a moment a replica was briefly up. (The two reported URLs — `/manual/...` and `/packages/search` — are app-local, so a `503 Backend fetch failed` on them means the app was dead, not that those pages have a slow upstream.)

### Why memory exploded

`search_index_digest` (`src/ocamlorg_package/lib/documentation.ml:369`) is called on **every** documentation-page render (three call sites in `handler.ml`, via `Package_helper.search_index_digest`). To produce a cache-busting hash for the page's script URL, it downloads the whole of the package's `index.js` into memory and MD5s it. The TTL cache stores **only the resulting digest**, so every cache miss re-downloads the entire file.

- `awso/0.9.1` `index.js` is **371 MB** (32× the next largest sampled: `core` at ~11 MB). Fetching it from the docs backend takes 60 s+, so requests overlapped and each held its own copy in the heap.
- The load was crawlers walking the newly published `awso` doc tree (Bytespider, AionBot, ClaudeBot, Bingbot) — not malicious, just a huge generated tree crawled for the first time. One replica saw 65 awso page requests / 10 full `index.js` downloads in 5 minutes.

### Secondary contributor

`Sidebar_cache` (same file) uses `Hashtbl.add` rather than `Hashtbl.replace` (`documentation.ml:244`) and never expires, so concurrent misses accumulate duplicate entries. awso's `sidebar.json` is 23 MB / 48 s to fetch.

## Mitigation in place

@mtelvers blocked awso's doc pages at Varnish (returns 503 with `Retry-After: 3600` so well-behaved crawlers back off). Backend memory is now flat at 1–2 GB, no further OOM kills. `/p/awso/*` currently 503s; everything else is normal.

## What actually needs fixing (ocaml.org side)

1. **Don't download the index to compute its digest.** There are already TODOs at `documentation.ml:354` and `:367` saying the digest should be computed by ocaml-docs-ci / voodoo and included in `status.json`. That removes the download entirely and is the real fix. (Needs docs-ci side support — cc @jonludlam, tracked in ocaml/infrastructure#192.)
2. **Add a size ceiling to `http_get`** (`documentation.ml:100`). It does `Cohttp_lwt.Body.to_string` on whatever the docs backend returns, with no cap — so any future large package reintroduces the same OOM. Bound the body size and fail closed above it.
3. **Fix `Sidebar_cache`** to use `Hashtbl.replace` and/or a bounded/expiring cache so concurrent misses can't accumulate duplicates.
4. Optionally: set a memory limit on the replicas so a single bad payload degrades one page rather than crashing the process (infra-side).

## Notes

- Earlier hypothesis in this issue (unbounded/slow docs-backend fetch cascading via timeouts) was directionally right about the vulnerable code (`http_get` / `Cohttp_lwt.Body.to_string`) but wrong on the mechanism: the failure was **OOM SIGKILL from payload size**, not latency, fd exhaustion, or Varnish pool saturation. A fetch *timeout* would not have prevented it — a size ceiling and removing the download-for-digest would.
- Latency sampling of the docs backend on 2026-09-08 showed it healthy (~90–145 ms) for normal packages; the problem was payload size on `awso` specifically, not backend slowness.

## Refs

- ocaml/infrastructure#192

cuihtlauac:
## What & why

Fixes #3765. Documentation-page renders call `search_index_digest`, which
downloaded a package's **entire** odoc search index (`doc/index.js`) into memory
just to MD5 it for a cache-busting script URL. When the `awso` package published
a **371 MB** `index.js`, crawler traffic caused overlapping 60 s+ downloads that
each held their own copy; the server heap grew to 10–12 GB and was repeatedly
OOM-killed, surfacing as site-wide Varnish `503 Backend fetch failed`.

Root-cause write-up: ocaml/infrastructure#192.

## Changes (all in `src/ocamlorg_package/lib/`)

1. **Size ceiling on `http_get`.** Responses larger than
   `OCAMLORG_MAX_DOC_FETCH_BYTES` (default **50 MB**) are rejected: a
   `Content-Length` over the cap is refused without downloading the body, and
   responses without a trustworthy `Content-Length` are streamed and aborted
   once the cap is exceeded. All five call sites already handle `Error`/`None`
   gracefully — for an oversized package the search box is simply absent and the
   page otherwise renders normally.

2. **Prefer the digest from `status.json`.** `search_index_digest` now uses a
   digest published in `status.json` when present, and only falls back to
   downloading-and-hashing the index otherwise. `Status.t` gains an optional
   `search_index_digest` field and is now parsed with `{ strict = false }` so the
   docs backend can extend `status.json` without breaking parsing on
   older/newer deployments. This half is a **no-op until the producer side ships
   the field** (tracked in ocurrent/ocaml-docs-ci#193) — but combined with 1 it
   removes the OOM vector regardless.

3. **`Sidebar_cache` uses `Hashtbl.replace`** instead of `add`, so concurrent
   misses can't accumulate duplicate entries in this never-expiring cache.

## Tests

Adds `documentation_status_test.ml`: `status.json` parses with the digest
missing (→ `None`), present (→ `Some`), and with unknown extra fields (ignored).
`dune build @runtest` is green.

## Notes for reviewers

- The interim mitigation (blocking `/p/awso/*` at Varnish) is already in place;
  this PR is the durable app-side fix.
- `{ strict = false }` is deliberate: `ppx_deriving_yojson` rejects unknown
  fields by default, so without it a future `status.json` carrying the new digest
  field would fail to parse on replicas not yet running this code.

Refs: ocaml/infrastructure#192, ocurrent/ocaml-docs-ci#193

### DIFF
--- a/src/ocamlorg_package/lib/config.ml
+++ b/src/ocamlorg_package/lib/config.ml
@@ -14,6 +14,13 @@ let documentation_status_url =
 let package_caches_ttl =
   env_with_default "OCAMLORG_PACKAGE_CACHES_TTL" "3600" |> float_of_string
 
+(* Upper bound on the size (in bytes) of a single response fetched from the
+   documentation backend. Prevents a pathologically large artifact (e.g. a
+   multi-hundred-MB odoc search index) from being buffered into memory and
+   OOM-killing the server. See ocaml/ocaml.org#3765. Default: 50 MB. *)
+let max_doc_fetch_bytes =
+  env_with_default "OCAMLORG_MAX_DOC_FETCH_BYTES" "52428800" |> int_of_string
+
 let default_cache_dir =
   match Sys.os_type with
   | "Unix" -> Fpath.(v (Sys.getenv "HOME") / ".cache" / "ocamlorg")
--- a/src/ocamlorg_package/lib/config.mli
+++ b/src/ocamlorg_package/lib/config.mli
@@ -4,5 +4,11 @@ val opam_polling : int
 val documentation_url : string
 val documentation_status_url : string
 val package_caches_ttl : float
+
+val max_doc_fetch_bytes : int
+(** Upper bound (bytes) on a single response fetched from the documentation
+    backend, to avoid buffering a pathologically large artifact into memory.
+    Overridable via [OCAMLORG_MAX_DOC_FETCH_BYTES]. *)
+
 val opam_repository_path : Fpath.t
 val package_state_path : Fpath.t
--- a/src/ocamlorg_package/lib/documentation.ml
+++ b/src/ocamlorg_package/lib/documentation.ml
@@ -8,8 +8,16 @@ module Status = struct
     failed : bool;
     files : string list;
     redirections : redirection list;
+    search_index_digest : string option; [@default None]
+        (* Optional: digest of the odoc search index (doc/index.js), computed by
+           ocaml-docs-ci / voodoo. When present, ocaml.org uses it directly
+           instead of downloading the whole index to hash it. See
+           ocaml/ocaml.org#3765 and ocurrent/ocaml-docs-ci#193. *)
   }
-  [@@deriving yojson]
+  [@@deriving yojson { strict = false }]
+  (* [strict = false]: tolerate unknown fields so the docs backend can extend
+     status.json (e.g. adding [search_index_digest]) without breaking parsing on
+     older/newer deployments. *)
 
   let has_file (v : t) (options : string list) : string option =
     let children = v.files in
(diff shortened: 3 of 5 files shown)


cuihtlauac:
@jonludlam, @mtelvers : any views on this?

PR #3769, opened 2026-09-13
fix: point ppx documentation link to latest version

khalidbelk:
On the **Metaprogramming** page, which is maybe the main (/ only?) one covering this topic and the one people visit when searching about it for OCaml, the link currently points to a 4-year-old version of the PPX library documentation page. As you may know, the docs have changed a lot since then and updating the link per-version would be tedious. This change makes sure it always stays up to date with the "latest"-tagged version.


cuihtlauac:
Thanks @khalidbelk

PR #3791, opened 2026-09-22
Adds the feed for The Functional Programmer blog

vim-github:
The feed is filtered to only include posts which bear the OCaml tag.

Feed URL: [funcp.org/feed-ocaml.xml](https://www.funcp.org/feed-ocaml.xml)
Unfiltered feed URL: [funcp.org/feed.xml](https://www.funcp.org/feed.xml)


cuihtlauac (approved):
Thanks for the contribution! Nice touch using the OCaml-tagged feed (`feed-ocaml.xml`) — since the feed is already filtered to OCaml posts upstream, the default `publish_all` is exactly right here. Merging now.

